PR #4, opened 2021-07-05
implemented FrameHeader parsing

TomasKralCZ:
Hi ! This PR implements parsing of the frame headers.

To do that, I needed to change some of the `derive` code:

- `#[size_coder]` now accepts either `implicit(...)` or `explicit(...)`. Implicit will read the size from the `BitReader`, explicit uses some already decoded variable to get the size.
- I added a new attribute `#[aligned]` that calls `jump_to_byte_boundary()` before and after decoding the bundle.
- I removed the `abort` caused by using `#[default]` with `#[size_coder]`. It should work as-is as far as I can see, but let me know if I missed something.

### DIFF
--- a/jxl_headers_derive/src/lib.rs
+++ b/jxl_headers_derive/src/lib.rs
@@ -45,7 +45,7 @@ fn parse_single_coder(input: &syn::Expr, extra_lit: Option<&syn::ExprLit>) -> To
             Some(elit) => quote! {U32::Val(#lit + #elit)},
         },
         syn::Expr::Call(expr_call) => {
-            let bits = get_bits(&expr_call);
+            let bits = get_bits(expr_call);
             match extra_lit {
                 None => quote! {U32::Bits(#bits)},
                 Some(elit) => quote! {U32::BitsOffset{n: #bits, off: #elit}},
@@ -64,7 +64,7 @@ fn parse_single_coder(input: &syn::Expr, extra_lit: Option<&syn::ExprLit>) -> To
             };
             match (&**left, &**right) {
                 (syn::Expr::Call(expr_call), syn::Expr::Lit(lit)) => {
-                    let bits = get_bits(&expr_call);
+                    let bits = get_bits(expr_call);
                     match extra_lit {
                         None => quote! {U32::BitsOffset{n: #bits, off: #lit}},
                         Some(elit) => quote! {U32::BitsOffset{n: #bits, off: #lit + #elit}},
@@ -109,7 +109,7 @@ fn parse_coder(input: syn::Expr) -> TokenStream2 {
     };
 
     match &input {
-        syn::Expr::Call(expr_call) => parse_u2s(&expr_call, None),
+        syn::Expr::Call(expr_call) => parse_u2s(expr_call, None),
         syn::Expr::Binary(syn::ExprBinary {
             attrs: _,
             left,
@@ -123,7 +123,7 @@ fn parse_coder(input: syn::Expr) -> TokenStream2 {
             };
             match (&**left, &**right) {
                 (syn::Expr::Call(expr_call), syn::Expr::Lit(lit)) => {
-                    parse_u2s(&expr_call, Some(&lit))
+                    parse_u2s(expr_call, Some(lit))
                 }
                 _ => abort!(
                     input,
@@ -579,11 +579,11 @@ impl Field {
                     cond.as_deref(),
                     &("\\hyperref[hdr:".to_owned() + &href_ty + "]{" + &minted(&ty) + "}"),
                     dfl.as_deref(),
-                    &ident,
+                    ident,
                 );
             }
             Coder::U32(U32 { coder: _, pretty }) => {
-                add_row(cond.as_deref(), &minted(pretty), dfl.as_deref(), &ident);
+                add_row(cond.as_deref(), &minted(pretty), dfl.as_deref(), ident);
             }
             Coder::Select(
                 Condition {
@@ -610,17 +610,12 @@ impl Field {
                 } else {
                     "!(".to_owned() + condition + ")"
                 };
-                add_row(
-                    Some(&cond_true),
-                    &minted(&coder_true),
-                    dfl.as_deref(),
-                    &ident,
-                );
+                add_row(Some(&cond_true), &minted(coder_true), dfl.as_deref(), ident);
                 add_row(
                     Some(&cond_false),
-                    &minted(&coder_false),
+                    &minted(coder_false),
                     dfl.as_deref(),
-                    &ident,
+                    ident,
                 );
             }
             Coder::Vector(size_coder, value_coder) => {
--- a/jxl_headers_derive/src/lib.rs
+++ b/jxl_headers_derive/src/lib.rs
@@ -135,6 +135,38 @@ fn parse_coder(input: syn::Expr) -> TokenStream2 {
     }
 }
 
+fn parse_size_coder(mut input: syn::Expr) -> TokenStream2 {
+    match input {
+        syn::Expr::Call(syn::ExprCall {
+            ref func,
+            ref mut args,
+            ..
+        }) => {
+            if args.len() != 1 {
+                abort!(input, "Expected 1 argument in sized_coder inner call");
+            }
+
+            match &**func {
+                syn::Expr::Path(expr_path) if expr_path.path.is_ident("implicit") => {
+                    let arg = args.first().unwrap().clone();
+                    parse_coder(arg)
+                }
+                syn::Expr::Path(expr_path) if expr_path.path.is_ident("explicit") => {
+                    quote! { U32Coder::Direct(U32::Val(#args)) }
+                }
+                _ => abort!(
+                    input,
+                    "Unexpected expression in size_coder, must be 'implicit()' or 'explicit()'"
+                ),
+            }
+        }
+        _ => abort!(
+            input,
+            "Unexpected expression in size_coder, must be 'implicit()' or 'explicit()'"
+        ),
+    }
+}
+
 fn prettify_condition(cond: &syn::Expr) -> String {
     (quote! {#cond})
         .to_string()
@@ -266,6 +298,7 @@ struct Field {
     kind: FieldKind,
     ty: syn::Type,
     default: Option<TokenStream2>,
+    default_element: Option<TokenStream2>,
     nonserialized_inits: Vec<TokenStream2>,
 }
 
@@ -284,6 +317,8 @@ impl Field {
         let mut size_coder = None;
 
         let mut nonserialized = vec![];
+
+        let mut default_element = None;
 
         // Parse attributes.
         for a in &f.attrs {
@@ -305,6 +340,13 @@ impl Field {
                     }
                     let default_expr = a.parse_args::<syn::Expr>().unwrap();
                     default = Some(quote! {#default_expr});
+                }
+                Some("default_element") => {
+                    if default_element.is_some() {
+                        abort!(f, "Repeated default_element")
+                    }
+                    let default_element_expr = a.parse_args::<syn::Expr>().unwrap();
+                    default_element = Some(quote! { #default_element_expr })
                 }
                 Some("condition") => {
                     if condition.is_some() {
@@ -365,7 +407,7 @@ impl Field {
                     let coder_ast = a.parse_args::<syn::Expr>().unwrap();
                     let pretty = prettify_coder(&coder_ast);
                     size_coder = Some(U32 {
-                        coder: parse_coder(coder_ast),
+                        coder: parse_size_coder(coder_ast),
                         pretty,
                     });
                 }
@@ -381,6 +423,10 @@ impl Field {
                 }
                 _ => {}
             }
+        }
+
+        if default.is_some() && default_element.is_some() {
+            abort!(f, "default is incompatible with default_element");
         }
 
         if let Some(select_coder) = select_coder {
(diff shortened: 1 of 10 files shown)


src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+    num_passes: u32,
+
+    #[coder(u2S(0, 1, 2, Bits(1) + 3))]
+    #[default(0)]
+    #[condition(num_passes != 1)]
+    num_ds: u32,
+
+    #[size_coder(explicit(num_passes - 1))]
+    #[coder(Bits(2))]
+    #[default(Vec::new())]
```
veluca93, line 35: IIRC this is implicit, no? If so, I'd probably rather not mention it (it likely looks ugly in the generated `.tex` files - IDK if you tried `make_all_tex.sh`)

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+
+#[derive(UnconditionalCoder, Debug, PartialEq, Clone)]
+#[nonserialized(BlendingInfoNonserialized)]
+struct BlendingInfo {
+    #[coder(u2S(0, 1, 2, Bits(2) + 3))]
+    #[default(BlendingMode::Replace)]
+    mode: BlendingMode,
+
+    /* Spec: "Let multi_extra be true if and only if and the number of extra channels is at least two."
+    libjxl condition is num_extra_channels > 0 */
```
veluca93, line 80: @jonsneyers I guess we should fix the spec...

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    #[default(false)]
+    #[condition(nonserialized.num_extra_channels > 0 &&
+        (mode == BlendingMode::Blend || mode == BlendingMode::AlphaWeightedAdd || mode == BlendingMode::Mul))]
+    clamp: bool,
+
+    #[coder(u2S(0, 1, 2, 3))]
+    #[default(0)]
+    /* Let full_frame be true if and only if have_crop is false or if the
+    frame area given by width and height and offsets x0 and y0 completely covers the image area. */
```
veluca93, line 95: Wishlist for the future: being

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+}
+
+struct RestorationFilterNonserialized {
+    encoding: Encoding,
+}
+
+#[derive(UnconditionalCoder, Debug, PartialEq)]
+#[nonserialized(RestorationFilterNonserialized)]
+struct RestorationFilter {
+    // all_default isn't mentioned in the spec, but libjxl has it
```
veluca93, line 110: Also @jonsneyers

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+    #[default(0)]
+    flags: u64,
+
+    #[default(false)]
+    #[condition(!nonserialized.xyb_encoded)]
+    do_ycbcr: bool,
+
+    #[coder(Bits(2))]
+    #[default([0, 0, 0])]
+    #[condition(do_ycbcr && flags & 0x20 == 0)]
```
veluca93, line 236: Can we give names to these constants?

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+    lf_level: u32,
+
+    #[default(false)]
+    #[condition(frame_type != FrameType::LFFrame)]
+    have_crop: bool,
+
+    #[coder(u2S(Bits(8), Bits(11) + 256, Bits(14) + 2304, Bits(30) + 18688))]
+    #[default(0)]
+    #[condition(have_crop && frame_type != FrameType::ReferenceOnly)]
+    x0: u32,
```
veluca93, line 281: I *think* this should be `i32` (it can be negative, for sure -- although IIRC in libjxl we never generate such files)

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+    && (frame_type == kRegularFrame || frame_type == kSkipProgressive) */
+    #[default(BlendingInfo::default())]
+    #[condition(frame_type == FrameType::RegularFrame || frame_type == FrameType::SkipProgressive)]
+    #[nonserialized(num_extra_channels : nonserialized.num_extra_channels,
+        have_crop : have_crop, x0: x0, y0: y0, width: width, height: height,
+        img_width: nonserialized.img_width, img_height: nonserialized.img_height)]
+    blending_info: BlendingInfo,
+
+    #[size_coder(explicit(nonserialized.num_extra_channels))]
+    #[default(vec![BlendingInfo::default(); nonserialized.num_extra_channels as usize])]
```
veluca93, line 308: I don't know how complicated it would be to do so, but it might be nicer if `#[default(...)]` would define the default of *vector elements*, instead of the whole vector. This would also help to avoid some code repetition. WDYT?

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    #[coder(Bits(2))]
+    #[default(0)]
+    #[condition(frame_type != FrameType::LFFrame && !is_last)]
+    save_as_reference: u32,
+
+    /* The default value for save_before_ct is d_sbct = !(full_frame && (frame_type ==
+    kRegularFrame || frame_type == kSkipProgressive) && blending_info.mode ==
+    kReplace && (duration == 0 || save_as_reference != 0) && !is_last). */
+    #[default(!(!have_crop || (x0 == 0 && y0 == 0 &&
```
veluca93, line 338: Same wish as with conditions :)

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub have_animation: bool,
+    pub have_timecode: bool,
+    pub img_width: u32,
+    pub img_height: u32,
+}
+
+#[derive(UnconditionalCoder, Debug, PartialEq)]
+#[nonserialized(FrameHeaderNonserialized)]
+#[aligned]
+pub struct FrameHeader {
```
veluca93, line 214: I believe `#[validate]` should also be here (possibly as a follow up) - at least for validating extra channel upsampling.

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+#[cfg(test)]
+mod test_frame_header {
+    use super::*;
+    use crate::{
+        bit_reader::BitReader,
+        bmff::JxlCodestream,
+        headers::{FileHeaders, JxlHeader},
+    };
+
+    macro_rules! test_frame_header {
```
veluca93, line 364: Is there a specific reason why this is a macro and not a function?

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+    fn test_basic() {
+        test_frame_header!(
+            vec![
+                0xFF, 0x0A, 0x00, 0x90, 0x01, 0x00, 0x12, 0x88, 0x02, 0x00, 0xD4, 0x00, 0x55, 0x0F,
+                0x00, 0x00, 0xA8, 0x50, 0x19, 0x65, 0xDC, 0xE0, 0xE5, 0x5C, 0xCF, 0x97, 0x1F, 0x3A,
+                0x2C, 0xA6, 0x6D, 0x5C, 0x67, 0x68, 0xAB, 0x6D, 0x0B, 0x4B, 0x12, 0x45, 0xC6, 0xB1,
+                0x49, 0x3A, 0x81, 0x43, 0x92, 0x58, 0x04, 0x36, 0x2E, 0x98, 0x07, 0x18, 0x00, 0x86,
+                0x99, 0x03, 0x27, 0x33, 0x50, 0xE4, 0x4A, 0x12, 0x00,
+            ],
+            // TODO(Tomáš Král): It would be nice to use struct update syntax like this:
```
veluca93, line 403: Can we do TODO(username) instead?

src/headers/frame_header.rs
```diff
@@ -0,0 +1,494 @@
...
+                all_default: false,
+                frame_type: FrameType::RegularFrame,
+                encoding: Encoding::Modular,
+                flags: 0,
+                do_ycbcr: false,
+                jpeg_upsampling: [0, 0, 0],
+                upsampling: 1,
+                ec_upsampling: vec![1],
+                group_size_shift: 2,
+                // libjxl x_qm_scale = 2, but condition is false (should be 3 according to the draft)
```
veluca93, line 466: @jonsneyers not that it matters, but we should probably also figure this one out or fix the spec.

veluca93 (approved):

veluca93:
Thanks a lot! Left a few comments.

src/headers/frame_header.rs
TomasKralCZ: Sure !

src/headers/frame_header.rs
TomasKralCZ: Yeah, it's better to use a function here, I originally wrote it in a way which wouldn't be possible with a function, but forgot about it.

src/headers/frame_header.rs
TomasKralCZ: It bothers me too, especially when part of the condition such as `full_frame` has to be duplicated. 
We could introduce a new attribute on top of the struct like `#[free_variables(foo = bar, ...)]`. But I'm just going to add a TODO there for now.

veluca93: Yeah, although I'm not sure how it would work at the top of the struct - I was thinking of adding new members with a #[computed(...)] attribute, which should work without too many problems but would likely waste some amount of space and potentially create inconsistent states...

src/headers/frame_header.rs
TomasKralCZ: Added a TODO for now...

src/headers/frame_header.rs
TomasKralCZ: Hmm, it's a bit surprising, but x0 and y0 are `int32_t` in libjxl.

veluca93: Yes, that can be useful to represent i.e. layers in a saved file of something like GIMP :)

src/headers/frame_header.rs
TomasKralCZ: I should've used better words than implicit / explicit.
My intention was to make it easy to distinguish between expressions like `num_passes` and `Bits(3) + 4` inside the size_coder. Although looking at the draft, size_coders only use `u2S`, so we don't need to care about it.

veluca93: Implicit/explicit is OK (but maybe I'd go with computed/signaled), but I was referring to the default value :)

TomasKralCZ: Oh, got it, I needed to include those in order for `Passes` to implement `Default`.

veluca93: Sounds good, thanks!

to-mas-kral:
Alright, I think all of the comments should be addressed now. Thank you for looking into the code in such detail !

jxl_headers_derive/src/lib.rs
```diff
@@ -306,6 +341,13 @@ impl Field {
                     let default_expr = a.parse_args::<syn::Expr>().unwrap();
                     default = Some(quote! {#default_expr});
                 }
+                Some("default_element") => {
+                    if default_element.is_some() {
+                        abort!(f, "Repeated default_element")
```
veluca93, line 346: This should probably be incompatible with default

src/headers/frame_header.rs
TomasKralCZ: It wouldn't be possible with`#[default] because that needs to be of the same type as the field for the `Default` trait. So I added `#[default_element]` attribute. But I still feel like the solution is a bit dirty...

src/headers/frame_header.rs
```diff
@@ -0,0 +1,544 @@
...
+    #[nonserialized(encoding : encoding)]
+    restoration_filter: RestorationFilter,
+
+    #[default(Extensions::default())]
+    extensions: Extensions,
+}
+
+impl FrameHeader {
+    fn check(&self, nonserialized: &FrameHeaderNonserialized) -> Result<(), Error> {
+        if let Some(val) = self.jpeg_upsampling.iter().find(|x| **x > 3) {
```
veluca93, line 376: I don't think you need this, since it's read by Bits(2), no?

jxl_headers_derive/src/lib.rs
```diff
@@ -579,7 +630,7 @@ impl Field {
                 Field::texify_coder_and_cond(
```
veluca93, line 630: Have you perhaps looked at the generated `.pdf`?

src/headers/frame_header.rs
```diff
@@ -0,0 +1,544 @@
...
+            return Err(Error::InvalidJpegUpsampling(*val));
+        }
+
+        if self.upsampling > 1 {
+            if let Some((info, upsampling)) = nonserialized
+                .extra_channel_info
+                .iter()
+                .zip(&self.ec_upsampling)
+                .find(|(info, upsampling)| info.dim_shift * *upsampling < self.upsampling)
```
veluca93, line 385: Also need to check if it is > 8

veluca93 (approved):

veluca93:
@TomasKralCZ Just wondering if you had time to look at this PR again :)

src/headers/frame_header.rs
TomasKralCZ: Yeah, I'll fix that. Jxl also does some [werid bit shifting](https://gitlab.com/wg1/jpeg-xl/-/blob/main/lib/jxl/frame_header.cc#L209-213) that I don't understand. Do we need to do that as well ?

veluca93: Ah, I'm not sure if dim_shift is a shift count or a multiplier (i.e. if it is 0-1-2 or 1-2-4). I don't think you need to reproduce that bit shifting :)

jonsneyers: It's << dim_shift and * (ec_)upsampling.
(first edition of the spec had that wrong in one place, I fixed it yesterday in the 2nd edition draft).

So an extra channel of width w should after all upsampling have width w * ec_upsampling << dim_shift. And also (ec_upsampling << dim_shift) >= upsampling (i.e. extra channels cannot be higher resolution than the color channels).

veluca93 (approved):

PR #13, opened 2024-09-17
Add owned and borrowed Image and ImageRect.

veluca93:


jxl/src/image.rs
```diff
@@ -0,0 +1,257 @@
...
+        || origin
+            .1
+            .checked_add(size.1)
+            .ok_or(Error::ArithmeticOverflow)?
+            > ssize.1
+    {
+        Err(Error::RectOutOfBounds(
+            size.0, size.1, origin.0, origin.1, ssize.0, ssize.1,
+        ))
+    } else {
```
mo271, line 103: no else needed here, just return `OK(())`?

jxl/src/image.rs
```diff
@@ -0,0 +1,257 @@
...
+        assert!(Image::<u8>::new(1 << 28, 1 << 28).is_err());
+    }
+
+    #[test]
+    fn rect_basic() -> Result<()> {
+        let mut image = Image::<u8>::new(32, 32)?;
+        assert_eq!(image.as_rect_mut().rect((0, 0), (1, 1))?.size(), (1, 1));
+        assert!(image.as_rect_mut().rect((30, 30), (3, 3)).is_err());
+        image.as_rect_mut().rect((30, 30), (1, 1))?.row(0)[0] = 1;
+        assert_eq!(image.as_rect_mut().row(30)[30], 1);
```
mo271, line 246: not super important, but perhaps non testing non-square images as well might make sense?

mo271 (commented):
LGTM

mo271 (approved):

jxl/src/image.rs
veluca93: else is needed to avoid a return above... Doesn't make too much of a difference IMO

PR #14, opened 2024-09-18
Copy container parser from jxl-oxide

tirr-c:
Almost the exact copy of jxl-oxide, except the error type. I'd like to add some notice about original code somewhere (jxl-oxide is MIT licensed after all), but I'm not entirely sure about where to add those...


jxl/src/container/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+                    if buf.starts_with(&Self::CONTAINER_SIG) {
+                        // tracing::debug!("Container signature found");
+                        *state = DetectState::WaitingBoxHeader;
+                        buf.drain(..Self::CONTAINER_SIG.len());
+                        continue;
+                    }
+                    if !Self::CODESTREAM_SIG.starts_with(buf)
+                        && !Self::CONTAINER_SIG.starts_with(buf)
+                    {
+                        // tracing::error!("Invalid signature");
```
veluca93, line 107: Feel free to uncomment those, but I would only use DEBUG/TRACE levels so they are omitted by default in builds.
(I believe there is no way to set the tracing static log level per-crate)

jxl/src/container/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+                            kind: BitstreamKind::Invalid,
+                            bytes_left: None,
+                        };
+                        continue;
+                    }
+                    return Ok(());
+                }
+                DetectState::WaitingBoxHeader => match ContainerBoxHeader::parse(buf)? {
+                    HeaderParseResult::Done { header, size } => {
+                        buf.drain(..size);
```
veluca93, line 118: I believe this is potentially quadratic if many small boxes are present and a large input is fed to the box decoder.
In general, unconditionally buffering the received input is best avoided too.

jxl/src/container/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+            | DetectState::WaitingJxlpIndex(..)
+            | DetectState::InAuxBox { .. } => BitstreamKind::Container,
+            DetectState::InCodestream { kind, .. } | DetectState::Done(kind) => kind,
+        }
+    }
+
+    pub fn feed_bytes(&mut self, input: &[u8]) -> Result<(), Error> {
+        let state = &mut self.state;
+        let buf = &mut self.buf;
+        buf.extend_from_slice(input);
```
veluca93, line 85: Here and below, I would use buf.try_reserve() to make sure that we do not get panics due to failed allocations

jxl/src/container/mod.rs
tirr-c: Yeah, I just realized while thinking about how to reduce buffering. Addressed in c3e8a51.

jxl/src/container/mod.rs
tirr-c: Also addressed in c3e8a51.

jxl/src/container/box_header.rs
```diff
@@ -0,0 +1,101 @@
...
+        header_size: usize,
+    },
+    NeedMoreData,
+}
+
+impl ContainerBoxHeader {
+    pub(super) fn parse(reader: &super::ConcatSlice<'_, '_>) -> Result<HeaderParseResult, Error> {
+        let mut buf = [0u8; 16];
+        let buf = reader.peek(&mut buf);
+        if buf.len() < 8 {
```
veluca93, line 30: Doesn't this follow anyway from the default case in the match?

veluca93 (approved):

jxl/src/container/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Default)]
+enum JxlpIndexState {
+    #[default]
+    Initial,
+    SingleJxlc,
+    Jxlp(u32),
+    JxlpFinished,
+}
+
+struct ConcatSlice<'first, 'second> {
```
veluca93, line 73: Maybe it makes sense to move this in util/?

jxl/src/container/box_header.rs
tirr-c: Ooh, right -- I've removed the if statement.

jxl/src/container/mod.rs
veluca93: https://users.rust-lang.org/t/turn-off-tracing-at-compile-time-for-specific-crate/104337/2

actually, maybe this is a better way... (feel free to do it in a separate PR)

veluca93 (approved):

jxl/src/container/mod.rs
tirr-c: I'd prefer doing that in a separate PR.

Could you merge this if it looks good to you? The branch protection rule doesn't seem to allow me to merge PRs...

PR #15, opened 2024-09-19
Draft interface for a render pipeline.

veluca93:


jxl/src/render/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+                    fill_fn: (x.fill_fn, fake_fill_fn),
+                    group_id: x.group_id,
+                    num_filled_passes: x.num_filled_passes,
+                })
+                .collect(),
+        );
+    }
+
+    // Same as fill_input_same_type, but the inputs might have different types.
+    // Which type is used for which channel is determined by the stages in the render pipeline.
```
tirr-c, line 128: I think the comments could be actual doc comments (`/// ...`) so that docs are picked up by rustdoc and IDEs.

tirr-c (approved):

PR #16, opened 2024-09-19
Added test of ToU8ForWriting.

zond:


veluca93:
Giving this to @sboukortt since he had $opinions

jxl/src/image.rs
```diff
@@ -189,7 +189,7 @@ pub mod debug_tools {
 
     impl ToU8ForWriting for u16 {
         fn to_u8_for_writing(self) -> u8 {
-            (self as u32 * 255 / 65535) as u8
+            (self as u32 >> 8) as u8
```
sboukortt, line 192: My impression is that what would likely make the most sense is something along the lines of: `((self as u32 * 255 + 32767) / 65535) as u8` (roughly equivalent to converting to a real between 0 and 255 and rounding it). How about it?

jxl/src/image.rs
```diff
@@ -201,7 +201,7 @@ pub mod debug_tools {
 
     impl ToU8ForWriting for u32 {
         fn to_u8_for_writing(self) -> u8 {
-            (self as f32 / (2.0f32.powi(32) - 1.0)).to_u8_for_writing()
+            (self >> 24) as u8
```
sboukortt, line 204: Likewise, maybe something like `((self as u64 * 255 + 2147483647) / 4294967295) as u8`.

sboukortt (approved):

jxl/src/image.rs
```diff
@@ -255,4 +255,56 @@ mod test {
         assert!(image.as_rect().to_pgm().starts_with(b"P5\n32 32\n255\n"));
         Ok(())
     }
+
+    #[cfg(feature = "debug_tools")]
```
veluca93, line 259: Might make sense to put these in the debug_tools module (in a further nested `test` submodule)

zond: Done

veluca93 (approved):

PR #17, opened 2024-09-19
Bufferless container parser

tirr-c:
Removes internal buffers from the container parser. Partial reads should be handled by user.

The parser itself doesn't buffer incomplete boxes. The user may decide to buffer data if needed, after pulling parser events.


jxl/src/container/mod.rs
```diff
@@ -84,206 +70,46 @@ impl ContainerParser {
...
+    pub fn process_bytes<'inner, 'buf>(
+        &'inner mut self,
+        input: &'buf [u8],
+    ) -> ParseEvents<'inner, 'buf> {
+        ParseEvents::new(self, input)
     }
 
-    pub fn take_bytes(&mut self) -> Vec<u8> {
-        std::mem::take(&mut self.codestream)
+    /// Get how much bytes are consumed by the previous call of [`process_bytes`].
```
veluca93, line 90: s/much/many/, call *to* process_bytes

jxl/src/container/parse.rs
```diff
@@ -0,0 +1,269 @@
...
+                        JxlpIndexState::Jxlp(expected_index) => {
+                            tracing::debug!(
+                                expected_index,
+                                actual_index = index,
+                                "Out-of-order jxlp box found",
+                            );
+                            return Err(Error::InvalidBox);
+                        }
+                        state => {
+                            tracing::debug!(?state, "invalid jxlp index state in WaitingJxlpIndex");
```
veluca93, line 154: somewhat redundant, no?

jxl/src/container/parse.rs
```diff
@@ -0,0 +1,269 @@
...
+        event.transpose()
+    }
+}
+
+/// Parser event emitted by [`ParseEvents`].
+pub enum ParseEvent<'buf> {
+    /// Bitstream structure is detected.
+    BitstreamKind(BitstreamKind),
+    /// Codestream data is read.
+    Codestream(&'buf [u8]),
```
veluca93, line 256: if I understand correctly, this need not be the full codestream, is that correct?
If so, maybe add a comment about that?

jxl/src/container/parse.rs
veluca93, line 1: What do you think about adding a test (perhaps with `arbtest`) which checks that, no matter how you slice up some input box-format, you ultimately extract the same codestream?

veluca93 (approved):

PR #19, opened 2024-09-24
add basic jxlinfo binary

mo271:


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 debug = true
 
 [workspace]
-members = [ "jxl", "jxl_cli", "jxl_headers_derive"]
+members = [ "jxl", "jxl_cli", "jxl_headers_derive", "jxlinfo"]
```
veluca93, line 5: Why is it not part of `jxl_cli`? You can have more than one binary per crate

mo271: ok, will put it there

mo271: moved, now it runs with 
```
cargo run -p jxl_cli --bin jxlinfo [your filename here]
```

jxl_cli/src/bin/jxlinfo.rs
```diff
@@ -0,0 +1,117 @@
...
+use std::fs;
+use std::io::Read;
+
+fn parse_jxl_codestream(data: &[u8], verbose: bool) -> Result<(), jxl::error::Error> {
+    let mut br = BitReader::new(data);
+    let fh = FileHeaders::read(&mut br)?;
+
+    // Non-verbose output
+    if !verbose {
+        // let color_space = &fh.image_metadata.color_encoding;
```
veluca93, line 21: leftover?

jxl_cli/src/bin/jxlinfo.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        // TODO(firsching): add non-verbose print for bit-depth
+        // TODO(firsching): add non-verbose print for color space
+        return Ok(());
+    }
+
+    // Verbose output: Use Debug trait to print the FileHeaders
+    println!("{:#?}", fh);
+
+    let _icc = if fh.image_metadata.color_encoding.want_icc {
```
veluca93, line 38: Why are you making a binding for the result here?

jxl_cli/src/bin/jxlinfo.rs
mo271: yes, removed

jxl_cli/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 jxl = { path = "../jxl" }
+clap = { version = "4.5.18", features = ["derive"] }
```
veluca93, line 8: I don't think you're actually using the derive feature...

veluca93 (approved):

jxl_cli/Cargo.toml
mo271: you are right: removed it

PR #23, opened 2024-09-28
Implement decoding of ANS streams

tirr-c:
Mostly copied from jxl-oxide, and modified to eliminate `unsafe`. (Reading streams can be branchless with unchecked indexing and by transmuting `Bucket` to `u64`.)

### DIFF
--- a/jxl/src/entropy_coding/ans.rs
+++ b/jxl/src/entropy_coding/ans.rs
@@ -2,6 +2,489 @@
 //
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
+//
+// Originally written for jxl-oxide.
 
-//use crate::bit_reader::BitReader;
-//use crate::error::Error;
+use crate::bit_reader::BitReader;
+use crate::error::{Error, Result};
+
+const LOG_SUM_PROBS: usize = 12;
+const SUM_PROBS: u16 = 1 << LOG_SUM_PROBS;
+
+const RLE_MARKER_SYM: u16 = LOG_SUM_PROBS as u16 + 1;
+
+#[derive(Debug)]
+struct AnsHistogram {
+    buckets: Vec<Bucket>,
+    log_bucket_size: usize,
+    bucket_mask: u32,
+    // For optimizing fast-lossless case.
+    #[allow(unused)]
+    single_symbol: Option<u32>,
+}
+
+// log_alphabet_size <= 8 and log_bucket_size <= 7, so u8 is sufficient for symbols and cutoffs.
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+struct Bucket {
+    alias_symbol: u8,
+    alias_cutoff: u8,
+    dist: u16,
+    alias_offset: u16,
+    alias_dist_xor: u16,
+}
+
+impl AnsHistogram {
+    fn decode_dist_two_symbols(br: &mut BitReader, dist: &mut [u16]) -> Result<usize> {
+        let table_size = dist.len();
+
+        let v0 = Self::read_u8(br)? as usize;
+        let v1 = Self::read_u8(br)? as usize;
+        if v0 == v1 {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let alphabet_size = v0.max(v1) + 1;
+        if alphabet_size > table_size {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let prob = br.read(LOG_SUM_PROBS)? as u16;
+        dist[v0] = prob;
+        dist[v1] = SUM_PROBS - prob;
+
+        Ok(alphabet_size)
+    }
+
+    fn decode_dist_single_symbol(br: &mut BitReader, dist: &mut [u16]) -> Result<usize> {
+        let table_size = dist.len();
+
+        let val = Self::read_u8(br)? as usize;
+        let alphabet_size = val + 1;
+        if alphabet_size > table_size {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        dist[val] = SUM_PROBS;
+
+        Ok(alphabet_size)
+    }
+
+    fn decode_dist_evenly_distributed(br: &mut BitReader, dist: &mut [u16]) -> Result<usize> {
+        let table_size = dist.len();
+
+        let alphabet_size = Self::read_u8(br)? as usize + 1;
+        if alphabet_size > table_size {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let base = SUM_PROBS as usize / alphabet_size;
+        let remainder = SUM_PROBS as usize % alphabet_size;
+        dist[0..remainder].fill(base as u16 + 1);
+        dist[remainder..alphabet_size].fill(base as u16);
+
+        Ok(alphabet_size)
+    }
+
+    fn decode_dist_complex(br: &mut BitReader, dist: &mut [u16]) -> Result<usize> {
+        let table_size = dist.len();
+
+        let mut len = 0usize;
+        while len < 3 {
+            if br.read(1)? != 0 {
+                len += 1;
+            } else {
+                break;
+            }
+        }
+
+        let shift = (br.read(len)? + (1 << len) - 1) as i16;
+        if shift > 13 {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let alphabet_size = Self::read_u8(br)? as usize + 3;
+        if alphabet_size > table_size {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        // TODO(tirr-c): This could be an array of length `SUM_PROB / 4` (4 is from the minimum
+        // value of `repeat_count`). Change if using array is faster.
+        let mut repeat_ranges = Vec::new();
+        let mut omit_data = None;
+        let mut idx = 0;
+        while idx < alphabet_size {
+            dist[idx] = Self::read_prefix(br)?;
+            if dist[idx] == RLE_MARKER_SYM {
+                let repeat_count = Self::read_u8(br)? as usize + 4;
+                if idx + repeat_count > alphabet_size {
+                    return Err(Error::InvalidAnsHistogram);
+                }
+                repeat_ranges.push(idx..(idx + repeat_count));
+                idx += repeat_count;
+                continue;
+            }
+            match &mut omit_data {
+                Some((log, pos)) => {
+                    if dist[idx] > *log {
+                        *log = dist[idx];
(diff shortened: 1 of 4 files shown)


jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+        if shift > 13 {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let alphabet_size = Self::read_u8(br)? as usize + 3;
+        if alphabet_size > table_size {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let mut repeat_ranges = Vec::new();
```
veluca93, line 106: This could be a `[..; MAX_ALPHABET_SIZE]`, instead of a vector, right?
I suspect it wouldn't change much performance wise, but leaving a todo in case it does might be nice.

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+        let mut omit_data = None;
+        let mut idx = 0;
+        while idx < alphabet_size {
+            dist[idx] = Self::read_prefix(br)?;
+            if dist[idx] == 13 {
```
veluca93, line 111: IIRC this is a RLE marker and equal to MAX_LOG_TABLE_SIZE+1 - might be worth making it into a constant

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+        Ok(if bitstream.read(1)? != 0 {
+            let n = bitstream.read(3)?;
+            ((1 << n) + bitstream.read(n as usize)?) as u8
+        } else {
+            0
+        })
+    }
+
+    fn read_prefix(br: &mut BitReader) -> Result<u16> {
+        Ok(match br.read(3)? {
```
veluca93, line 310: I assume this is equivalent to the lookup table in libjxl and the spec - any reason not to do just that?

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+            }
+        } else if br.read(1)? != 0 {
+            // evenly distributed
+            Self::decode_dist_evenly_distributed(br, &mut dist)?
+        } else {
+            // compressed distribution info
+            Self::decode_dist_complex(br, &mut dist)?
+        };
+
+        if let Some(single_sym_idx) = dist.iter().position(|&d| d == 1 << 12) {
```
veluca93, line 216: What do you think of splitting the alias table construction into a separate function?

Also, shouldn't it be table_size instead of 1<<12 here?

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+            buckets[o].alias_cutoff -= by;
+            buckets[u].alias_symbol = o as u16;
+            buckets[u].alias_offset = buckets[o].alias_cutoff;
+            match buckets[o].alias_cutoff.cmp(&bucket_size) {
+                std::cmp::Ordering::Less => underfull.push(o),
+                std::cmp::Ordering::Equal => {}
+                std::cmp::Ordering::Greater => overfull.push(o),
+            }
+        }
+
```
veluca93, line 267: assert that both undrfull and overfull are empty?

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+        debug_assert!((5..=8).contains(&log_alpha_size));
+        let table_size = (1u16 << log_alpha_size) as usize;
+        // 4 <= log_bucket_size <= 7
+        let log_bucket_size = 12 - log_alpha_size;
+        let bucket_size = 1u16 << log_bucket_size;
+
+        let mut dist = vec![0u16; table_size];
+        let alphabet_size = if br.read(1)? != 0 {
+            if br.read(1)? != 0 {
+                // binary
```
veluca93, line 202: This comment seems somewhat superfluous :)

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+                continue;
+            }
+            if *code > 1 {
+                let zeros = (*code - 1) as i16;
+                let bitcount = (shift - ((12 - zeros) >> 1)).clamp(0, zeros);
+                *code = (1 << zeros) + ((br.read(bitcount as usize)? as u16) << (zeros - bitcount));
+            }
+            prev_dist = *code;
+            acc += *code;
+            if acc > (1 << 12) {
```
veluca93, line 174: I'm pretty sure this check should be `acc >= table_size` instead (and similarly above; and below it should be `table_size - acc`), or at least this seems to be what libjxl does...

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+
+    pub fn read(&self, br: &mut BitReader, ctx: usize, state: &mut u32) -> Result<u32> {
+        self.histograms[ctx].read(br, state)
+    }
+}
 
-//use crate::bit_reader::BitReader;
-//use crate::error::Error;
+#[inline]
+pub fn ans_validate(state: u32) -> Result<()> {
```
veluca93, line 405: It seems better to me to make an `AnsReader` (or similar) newtype around `u32` with a `read` function taking a `AnsCodes` and a `finalize` or similar method to check the final state, wdyt? This has the advantage of making the existence of a `AnsHistogram` type an implementation-detail, in case we ever want to avoid that i.e. for better performance.

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct AnsCodes {
+    histograms: Vec<AnsHistogram>,
+}
+
+impl AnsCodes {
+    pub fn decode(num: usize, log_alpha_size: usize, br: &mut BitReader) -> Result<AnsCodes> {
```
veluca93, line 392: Can we add a test that checks that either reading a histogram fails, or the resulting histogram has the appropriate sum?

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,410 @@
...
+        if v0 == v1 {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let alphabet_size = v0.max(v1) + 1;
+        if alphabet_size > table_size {
+            return Err(Error::InvalidAnsHistogram);
+        }
+
+        let prob = br.read(12)? as u16;
```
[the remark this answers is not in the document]
tirr-c: Ah yeah, I should have done that...

jxl/src/entropy_coding/ans.rs
tirr-c: This is one of the oldest code in jxl-oxide, and I think I wasn't aware of doing prefix code with lookup tables at that time.

jxl/src/entropy_coding/ans.rs
tirr-c: I think `1 << 12` is correct, because the sum of probability is `1 << 12`.

tirr-c: Uh well, it seems that probabilities sum to `table_size` in libjxl but they sum to `1 << 12` in the spec.

tirr-c: Nevermind, it's a bit confusing that libjxl and jxl-oxide differ slightly... libjxl uses `range = 1 << precision_bits` and probs sum to `range`, where `precision_bits` is always 12 (`ANS_LOG_TAB_SIZE`).

veluca93: Ah, right, the table size is always the same, it's just the alphabet size that changes, I had forgotten that ^^'.
So ignore all the `MAX` from my comments on constant names :P

jxl/src/entropy_coding/ans.rs
tirr-c: Newtype around ANS state sounds great, though I think `AnsHistogram` is already hidden implementation detail (it's a private type in `ans` module).

tirr-c:
I've addressed most of the reviews. I'll add tests tomorrow.

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,486 @@
...
+                continue;
+            }
+            if *code > 1 {
+                let zeros = (*code - 1) as i16;
+                let bitcount = (shift - ((LOG_SUM_PROBS as i16 - zeros) >> 1)).clamp(0, zeros);
+                *code = (1 << zeros) + ((br.read(bitcount as usize)? as u16) << (zeros - bitcount));
+            }
+            prev_dist = *code;
+            acc += *code;
+            if acc > SUM_PROBS {
```
veluca93, line 181: Shouldn't this be `>=`? I don't think it is valid to have the omit_pos symbol have a probability of 0...

jxl/src/entropy_coding/ans.rs
```diff
@@ -2,6 +2,486 @@
...
+
+#[derive(Debug)]
+pub struct AnsReader(u32);
+
+impl AnsReader {
+    /// Expected final ANS state.
+    const CHECKSUM: u32 = 0x130000;
+
+    pub fn new_unused() -> Self {
+        Self(0)
```
veluca93, line 414: `Self(Self::CHECKSUM)`?

veluca93 (approved):

jxl/src/entropy_coding/ans.rs
tirr-c: The spec says `D[omit_pos] >= 0`, but yeah, omit symbol having a probability of 0 doesn't make much sense... Is this a spec bug?

veluca93: Interesting, libjxl definitely checks for `D[omit_pos] > 0`: https://github.com/libjxl/libjxl/blob/main/lib/jxl/dec_ans.cc#L178C1-L183C1

@jonsneyers thoughts?

jonsneyers: So this is either a spec bug or libjxl being too strict. If we follow the spec, what happens? Having `D[omit_pos] == 0` is never useful, right? It's just a non-optimal way of writing a distribution.
I think it's best to see it as a spec bug and update the spec to make the libjxl behavior correct.

veluca93: I think I agree with Jon here - let's call this a spec bug, and check for strictly-positive (i.e. `acc >= SUM_PROBS`)

veluca93 (approved):

PR #24, opened 2024-09-29
Simple pipeline implementation.

veluca93:
Also add f64 to the possible image types, use an explicit enum to tag image types, and rework the pipeline interface.

### DIFF
--- a/jxl/src/image.rs
+++ b/jxl/src/image.rs
@@ -3,35 +3,72 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+use std::fmt::Debug;
+
 use crate::error::{Error, Result};
 
 mod private {
     pub trait Sealed {}
 }
 
-pub trait ImageDataType: private::Sealed + Copy + Default + 'static {
-    fn data_type_id() -> usize;
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum DataTypeTag {
+    U8,
+    U16,
+    U32,
+    F32,
+    I8,
+    I16,
+    I32,
+    F16,
+    F64,
+}
+
+pub trait ImageDataType: private::Sealed + Copy + Default + 'static + Debug {
+    /// ID of this data type. Different types *must* have different values.
+    const DATA_TYPE_ID: DataTypeTag;
+
+    fn from_f64(f: f64) -> Self;
+    fn to_f64(self) -> f64;
 }
 
 macro_rules! impl_image_data_type {
-    ($ty: ty, $id: literal) => {
+    ($ty: ty, $id: ident) => {
         impl private::Sealed for $ty {}
         impl ImageDataType for $ty {
-            fn data_type_id() -> usize {
-                $id
+            const DATA_TYPE_ID: DataTypeTag = DataTypeTag::$id;
+            fn from_f64(f: f64) -> $ty {
+                f as $ty
+            }
+            fn to_f64(self) -> f64 {
+                self as f64
             }
         }
     };
 }
 
-impl_image_data_type!(u8, 0);
-impl_image_data_type!(u16, 1);
-impl_image_data_type!(u32, 2);
-impl_image_data_type!(f32, 3);
-impl_image_data_type!(i8, 4);
-impl_image_data_type!(i16, 5);
-impl_image_data_type!(i32, 6);
-impl_image_data_type!(half::f16, 7);
+impl_image_data_type!(u8, U8);
+impl_image_data_type!(u16, U16);
+impl_image_data_type!(u32, U32);
+impl_image_data_type!(f32, F32);
+impl_image_data_type!(i8, I8);
+impl_image_data_type!(i16, I16);
+impl_image_data_type!(i32, I32);
+
+impl private::Sealed for half::f16 {}
+impl ImageDataType for half::f16 {
+    const DATA_TYPE_ID: DataTypeTag = DataTypeTag::F16;
+    fn from_f64(f: f64) -> half::f16 {
+        half::f16::from_f64(f)
+    }
+    fn to_f64(self) -> f64 {
+        <half::f16>::to_f64(self)
+    }
+}
+
+// Meant to be used by the simple render pipeline and in general for
+// testing purposes.
+impl_image_data_type!(f64, F64);
 
 pub struct Image<T: ImageDataType> {
     size: (usize, usize),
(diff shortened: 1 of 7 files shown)


veluca93:
The code is certainly missing at least a few tests, but given that the PR is large enough as it is I figured I'd have you take a look first ;) (and possibly split it up in more manageable commits if the overall approach seems sound)

jxl/src/render/simple_pipeline.rs
```diff
@@ -0,0 +1,547 @@
...
+    fn build(mut self) -> Result<Self::RenderPipeline> {
+        let channel_info = &mut self.pipeline.channel_info;
+        for stage in (0..self.pipeline.stages.len()).rev() {
+            for chan in 0..channel_info[0].len() {
+                if channel_info[stage][chan].ty.is_none() {
+                    channel_info[stage][chan].ty = channel_info[stage + 1][chan].ty;
+                }
+                // Arithmetic overflows here should be very uncommon, so custom error variants
+                // are probably unwarranted.
+                channel_info[stage][chan].downsample.0 = channel_info[stage][chan]
```
sboukortt, line 101: Would a reference to this tuple make sense? I feel it would make this slightly easier to follow.

jxl/src/render/mod.rs
```diff
@@ -3,103 +3,106 @@
...
 /// call that happens inside image boundaries.
-pub trait RenderPipelineInputStage<InputT: ImageDataType> {
-    fn uses_channel(&self, c: usize) -> bool;
-
-    /// `row` has as many entries as the number of used channels for the stage, i.e. the channels
-    /// for which `uses_channel` returns true.
-    fn process_row_chunk(&mut self, position: (usize, usize), xsize: usize, row: &[&[InputT]]);
+/// For these stages, `process_row_chunk` receives read-only access to each row in each channel.
+pub struct RenderPipelineInputStage<InputT: ImageDataType> {
+    _phantom: PhantomData<InputT>,
```
sboukortt, line 25: What are these phantoms for?

jxl/src/image.rs
```diff
@@ -3,35 +3,72 @@
...
+impl_image_data_type!(i32, I32);
+
+impl private::Sealed for half::f16 {}
+impl ImageDataType for half::f16 {
+    const DATA_TYPE_ID: DataTypeTag = DataTypeTag::F16;
+    fn from_f64(f: f64) -> half::f16 {
+        half::f16::from_f64(f)
+    }
+    fn to_f64(self) -> f64 {
+        <half::f16>::to_f64(self)
```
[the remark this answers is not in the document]
veluca93: If I just do `Self::to_f64`, this becomes a recursive call - and the fully qualified method call syntax is `<Type>::method()`.

jxl/src/render/mod.rs
veluca93: Rust needs every generic parameter to be used in the struct definition, `PhantomData` is a lang-item ZST that is usually used for that purpose

jxl/src/render/internal.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum RenderPipelineStageType {
+    Input,
+    InPlace,
+    InOut,
+    Extend,
+}
+
+pub trait RenderPipelineStageInfo: super::simple_pipeline::RenderPipelineRunStage {
```
tirr-c, line 21: I'm not sure if I understand the design correctly... How do you add additional render pipeline implementation, when `simple_pipeline::RenderPipelineRunStage` needs to be impl-ed for all stages? Doesn't it conflict with existing `RenderPipelineRunStage` impls?

jxl/src/render/mod.rs
```diff
@@ -3,103 +3,106 @@
...
     error::Result,
     image::{ImageDataType, ImageRectMut},
 };
 
+mod internal;
+mod simple_pipeline;
+
+use internal::RenderPipelineStageInfo;
+
+// pub use simple_pipeline::SimpleRenderPipeline;
```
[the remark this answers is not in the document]
veluca93: Ah that was not meant to be commented

jxl/src/render/simple_pipeline.rs
veluca93: Made the code somewhat nicer

jxl/src/render/internal.rs
veluca93: The idea would be that an additional implementation adds another constraint on `RenderPipelineStageInfo` for its own `RenderPipelineRunStage` trait, which is implemented (by the pipeline impl itself) on all the RenderPipelineExtendStage etc (which are just marker types to be used in `ActualStage::Type`).
Does that seem reasonable?

tirr-c: Yeah that would work, I first thought the supertrait bound will not change. Though I think `#[cfg]`-ing render pipeline would be somewhat tedious, because every combination of cfg would need its own trait bound, but still it might not be an issue if there won't be so many implementations...

veluca93: I don't expect more than 2 implementations :-)
Also I did not come up with a better idea, so...

jxl/src/util/shift_right_ceil.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use std::ops::{Add, Shl, Shr, Sub};
+
+pub trait ShiftRightCeil: Copy {
+    fn shrc<T: Copy>(self, rhs: T) -> Self
+    where
+        Self: Shr<T, Output = Self> + Shl<T, Output = Self>;
+}
+
+impl<S: Copy + Add<Self, Output = Self> + Sub<Self, Output = Self> + From<u8>> ShiftRightCeil
```
mo271, line 14:
```suggestion
impl<S: Copy + Add<Self, Output = Self> + Sub<Self, Output = Self> + From<u8>  + Shl<T, Output = Self> + Shr<T, Output = Self>> ShiftRightCeil
```

Don't we also need shifting here?

mo271 (approved):
LGTM

We should add tests for simple_pipeline.rs and shift_right_ceil.rs, I suppose...

jxl/src/util/shift_right_ceil.rs
veluca93: If it compiles this way, we don't need additional constraints ;)

veluca93:
> LGTM
> 
> We should add tests for simple_pipeline.rs and shift_right_ceil.rs, I suppose...

I added tests for shrc, and I am wondering whether it would be best to put tests for the pipeline in another CL - ideally we'd want some more elaborate test infrastructure for all (both?) pipelines...

PR #32, opened 2024-10-05
Use `Display` to set render stage names

tirr-c:
Use `Display::fmt` instead of separate `name` method, like [`std::error::Error`][error] does. (`.to_string()` can be used if `String` is needed.)

[error]: https://doc.rust-lang.org/stable/std/error/trait.Error.html

### DIFF
--- a/jxl/src/render/simple_pipeline.rs
+++ b/jxl/src/render/simple_pipeline.rs
@@ -77,10 +77,9 @@ impl RenderPipelineBuilder for SimpleRenderPipelineBuilder {
     fn add_stage<Stage: RenderPipelineStage>(mut self, stage: Stage) -> Result<Self> {
         let current_info = self.pipeline.channel_info.last().unwrap().clone();
         info!(
-            "adding stage '{}'. last stage channel info: {:?} can_shift {}",
-            stage.name(),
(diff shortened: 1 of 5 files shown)


jxl/src/render/simple_pipeline.rs
```diff
@@ -652,11 +643,9 @@ trait RunStage: Any {
     fn as_any(self: Box<Self>) -> Box<dyn Any>;
     fn input_type(&self) -> DataTypeTag;
     fn output_type(&self) -> DataTypeTag;
-    #[allow(unused)]
-    fn name(&self) -> String;
 }
 
-impl<T: RenderPipelineStage> RunStage for T {
+impl<T: RenderPipelineStage + std::fmt::Display> RunStage for T {
```
veluca93, line 648: This should be redundant, right?

veluca93 (approved):
Good idea, I wasn't sure about using display when I decided to add name() but it does look quite a bit nicer this way

jxl/src/render/simple_pipeline.rs
tirr-c: Ah, nice catch!

PR #40, opened 2024-11-10
Fix Huffman decoding table construction with long tables.

veluca93:
Also add some debug instrumentation to Huffman parsing and rename `util/tracing` to `util/tracing_wrappers`.


tirr-c (approved):
Seems to be working great! Now it's time to debug my ICC profile decoder...

PR #42, opened 2024-11-12
read all frames in jxlinspect

mo271:


jxl_cli/src/bin/jxlinspect.rs
```diff
@@ -127,7 +129,42 @@ fn parse_jxl_codestream(data: &[u8], verbose: bool) -> Result<(), jxl::error::Er
         println!("ICC profile length: {} bytes", icc_data.len());
     }
     // TODO(firsching): add frame header parsing for each frame
-
+    if file_header.image_metadata.animation.is_some() {
```
zond, line 132: I assume the spec only allows multiple frames if `image_metadata.animation.is_some()`, but since the frame header has an `is_last` field, wouldn't it be more defensive to the presence of more frame headers as the criteria for showing multiple frame headers? That way a broken jxl image (with multiple frame headers but no animation set) would still be more easily debuggable using jxlinspect. WDYT?

mo271: no, the spec also allows multiple frame if it is not an animation.
I guess we have to decide what info about frames we want to display in general...

jxl_cli/src/bin/jxlinspect.rs
```diff
@@ -60,74 +63,133 @@ fn parse_jxl_codestream(data: &[u8], verbose: bool) -> Result<(), jxl::error::Er
...
-        return Ok(());
     }
+    //TODO(firsching): handle frames which are blended together, also within animations.
+    if file_header.image_metadata.animation.is_some() {
+        let mut total_duration = 0.0f64;
+        let animation = file_header
+            .image_metadata
+            .animation
+            .as_ref()
+            .expect("This should never fail, it was just check in the if condition above");
```
veluca93, line 136: you could just unwrap() here, but ideally you should use `if let Some(animation) = file_header.image_metadata.animation {`...

jxl_cli/src/bin/jxlinspect.rs
```diff
@@ -60,74 +63,133 @@ fn parse_jxl_codestream(data: &[u8], verbose: bool) -> Result<(), jxl::error::Er
...
 
-    // Verbose output: Use Debug trait to print the FileHeaders
-    println!("{:#?}", file_header);
-
-    // TODO(firsching): consider printing more of less information for ICC
-    // for verbose/non-verbose cases
-    if file_header.image_metadata.color_encoding.want_icc {
-        let icc_data = read_icc(&mut br)?;
-        println!("ICC profile length: {} bytes", icc_data.len());
+            let ms = (frame_header.duration as f64) * 1000.0 * (animation.tps_denominator as f64)
```
veluca93, line 146: I would make this a `duration(&self, &AnimationHeader)` method on `frame_header`

jxl_cli/src/bin/jxlinspect.rs
```diff
@@ -60,74 +63,133 @@ fn parse_jxl_codestream(data: &[u8], verbose: bool) -> Result<(), jxl::error::Er
...
+                &TocNonserialized {
+                    num_entries: num_toc_entries,
+                },
+            )
+            .unwrap();
+            let entries = toc.entries;
+            let num_bytes_to_skip: u32 = entries.into_iter().sum();
+            br.jump_to_byte_boundary()?;
+            not_is_last = !frame_header.is_last;
+            // TODO: use return value
```
veluca93, line 170: seems used to me :-)

jxl_cli/src/bin/jxlinspect.rs
```diff
@@ -60,74 +63,133 @@ fn parse_jxl_codestream(data: &[u8], verbose: bool) -> Result<(), jxl::error::Er
...
+    //TODO(firsching): handle frames which are blended together, also within animations.
+    if file_header.image_metadata.animation.is_some() {
+        let mut total_duration = 0.0f64;
+        let animation = file_header
+            .image_metadata
+            .animation
+            .as_ref()
+            .expect("This should never fail, it was just check in the if condition above");
+        let mut not_is_last = true;
+        while not_is_last {
```
veluca93, line 138: Any thoughts about making this a `fn for_each_frame<F: FnMut(FrameHeader, Toc, ...) -> Result<bool>>(br: BitReader, f: F) -> Result<()>` or similar? (with the function returning Err for errors, Ok(true) for "continue to next frame", and Ok(false) for "exit")

veluca93 (approved):

jxl_cli/src/bin/jxlinspect.rs
zond: General Rust question: What are the benefits of a function with a callback as opposed to an iterator?

veluca93: Iterators are harder to write :-)

I would also be OK with an iterator, fwiw.

jxl_cli/src/bin/jxlinspect.rs
mo271: done
(with `if let Some(ref animation)= ...`)

jxl_cli/src/bin/jxlinspect.rs
mo271: removed

jxl_cli/src/bin/jxlinspect.rs
mo271: done

jxl_cli/src/bin/jxlinspect.rs
mo271: will do in a follow-up, added TODO

PR #44, opened 2024-11-12
Decode ICC profile stream

tirr-c:
No tests yet, but it seems to work fine with CMYK profile used in conformance test.

### DIFF
--- a/jxl/src/icc.rs
+++ b/jxl/src/icc.rs
@@ -3,23 +3,116 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+use std::io::Cursor;
+
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+
 use crate::bit_reader::*;
 use crate::entropy_coding::decode::Histograms;
-use crate::error::Error;
+use crate::error::{Error, Result};
 use crate::headers::encodings::*;
+use crate::util::tracing_wrappers::warn;
+
+mod header;
+mod stream;
+mod tag;
+
+use header::read_header;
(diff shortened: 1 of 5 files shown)


jxl/src/icc.rs
```diff
@@ -3,23 +3,113 @@
...
+use stream::{read_varint_from_reader, IccStream};
+use tag::{read_single_command, read_tag_list};
 
 const ICC_CONTEXTS: usize = 41;
+const ICC_HEADER_SIZE: u64 = 128;
+
+fn read_icc_inner(stream: &mut IccStream) -> Result<Vec<u8>, Error> {
+    let output_size = stream.read_varint()?;
+    let commands_size = stream.read_varint()?;
+    if stream.bytes_read() + commands_size > stream.len() {
```
veluca93, line 30: might want to do a checked_add here

jxl/src/icc.rs
```diff
@@ -3,23 +3,113 @@
...
+    let data_stream = stream;
+
+    // Decode ICC profile header.
+    let mut decoded_profile = read_header(data_stream, output_size)?;
+    if output_size <= ICC_HEADER_SIZE {
+        return Ok(decoded_profile);
+    }
 
+    // Convert to slice writer to prevent buffer from growing.
+    decoded_profile.resize(output_size as usize, 0);
```
veluca93, line 56: Do we want to do try_reserve here?

veluca93 (approved):
Can we leave a TODO about making ICC decoding resumable?

jxl/src/icc.rs
veluca93: It's fine as is for me :-)

PR #45, opened 2024-11-13
Added copy of file header image size to frame header.

zond:


veluca93 (approved):

jxl/src/headers/frame_header.rs
```diff
@@ -407,6 +407,16 @@ pub struct FrameHeader {
 
     #[default(Extensions::default())]
     extensions: Extensions,
+
+    #[coder(Bits(0))]
+    #[default(nonserialized.img_width)]
```
mo271, line 412: Let's initialize directly with the condition below, so that the functions become superflous

mo271 (approved):

PR #47, opened 2024-11-19
LfGlobal decoding: LF quantization factors, scaffolding

veluca93:


jxl/src/frame/quantizer.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub inv_quant_factors: [f32; 3],
+}
+
+impl LfQuantFactors {
+    pub fn new(br: &mut BitReader) -> Result<LfQuantFactors> {
+        let mut quant_factors = [0.0f32; 3];
+        if br.read(1)? == 1 {
+            quant_factors[0] = 1.0 / 4096.0;
+            quant_factors[1] = 1.0 / 512.0;
+            quant_factors[2] = 1.0 / 256.0;
```
mo271, line 25: I suppose one could set
const DEFAULT_QUANT_FACTORS: [f32; 3] = [1.0 / 4096.0, 1.0 / 512.0, 1.0 / 256.0];
and then copy_from_slice or something, also looks good as is...

mo271 (commented):
LGTM

jxl/src/frame/quantizer.rs
veluca93: IMO it's clear enough as it is :-)

mo271 (approved):

PR #48, opened 2024-11-19
add Non444ChromaSubsampling check

mo271:


jxl/src/headers/frame_header.rs
```diff
@@ -417,6 +420,26 @@ pub struct FrameHeader {
     #[default(if frame_height == 0 { nonserialized.img_height } else { frame_height })]
     #[condition(false)]
     pub height: u32,
+
+    #[coder(Bits(0))]
+    #[default({
```
veluca93, line 425: Make that into a free function? It's a fair amount of code for an attribute, and effectively duplicated

mo271: like this?

jxl/src/headers/frame_header.rs
```diff
@@ -430,22 +451,13 @@ impl FrameHeader {
...
-        let mut maxvs = 0;
-        for ch in self.jpeg_upsampling {
-            maxhs = maxhs.max(H_SHIFT[ch as usize]);
-            maxvs = maxvs.max(V_SHIFT[ch as usize]);
-        }
         let xsize = self.width as usize;
         let ysize = self.height as usize;
-        let xsize_blocks = xsize.div_ceil(BLOCK_DIM << maxhs) << maxhs;
-        let ysize_blocks = ysize.div_ceil(BLOCK_DIM << maxvs) << maxvs;
+        let xsize_blocks = xsize.div_ceil(BLOCK_DIM << self.maxhs) << self.maxhs;
```
veluca93, line 459: For later: might now make sense to move this logic to `num_groups`/`num_dc_groups` and drop `group_counts`

jxl/src/headers/frame_header.rs
```diff
@@ -256,6 +256,17 @@ pub struct FrameHeaderNonserialized {
     pub img_height: u32,
 }
 
+const H_SHIFT: [usize; 4] = [0, 1, 1, 0];
+const V_SHIFT: [usize; 4] = [0, 1, 0, 1];
+
+fn compute_max(jpeg_upsampling: &[u32], shift_table: &[usize]) -> u32 {
```
veluca93, line 262: nit: compute_jpeg_shift?

jxl/src/headers/frame_header.rs
mo271: done, see #50

PR #54, opened 2024-11-24
Implement header decoding for Modular groups

veluca93:


jxl/src/frame/modular/predict.rs
```diff
@@ -0,0 +1,39 @@
...
+    Select = 4,
+    Gradient = 5,
+    Weighted = 6,
+    TopRight = 7,
+    TopLeft = 8,
+    LeftLeft = 9,
+    Average1 = 10,
+    Average2 = 11,
+    Average3 = 12,
+    Average4 = 13,
```
jonsneyers, line 26: We are using too many different names for the predictors. The above are from the libjxl code I guess?

predictor index | spec name | jxl_from_tree name | libjxl code | proposal
-- | -- | -- | -- | --
0 | Zero | Set | Zero | Zero
1 | West | W | Left | West
2 | North | N | Top | North
3 | Avg(W,N) | AvgW+N | Average0 | AvgW_N
4 | Select | Select | Select | Select
5 | Gradient | Gradient | Gradient | Gradient
6 | Self-correcting | Weighted | Weighted | Weighted
7 | NorthEast | NE | TopRight | NorthEast
8 | NorthWest | NW | TopLeft | NorthWest
9 | WestWest | WW | LeftLeft | WestWest
10 | Avg(W,NW) | AvgW+NW | Average1 | AvgW_NW
11 | Avg(N,NW) | AvgN+NW | Average2 | AvgN_NW
12 | Avg(N,NE) | AvgN+NE | Average3 | AvgN_NE
13 | AvgAll | AvgAll | Average4 | AvgAll

The rightmost column is what I would propose to use as enum values in both the libjxl and jxl-rs code.

veluca93: Switched to North/East/West names, but somewhat more verbose.

jxl/src/frame/modular/tree.rs
```diff
@@ -0,0 +1,165 @@
...
+            if tree.len() >= tree.capacity() {
+                tree.try_reserve(tree.len() * 2 + 1)?;
+            }
+            to_decode -= 1;
+            let property = tree_reader.read(br, PROPERTY_CONTEXT)?;
+            trace!(property);
+            if let Some(property) = property.checked_sub(1) {
+                // inner node.
+                if property > 255 {
+                    return Err(Error::InvalidProperty(property));
```
jonsneyers, line 71: Strictly speaking the spec does allow properties > 255, so this check is too strict. It would only make sense for images with a _lot_ of channels, and it wouldn't be something libjxl would ever produce, but in principle it is even legal to have a tree using property number 10000000 which is always always referencing some previous channel that doesn't exist so the value of the property is always zero.

It would probably make sense to add some sanity limits to the maximum property index that is allowed; e.g. limit it to 63 for Level 5 (which would be enough for 12 previous channels, since there are 16 fixed properties 0..15 and then 4 per prev-channel) and 255 for Level 10 (which is enough for 60 previous channels).

jonsneyers (approved):

jxl/src/frame/modular/predict.rs
jonsneyers: That works too, it's not like the length of those enum values really matters anyway...

jxl/src/frame/modular/tree.rs
veluca93: This is the same check that libjxl has IIRC. I'd add it to the spec as a level 10 limit as you suggest.

mo271 (approved):

PR #55, opened 2024-11-25
`assert_almost_eq!`

mo271:


jxl_test_utils/src/lib.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#[macro_export]
```
veluca93, line 6: `pub use assert_almost_eq` works if staying in the same crate.

jxl_test_utils/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    };
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_with_floats() {
```
veluca93, line 33: Can you also test NaNs?

veluca93 (commented):
Isn't a new crate a bit overkill? Feels like a module in `util/test.rs` would be sufficient...

jxl_test_utils/src/lib.rs
mo271: done, also INFINITY

jxl_test_utils/src/lib.rs
mo271: move to util/test.rs

jxl/src/util/test.rs
```diff
@@ -0,0 +1,101 @@
...
+        );
+
+        assert_almost_eq!(1.0, 1.0, INFINITY);
+
+        let result = panic::catch_unwind(|| {
+            assert_almost_eq!(NAN, NAN, INFINITY);
+        });
+        assert!(
+            result.is_err(),
+            "Expected assert_almost_eq! to panic, but it didn't"
```
veluca93, line 90: https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#testing-panics :-)

jxl/src/util/test.rs
```diff
@@ -0,0 +1,101 @@
...
+        );
+
+        let result = panic::catch_unwind(|| {
+            assert_almost_eq!(INFINITY, INFINITY, INFINITY);
```
veluca93, line 94: It's surprising to me that this should panic.

veluca93: I agree that this is what *happens*, but I am not sure I agree that this is what _should_ happen ;)

jxl/src/util.rs
```diff
@@ -15,3 +15,5 @@ pub use bits::*;
 pub use concat_slice::*;
 pub use log2::*;
 pub use shift_right_ceil::*;
+
+mod test;
```
veluca93, line 19: I assume you want this to be pub ;-)

jxl/src/util/test.rs
veluca93: Then again probably we don't want infinities to be around anyway, so this might be fine

jxl/src/util/test.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#[macro_export]
```
veluca93, line 6: I think `pub use assert_almost_eq` and then importing `util::test::assert_almost_eq` would be better

jxl/src/util.rs
mo271: done

jxl/src/util/test.rs
mo271: using `pub(crate)...` now

jxl/src/render/stages/upsample.rs
```diff
@@ -87,7 +87,11 @@ impl RenderPipelineStage for Upsample2x {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{error::Result, image::Image, render::test::make_and_run_simple_pipeline};
+    //use util::test::assert_almost_eq;
```
veluca93, line 90: leftover

mo271: removed...

PR #59, opened 2024-11-27
start with splines

mo271:


jxl/src/error.rs
```diff
@@ -111,6 +111,8 @@ pub enum Error {
     InvalidPredictor(u32),
     #[error("Invalid modular mode property: {0}")]
     InvalidProperty(u32),
+    #[error("To many splines: {0}, limit is {1}")]
```
veluca93, line 114: *too

veluca93 (commented):
Seems good, I assume you intend to proceed with decoding before merging?

mo271:
> Seems good, I assume you intend to proceed with decoding before merging?

perhaps let's merge this now, because this completes reading all what is needed for splines.

I added two out-of-bounds checks (not in spec) just like we have in libjxl

veluca93 (approved):

PR #60, opened 2024-11-27
Added noise stage

zond:


zond:
I realized I probably need to add a test for the noise parameter decoding as well.

zond:
> I realized I probably need to add a test for the noise parameter decoding as well.

OK, I think it's properly ready to review now.

jxl/src/frame.rs
```diff
@@ -145,10 +145,11 @@ impl Frame {
             None
         };
 
-        if self.header.has_noise() {
-            info!("decoding noise");
-            todo!("noise not implemented");
-        }
+        let noise = if self.header.has_noise() {
+            Some(Noise::read(br)?)
```
veluca93, line 149: Can you leave the `info!`?

zond: Done!

jxl/src/frame.rs
```diff
@@ -188,3 +191,39 @@ impl Frame {
...
+        bit_reader::BitReader,
+        container::ContainerParser,
+        error::Error,
+        headers::{FileHeader, JxlHeader},
+        util::test::assert_almost_eq,
+    };
+
+    use super::Frame;
+
+    fn read_frame(image: &[u8]) -> Result<Frame, Error> {
```
veluca93, line 207: not now, but at some point we might want to move this into src/util/test.rs or similar.

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// TODO(zond): remove once we use this!
+#![allow(dead_code, unused_variables)]
```
veluca93, line 7: I'm not sure we need `unused_variables` here

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+
+use crate::{
+    features::noise::Noise,
+    render::{RenderPipelineInOutStage, RenderPipelineInPlaceStage, RenderPipelineStage},
+};
+
+pub struct ConvolveNoiseStage {
```
veluca93, line 14: Since this operates independently on each channel, wouldn't it make more sense to have a stage that operates on a channel at a time? (I am aware this is different from what libjxl does :-))

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+                others += input[2][x + 3];
+                others += input[2][x + 4];
+                output[0][x] = others * 0.16 + input[2][x + 2] * -3.84;
+            }
+        }
+    }
+}
+
+pub struct AddNoiseStage {
+    noise: Noise,
```
veluca93, line 70: Any reason not to have this be a StrengthEvalLut which owns Noise?

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+    // clamp doesn't handle NaN the same way as min/max.
+    #[allow(clippy::manual_clamp)]
+    f32::max(0.0, f32::min(noise_model.eval(x), 1.0))
+}
+
+impl RenderPipelineStage for AddNoiseStage {
+    type Type = RenderPipelineInPlaceStage<f32>;
+
+    fn uses_channel(&self, c: usize) -> bool {
+        c >= self.first_channel && c < self.first_channel + 3
```
veluca93, line 133: This doesn't seem correct, we should be using the 3 color channels [0,3) and [first_noise_channel, first_noise_channel+3) 

Relatedly, since `process_row_chunk` only gets passed channels that are used, we should just use 0..3 and 3..6 there as indices.
Also, the constructor should probably assert `first_noise_channel >= 3`.

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+            let vx = row[0][x];
+            let vy = row[1][x];
+            let in_g = vy - vx;
+            let in_r = vy + vx;
+            let noise_strength_g = noise_strength(&noise_model, in_g * 0.5);
+            let noise_strength_r = noise_strength(&noise_model, in_r * 0.5);
+            let addit_rnd_noise_red = row_rnd_r * norm_const;
+            let addit_rnd_noise_green = row_rnd_g * norm_const;
+            let addit_rnd_noise_correlated = row_rnd_c * norm_const;
+            println!(
```
veluca93, line 161: Debugging leftover?

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+        }
+        assert!(
+            !stage.uses_channel(4),
+            "shouldn't use channel > first_channel + 3"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn convolve_noise_process_row_chunk() -> Result<()> {
```
veluca93, line 221: I would rather test properties such as "doing the noise convolution on a constant image produces an all-0 image" rather than specific values.

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+        features::noise::Noise,
+        image::Image,
+        render::{
+            stages::noise::{AddNoiseStage, ConvolveNoiseStage},
+            test::make_and_run_simple_pipeline,
+            RenderPipelineStage,
+        },
+        util::test::assert_almost_eq,
+    };
+    use test_log::test;
```
veluca93, line 198: Please add stage consistency tests as is done for other stages.

jxl/src/render/stages/noise.rs
veluca93: Also testing the impulse response on a 5x5 all-0 image except for the center pixel which is 1 seems a reasonable test.

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    #[test]
+    fn convolve_noise_uses_channels() -> Result<()> {
```
veluca93, line 201: Not sure if there's a lot of value in testing the formula we wrote above...

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+                "{}, {}, {}, {}, {}, {}, {}",
+                addit_rnd_noise_red,
+                addit_rnd_noise_green,
+                addit_rnd_noise_correlated,
+                noise_strength_g,
+                noise_strength_r,
+                ytox,
+                ytob
+            );
+            let k_rg_corr = 0.9921875f32;
```
veluca93, line 171: nit: you don't need the `f32`

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+        let low = self.noise.lut[floor_x_int];
+        let hi = self.noise.lut[floor_x_int + 1];
+        (hi - low) * frac_x + low
+    }
+}
+
+fn noise_strength(noise_model: &StrengthEvalLut, x: f32) -> f32 {
+    // clamp doesn't handle NaN the same way as min/max.
+    #[allow(clippy::manual_clamp)]
+    f32::max(0.0, f32::min(noise_model.eval(x), 1.0))
```
veluca93, line 126: the eval LUT should never produce NaN, so we can use clip here.

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,357 @@
...
+        assert!(stage.uses_channel(1), "should use channel == first_channel");
+        assert!(
+            !stage.uses_channel(4),
+            "shouldn't use channel > first_channel + 3"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn add_noise_process_row_chunk() -> Result<()> {
```
veluca93, line 259: I'd prefer to test specific situations for which we can easily know what is supposed to happen.

For example, we could have a constant noise LUT, 0 values in the first two noise channels, and check that adding noise does nothing on color channel 0, and adds the 0.0034375*noise_channel_2 to the second and third color channels.

Another reasonable test is to verify that the applied noise does not depend on the third color channel.

Some other tests with the noise LUT below should also be relatively easy, as in that setup noise.eval(x) ought to just be a linear function in the appropriate range.

jxl/src/frame.rs
```diff
@@ -188,3 +191,39 @@ impl Frame {
...
+        result.decode_lf_global(&mut br)?;
+        Ok(result)
+    }
+
+    #[test]
+    fn noise() -> Result<(), Error> {
+        let frame = read_frame(include_bytes!("../resources/test/8x8_noise.jxl"))?;
+        let lf_global = frame.lf_global.unwrap();
+        let noise = lf_global.noise.unwrap();
+        let want_noise = [
```
mo271, line 221: perhaps `want_noise` could be a const, but I'm not sure what is more idiomatic here.

jxl/src/render/stages/noise.rs
zond: Done

jxl/src/render/stages/noise.rs
zond: Removed it

jxl/src/render/stages/noise.rs
zond: Done

jxl/src/render/stages/noise.rs
zond: Ah yes, removed it :P

jxl/src/render/stages/noise.rs
zond: Done

jxl/src/render/stages/noise.rs
zond: I moved the entire strength calculation to the Noise struct instead, that makes more sense IMO.

jxl/src/render/stages/noise.rs
zond: I have no opinion about this. I suggest we leave it like this for now and let anyone interested refactor it.

jxl/src/render/stages/noise.rs
zond: Removed it.

jxl/src/render/stages/noise.rs
zond: Added TODO(mo271) for this as per Discord discussion.

jxl/src/render/stages/noise.rs
zond: Added TODO(mo271) for this as per Discord discussion.

jxl/src/frame.rs
mo271: then not...

jxl/src/render/stages/noise.rs
veluca93: sure, can you leave a TODO then?

veluca93 (approved):

jxl/src/render/stages/noise.rs
zond: OK, adding the TODO was ~ as much work as making the change, so I made the change. WDYT?

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,319 @@
...
+        util::test::assert_almost_eq,
+    };
+    use test_log::test;
+
+    /*
+     * TODO(mo271): Add more relevant ConvolveNoise tests.
+     *
+     * I would rather test properties such as "doing the noise convolution on a constant image produces an all-0 image" rather than specific values.
+     *
+     * Member
```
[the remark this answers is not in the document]
zond: I replaced the text with a link to this PR.

Removed the UI copy :P

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+impl RenderPipelineStage for ConvolveNoiseStage {
+    type Type = RenderPipelineInOutStage<f32, f32, 2, 2, 0, 0>;
+
+    fn uses_channel(&self, c: usize) -> bool {
+        c == self.channel
+    }
+
+    // TODO: This
```
veluca93, line 37: leftover todo :-)

jxl/src/render/stages/noise.rs
```diff
@@ -0,0 +1,293 @@
...
+    fn process_row_chunk(
+        &mut self,
+        _position: (usize, usize),
+        xsize: usize,
+        row: &mut [(&[&[f32]], &mut [&mut [f32]])],
+    ) {
+        for (input, output) in row {
```
veluca93, line 44: could be a `let (input, output) = row[0];` now

jxl/src/render/stages/noise.rs
zond: Removed

jxl/src/render/stages/noise.rs
zond: Done

PR #61, opened 2024-11-28
start with patches

mo271:


jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+const PATCH_POSITION_CONTEXT: usize = 4;
+const PATCH_BLEND_MODE_CONTEXT: usize = 5;
+const PATCH_OFFSET_CONTEXT: usize = 6;
+const PATCH_COUNT_CONTEXT: usize = 7;
+const PATCH_ALPHA_CHANNEL_CONTEXT: usize = 8;
+const PATCH_CLAMP_CONTEXT: usize = 9;
+const NUM_PATCH_DICTIONARY_CONTEXTS: usize = 10;
+
+// Blend modes
+// The new values are the old ones. Useful to skip some channels.
```
veluca93, line 31: What about a enum? (and then using `#[derive(FromPrimitive)]`)

(probably also makes sense for the contexts)

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+const PATCH_BLEND_MODE_BLEND_BELOW: u8 = 5;
+// The new values (in the crop) are added to the old ones if alpha>0:
+// For first alpha channel: sample = sample = old + new * (1 - old)
+// For other channels: sample = old + alpha * new
+const PATCH_BLEND_MODE_ALPHA_WEIGHTED_ADD_ABOVE: u8 = 6;
+const PATCH_BLEND_MODE_ALPHA_WEIGHTED_ADD_BELOW: u8 = 7;
+const PATCH_BLEND_MODE_NUM_BLEND_MODES: u8 = 8;
+
+#[inline]
+fn uses_alpha(blend_mode: u8) -> bool {
```
veluca93, line 66: ... then these can become methods.

Btw, why the `#[inline]`?

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+struct PatchBlending {
+    mode: u8,
+    alpha_channel: usize,
+    clamp: bool,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct PatchReferencePosition {
+    // Not using `ref` like in the spec here, because it is a keyword.
+    reference: usize,
```
veluca93, line 88: note: you *could* use r#ref, if you wanted (but don't, reference is perfectly fine)

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+    ) -> Result<PatchesDictionary> {
+        let num_extra_channels = decoder_state.extra_channel_info().len();
+        let blendings_stride = num_extra_channels + 1;
+        let patches_histograms = Histograms::decode(NUM_PATCH_DICTIONARY_CONTEXTS, br, true)?;
+        let mut patches_reader = patches_histograms.make_reader(br)?;
+        let num_ref_patch = patches_reader.read(br, NUM_REF_PATCH_CONTEXT)? as usize;
+        let num_pixels = xsize * ysize;
+        let max_ref_patches = 1024 + num_pixels / 4;
+        let max_patches = max_ref_patches * 4;
+        let _max_blending_infos = max_patches * 4;
```
veluca93, line 126: Was this supposed to be used?

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+
+            if total_patches > max_patches {
+                return Err(Error::PatchesTooMany(total_patches, max_patches));
+            }
+
+            if next_size < total_patches {
+                next_size *= 2;
+                next_size = std::cmp::min(next_size, max_patches);
+            }
+            positions.reserve(next_size);
```
veluca93, line 188: These should probably be try_reserve.

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+        let max_patches = max_ref_patches * 4;
+        let _max_blending_infos = max_patches * 4;
+        if num_ref_patch > max_ref_patches {
+            return Err(Error::PatchesTooMany(num_ref_patch, max_ref_patches));
+        }
+        let mut total_patches = 0;
+        let mut next_size = 1;
+        let mut positions: Vec<PatchPosition> = Vec::new();
+        let mut blendings = Vec::new();
+        let mut ref_positions: Vec<PatchReferencePosition> = Vec::with_capacity(num_ref_patch);
```
veluca93, line 134: new() + try_reserve?

(we could also impl try_with_capacity() in util, I guess we use the pattern more than enough)

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+            let ref_pos_xsize = patches_reader.read(br, PATCH_SIZE_CONTEXT)? as usize + 1;
+            let ref_pos_ysize = patches_reader.read(br, PATCH_SIZE_CONTEXT)? as usize + 1;
+            let reference_frame = decoder_state.reference_frame(reference);
+            match reference_frame {
+                None => return Err(Error::PatchesInvalidReference(reference)),
+                Some(reference) => {
+                    if !reference.saved_before_color_transform {
+                        return Err(Error::PatchesPostColorTransform());
+                    }
+                    if x0 + ref_pos_xsize > reference.frame[0].size.0 {
```
veluca93, line 155: Add a TODO about checking that this check is correct also in the presence of downsampled extra channels? (and also in libjxl)

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,290 @@
...
+            if total_patches > max_patches {
+                return Err(Error::PatchesTooMany(total_patches, max_patches));
+            }
+
+            if next_size < total_patches {
+                next_size *= 2;
+                next_size = std::cmp::min(next_size, max_patches);
+            }
+            positions.reserve(next_size);
+            blendings.reserve(next_size * PATCH_BLEND_MODE_NUM_BLEND_MODES as usize);
```
veluca93, line 189: note that reserve() takes as argument the *additional* space to ensure, not the total size...

jxl/src/frame.rs
```diff
@@ -91,12 +150,8 @@ impl Frame {
         self.toc.entries.iter().map(|x| *x as usize).sum()
     }
 
-    pub fn is_last(&self) -> bool {
-        self.header.is_last
-    }
-
     /// Given a bit reader pointing at the end of the TOC, returns a vector of `BitReader`s, each
-    /// of which reads a specific section.
+    /// of which reads a specific sectionalpha_channel.
```
veluca93, line 154: this looks like a typo :-)

jxl/src/frame.rs
```diff
@@ -190,6 +256,27 @@ impl Frame {
 
         Ok(())
     }
+
+    pub fn finalize(mut self) -> Result<Option<DecoderState>> {
+        if self.header.can_be_referenced {
+            // TODO(firsching): actually decode the image here, instead of setting it
```
veluca93, line 262: Probably the image has already been decoded by this point, in the long run... It probably should not be decoded here at all

jxl/src/frame.rs
```diff
@@ -190,6 +256,27 @@ impl Frame {
...
+            // to a blank image here, once we can decode images.
+            self.decoder_state.reference_frames[self.header.save_as_reference as usize] =
+                Some(ReferenceFrame::blank(
+                    self.header.width as usize,
+                    self.header.height as usize,
+                    // Set num_channels to "1" here unconditionally for now, since we will use this
+                    // channel for the size checks
+                    1,
```
veluca93, line 270: why not 3 + self.decoder_state.extra_channel_info().len()?

jxl/src/frame.rs
```diff
@@ -207,28 +294,34 @@ mod test {
...
-            let mut frame = Frame::new(&mut br, &file_header)?;
-            let is_last = frame.is_last();
+            let mut frame = Frame::new(&mut br, decoder_state)?;
             let mut sections = frame.sections(&mut br)?;
             frame.decode_lf_global(&mut sections[frame.get_section_idx(Section::LfGlobal)])?;
-            frames.push(frame);
-            if is_last {
+
+            // Call the callback with the frame
+            if let Some(state) = callback(frame)? {
```
veluca93, line 313: I'm not sure I like the interface her - I think `callback(frame_index, &frame)` would likely work better (and should still let us do the same checks), but no strong opinion for now

veluca93 (approved):

jxl/src/features/patches.rs
mo271: ack

jxl/src/features/patches.rs
mo271: made both an enum

jxl/src/features/patches.rs
mo271: now it is a method

The inline was just there because libjxl had it. I hope the compiler is smart enough to make the right choice here even without it.

veluca93: I assume in libjxl it was because the function was in a header :-)

jxl/src/features/patches.rs
```diff
@@ -0,0 +1,321 @@
...
+    PatchPosition = 4,
+    PatchBlendMode = 5,
+    PatchOffset = 6,
+    PatchCount = 7,
+    PatchAlphaChannel = 8,
+    PatchClamp = 9,
+}
+
+impl PatchContext {
+    const NUM_PATCH_DICTIONARY_CONTEXTS: usize = 10;
```
veluca93, line 37: could also just be NUM, PatchContext::NUM is clear enough IMO

jxl/src/features/patches.rs
mo271: yes, for some check, using it now...

jxl/src/features/patches.rs
mo271: using just `NUM` now

jxl/src/features/patches.rs
mo271: done

jxl/src/features/patches.rs
mo271: done. 
Will make a pr with try_with_capacity (and remove a few places where we currently only do `with_capacity` next).
Leaving this unresolved, so I don't forget

jxl/src/features/patches.rs
mo271: done

jxl/src/features/patches.rs
mo271: Hah, then sometimes the calls in libjxl are reserving less than the capacity of the vector size, which then has no effect. Not quite sure if this is intended. I think in any case asymptotically this makes sense. We could do exacly the same here by doing 
```
if next_size > positions.len() {
    positions.reserve(next_size - positions.len());
}

```
Or, and I did this now, because I think it makes sense, to add `1` to `next_size`, everytime we push to positions. Then we don't need the check because as an invariant `next_size` will be larger than `positions.len`verytime we push to positions.
Please check that this works, or I can just go back to exactly what libjxl does, which does not quite make sense to me.

jxl/src/frame.rs
mo271: yes, fixed

jxl/src/frame.rs
mo271: changed

jxl/src/frame.rs
mo271: sure, why not, will be deleted anyhow. (I'm testing on a grayscale image, so "1" seemed more natural)

jxl/src/frame.rs
mo271: ok, then let's leave it for now...

jxl/src/features/patches.rs
veluca93: Could also do `positions.try_reserve(next_size.saturating_sub(positions.len())`, I guess..

mo271: ok, doing that now...

veluca93 (approved):

PR #68, opened 2024-12-15
Modular mode: apply transforms to channel shapes.

veluca93:


jxl/src/headers/frame_header.rs
```diff
@@ -71,6 +72,25 @@ struct Passes {
     last_pass: Vec<u32>,
 }
 
+impl Passes {
+    pub fn get_downsampling_bracket(&self, pass: usize) -> (usize, usize) {
```
mo271, line 76: how about `downsampling_bracket`, without the `get_`, not important of course...

mo271 (approved):
LGTM

PR #69, opened 2024-12-16
Added more bits and pieces to the spline rendering.

zond:


jxl/src/features/spline.rs
```diff
@@ -270,11 +347,213 @@ pub struct Splines {
...
+fn draw_centripetal_catmull_rom_spline(points: &Vec<Point>) -> Vec<Point> {
+    if points.is_empty() {
+        return [].to_vec();
+    }
+    if points.len() == 1 {
+        return vec![points[0]];
+    }
+    const NUM_POINTS: usize = 16;
+    // Create a view of points with one prepended and one appended point.
+    let extended_points = ExtendedPoints {
```
veluca93, line 409: While this works, it doesn't seem to me like the most idiomatic way to do this -- what about `iter::once(first).chain(points.iter()).chain(iter::once(last)).scan()`, for example?
I have not thought too much about the other uses of extended_points below, but ISTM that they'd be reasonable to express as operations on iterators too, avoiding the need for ExtendedPoints.

jxl/src/features/spline.rs
```diff
@@ -270,11 +347,213 @@ pub struct Splines {
...
+        .enumerate()
+        .map(|(index, point)| {
+            if index + 1 < extended_points.len() {
+                (extended_points[index + 1] - point).abs().sqrt()
+            } else {
+                0.0
+            }
+        })
+        .collect();
+    let mut result = Vec::<Point>::with_capacity((points.len() - 1) * NUM_POINTS + 1);
```
veluca93, line 427: please use the try_reserve() methods :-)

jxl/src/features/spline.rs
zond: The extended points list is indexed a few steps in front of the iterator all over the place, so a regular iterator wouldn't be enough.

What do you think is a reasonable way to avoid copying everything but still mapping index 0 to first_element and index len-1 to last_element?

jxl/src/features/spline.rs
zond: Done

jxl/src/features/spline.rs
veluca93: I would not use indexing in the first place. All indexing accesses contiguous positions, so it should be doable with iterators without major issues, which likely results in more efficient and more idiomatic code.

zond: Done, is this better?

jxl/src/features/spline.rs
```diff
@@ -270,11 +350,207 @@ pub struct Splines {
...
+    // Create a view of points with one prepended and one appended point.
+    let first_point = points[0] + (points[0] - points[1]);
+    let last_point =
+        points[points.len() - 1] + (points[points.len() - 1] - points[points.len() - 2]);
+    let extended_points = iter::once(&first_point)
+        .chain(points.iter())
+        .chain(iter::once(&last_point));
+    // Pair each point with the sqrt of the distance to the next point.
+    let points_and_deltas = extended_points
+        .chain(iter::once(&last_point))
```
veluca93, line 420: I assume this is not intended.

jxl/src/features/spline.rs
```diff
@@ -270,11 +350,207 @@ pub struct Splines {
...
+    if points.len() == 1 {
+        return Ok([points[0]].to_vec());
+    }
+    const NUM_POINTS: usize = 16;
+    // Create a view of points with one prepended and one appended point.
+    let first_point = points[0] + (points[0] - points[1]);
+    let last_point =
+        points[points.len() - 1] + (points[points.len() - 1] - points[points.len() - 2]);
+    let extended_points = iter::once(&first_point)
+        .chain(points.iter())
```
veluca93, line 416: Might make sense to do `points.iter().cloned()` and iterate over `Point` instead of `&Point`

jxl/src/features/spline.rs
```diff
@@ -270,11 +350,207 @@ pub struct Splines {
     segment_y_start: Vec<usize>,
 }
 
+struct ExtendedPoints<'a> {
+    first_element: Point,
+    last_element: Point,
+    backend: &'a Vec<Point>,
+}
+
+impl ExtendedPoints<'_> {
```
veluca93, line 359: I don't think this is needed anymore

jxl/src/features/spline.rs
```diff
@@ -270,11 +350,207 @@ pub struct Splines {
...
+    let first_point = points[0] + (points[0] - points[1]);
+    let last_point =
+        points[points.len() - 1] + (points[points.len() - 1] - points[points.len() - 2]);
+    let extended_points = iter::once(&first_point)
+        .chain(points.iter())
+        .chain(iter::once(&last_point));
+    // Pair each point with the sqrt of the distance to the next point.
+    let points_and_deltas = extended_points
+        .chain(iter::once(&last_point))
+        .tuple_windows()
```
veluca93, line 421: (note: we can probably avoid tuple_windows() and the itertools dependency by using `scan` instead, but let's ignore that for now)

veluca93 (approved):

sboukortt (approved):

jxl/src/features/spline.rs
zond: Done

jxl/src/features/spline.rs
zond: Windowing required some padding - and it still requires a bit of padding but not in the exact same way.

jxl/src/features/spline.rs
zond: Done

PR #70, opened 2024-12-18
read all the frames

mo271:
not just the first one

### DIFF
--- a/jxl/src/frame.rs
+++ b/jxl/src/frame.rs
@@ -209,18 +209,26 @@ mod test_frame {
 
     use super::Frame;
 
-    fn read_frame(image: &[u8]) -> Result<Frame, Error> {
+    fn read_frames(image: &[u8]) -> Result<Vec<Frame>, Error> {
         let codestream = ContainerParser::collect_codestream(image).unwrap();
         let mut br = BitReader::new(&codestream);
         let file_header = FileHeader::read(&mut br).unwrap();
-        let mut result = Frame::new(&mut br, &file_header)?;
-        result.decode_lf_global(&mut br)?;
-        Ok(result)
+        let mut frames = vec![];
+        loop {
+            let mut frame = Frame::new(&mut br, &file_header)?;
+            let is_last = frame.is_last();
+            frame.decode_lf_global(&mut br)?;
+            frames.push(frame);
+            if is_last {
+                break;
+            }
+        }
+        Ok(frames)
     }
-
+    // TODO(rename)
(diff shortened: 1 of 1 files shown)


mo271:
probably still needs to ~use the TOC to~ jump over pass groups

jxl/src/frame.rs
```diff
@@ -209,18 +209,26 @@ mod test_frame {
 
     use super::Frame;
 
-    fn read_frame(image: &[u8]) -> Result<Frame, Error> {
+    fn read_frames(image: &[u8]) -> Result<Vec<Frame>, Error> {
```
veluca93, line 212: I just noticed now that the mod is called test_frame... why is it not just test?

jxl/src/frame.rs
```diff
@@ -209,18 +209,26 @@ mod test_frame {
...
+            frame.decode_lf_global(&mut br)?;
+            frames.push(frame);
+            if is_last {
+                break;
+            }
+        }
+        Ok(frames)
     }
-
+    // TODO(rename)
```
veluca93, line 228: ... can you rename it? :-)

jxl/src/frame.rs
```diff
@@ -209,18 +209,26 @@ mod test_frame {
...
         let mut br = BitReader::new(&codestream);
         let file_header = FileHeader::read(&mut br).unwrap();
-        let mut result = Frame::new(&mut br, &file_header)?;
-        result.decode_lf_global(&mut br)?;
-        Ok(result)
+        let mut frames = vec![];
+        loop {
+            let mut frame = Frame::new(&mut br, &file_header)?;
+            let is_last = frame.is_last();
+            frame.decode_lf_global(&mut br)?;
```
veluca93, line 220: TODO() read / skip rest of frame?

mo271: How to to that exactly? How many bits to skip?

veluca93: I am not sure it's very easy, which is why I'd make it a Todo :-)

jxl/src/frame.rs
mo271: done

jxl/src/frame.rs
mo271: done

jxl/src/frame.rs
mo271: added todo, but without this, it will just fail on the three images with multiple frames that we have...

PR #74, opened 2024-12-20
Added more parts of the spline rendering.

zond:


jxl/src/features/spline.rs
```diff
@@ -439,7 +445,132 @@ fn for_each_equally_spaced_point<F: FnMut(Point, f32)>(
     f(points[points.len() - 1], accumulated_distance);
 }
 
-const DESIRED_RENDERING_DISTANCE: f32 = 1.0;
+trait Dct32 {
+    fn continuous_idct(&self, t: f32) -> f32;
+}
+
+impl Dct32 for [f32; 32] {
+    fn continuous_idct(&self, t: f32) -> f32 {
```
veluca93, line 453: Why is this a trait instead of a free function?

jxl/src/features/spline.rs
```diff
@@ -439,7 +445,132 @@ fn for_each_equally_spaced_point<F: FnMut(Point, f32)>(
...
+        &mut self,
+        spline: &Spline,
+        points_to_draw: &[(Point, f32)],
+        length: f32,
+        segments_by_y: &mut Vec<(u64, usize)>,
+        desired_distance: f32,
+    );
+}
+
+impl SplineSegments for Vec<SplineSegment> {
```
veluca93, line 514: Same here, could be two functions, no?

veluca93 (approved):

jxl/src/features/spline.rs
zond: I just like the aesthetic, but I can change it to a function. WDYT?

mo271 (approved):

jxl/src/features/spline.rs
veluca93: I prefer functions unless there's a good reason to go with traits :-)

zond: Done

jxl/src/features/spline.rs
zond: Done

PR #75, opened 2024-12-22
add try_with_capacity

mo271:


mo271:
I suppose one could also extend the `Vec` trait and add `try_with_capacity` to the extension, so we could write `Vec::try_with_capacity`, not sure which one you had in mind, @veluca93.

veluca93:
I was thinking of adding a `TryWithCapacity` trait implemented on `Vec<T>` (and on other types where it might be useful, i.e. `String` or `HashSet`/`HashMap`, when it does become useful)

mo271:
tried that, now I'm getting 
```
error: an associated function with this name may be added to the standard library in the future
```
Not sure how to silence this best. Of course the hope would be that the newly added function does exactly what is done now.

veluca93:
> tried that, now I'm getting
> 
> ```
> error: an associated function with this name may be added to the standard library in the future
> ```
> 
> Not sure how to silence this best. Of course the hope would be that the newly added function does exactly what is done now.

That's annoying... I guess we could just use a different name (say, try_new? new_with_capacity? make?)

mo271:
now using `new_with_capacity` with a note to switch to `try_with_capacity` as soon as it is available.

veluca93 (approved):

PR #76, opened 2024-12-24
cleanup a few `allow(unused)`

mo271:


jxl/src/util/test.rs
veluca93, line 1: You could also make the module test-only instead, i.e. `#[cfg(test)]` on `pub mod test`

veluca93 (approved):

jxl/src/util/test.rs
mo271: much nicer!

PR #78, opened 2024-12-28
XYB to linear sRGB stage

tirr-c:
Implement F32 in-place XYB to linear sRGB stage.

### DIFF
--- a/jxl/src/headers/transform_data.rs
+++ b/jxl/src/headers/transform_data.rs
@@ -21,9 +21,9 @@ pub struct OpsinInverseMatrix {
     #[default([11.031566901960783, -9.866943921568629, -0.16462299647058826,
                -3.254147380392157,  4.418770392156863,  -0.16462299647058826,
                -3.6588512862745097, 2.7129230470588235, 1.9459282392156863])]
-    inverse_matrix: [f32; 9],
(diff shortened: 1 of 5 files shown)


jxl/src/headers/transform_data.rs
```diff
@@ -21,9 +21,9 @@ pub struct OpsinInverseMatrix {
     #[default([11.031566901960783, -9.866943921568629, -0.16462299647058826,
                -3.254147380392157,  4.418770392156863,  -0.16462299647058826,
                -3.6588512862745097, 2.7129230470588235, 1.9459282392156863])]
-    inverse_matrix: [f32; 9],
-    #[default([0.0037930732552754493, 0.0037930732552754493, 0.0037930732552754493])]
-    opsin_biases: [f32; 3],
+    pub inverse_matrix: [f32; 9],
+    #[default([-0.0037930732552754493; 3])]
```
veluca93, line 25: Uhm, where's the sign change coming from?

veluca93 (approved):

jxl/src/headers/transform_data.rs
tirr-c: Uh, the spec says it's negative.

veluca93: Ok, so does libjxl (took way too long to figure that out...)
I guess it was just wrong before :-)

PR #82, opened 2025-01-01
Increase tolerance for linear_to_srgb_fast_arb.

veluca93:


tirr-c (approved):
Maybe the comment in libjxl also needs to be updated?

veluca93:
> Maybe the comment in libjxl also needs to be updated?

@eustas can you double check that when we're back from holidays? :-)

jxl/src/color/tf.rs
```diff
@@ -146,7 +146,7 @@ mod test {
 
             linear_to_srgb(&mut samples);
             linear_to_srgb_fast(&mut fast);
-            assert_all_almost_eq!(&samples, &fast, 1.2e-4);
+            assert_all_almost_eq!(&samples, &fast, 1.7e-4);
```
eustas, line 149: 23 times less than 1/255, so looks ok to me.

PR #83, opened 2025-01-02
Implement BT.709 transfer function

tirr-c:
Also copied from jxl-oxide which is ported from libjxl.

### DIFF
--- a/jxl/src/color/tf.rs
+++ b/jxl/src/color/tf.rs
@@ -99,6 +99,32 @@ pub fn srgb_to_linear(samples: &mut [f32]) {
     }
 }
 
+/// Converts the linear samples with the BT.709 transfer curve.
+pub fn linear_to_bt709(samples: &mut [f32]) {
+    for s in samples {
+        let a = s.abs();
+        *s = if a <= 0.018 {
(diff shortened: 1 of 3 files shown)


jxl/src/color/tf.rs
```diff
@@ -138,6 +164,19 @@ mod test {
...
 
+    #[test]
+    fn bt709_roundtrip_arb() {
+        arbtest::arbtest(|u| {
+            let samples: Vec<f32> = arb_samples(u)?;
+            let mut output = samples.clone();
+
+            linear_to_bt709(&mut output);
+            bt709_to_linear(&mut output);
+            assert_all_almost_eq!(&output, &samples, 5e-6);
```
tirr-c, line 175: I'm not sure about the max error here...

veluca93: A good way to find out is to run tests in release mode with ARBTEST_BUDGET_MS=100000 and see if it fails :-)

veluca93 (approved):

tirr-c:
There's a concern that using BT.709 OETF might not be the best choice, but I'm merging this as is for now.

PR #86, opened 2025-01-05
Implement HLG transfer functions

tirr-c:
Implements HLG OETF and OOTF (and their inverses).


jxl/src/color/tf.rs
```diff
@@ -46,6 +46,10 @@ const PQ_INV_EOTF_P_SMALL: [f32; 5] = [
 const PQ_INV_EOTF_Q_SMALL: [f32; 5] =
     [3.371868e1, 1.477719e3, 1.608477e4, -4.389884e4, -2.072546e5];
 
+const HLG_A: f64 = 0.17883277;
```
veluca93, line 49: wdyt of putting the constants closer to the functions?
(and similarly for the PQ ones, I guess)

veluca93 (approved):

jxl/src/color/tf.rs
tirr-c: Sure, moved the constants.

PR #89, opened 2025-01-09
Add stage spot

inflation:


jxl/src/render/stages/spot.rs
```diff
@@ -0,0 +1,118 @@
...
+    // `row` should only contain color channels and the spot channel.
+    fn process_row_chunk(
+        &mut self,
+        _position: (usize, usize),
+        xsize: usize,
+        row: &mut [&mut [f32]],
+    ) {
+        let [row_r, row_g, row_b, row_s] = row else {
+            panic!(
+                "incorrect number of channels; expected 3, found {}",
```
veluca93, line 45: 4 :-)

veluca93 (approved):

jxl/src/render/stages/spot.rs
inflation: wow that's a quick catch

PR #91, opened 2025-01-19
Add `ApplyTransferStage`

tirr-c:
Corresponds to `from_linear` of libjxl.

### DIFF
--- a/jxl/src/render/stages/apply_transfer.rs
+++ b/jxl/src/render/stages/apply_transfer.rs
@@ -1,0 +1,261 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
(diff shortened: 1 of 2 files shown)


jxl/src/render/stages/apply_transfer.rs
```diff
@@ -0,0 +1,261 @@
...
+// license that can be found in the LICENSE file.
+
+use crate::color::tf;
+use crate::headers::color_encoding::CustomTransferFunction;
+use crate::render::{RenderPipelineInPlaceStage, RenderPipelineStage};
+
+/// Apply transfer function to display-referred linear color samples.
+#[derive(Debug)]
+pub struct ApplyTransferStage {
+    first_channel: usize,
```
veluca93, line 13: (I *think* this will be always 0, but fine to leave as is)

veluca93 (approved):

PR #92, opened 2025-01-22
Implement a first version of modular channel decoding.

veluca93:
Does not support the Weighted predictor or prev-channel properties yet, and is not optimized.

### DIFF
--- a/jxl/src/frame/modular/transforms.rs
+++ b/jxl/src/frame/modular/transforms.rs
@@ -463,11 +463,30 @@ pub fn make_grids(
     // Create grids.
     for g in buffer_info.iter_mut() {
         g.buffer_grid = get_grid_indices(g.grid_kind)
-            .map(|_| ModularBuffer {
-                data: None,
-                auxiliary_data: None,
-                remaining_uses: 0,
-                used_by_transforms: vec![],
+            .map(|(x, y)| {
+                let chan_size = g.info.size;
+                let size = match g.grid_kind {
+                    ModularGridKind::None => chan_size,
+                    ModularGridKind::Lf | ModularGridKind::Hf => {
+                        let group_dim = if g.grid_kind == ModularGridKind::Lf {
+                            frame_header.lf_group_dim()
+                        } else {
+                            frame_header.group_dim()
+                        };
+                        let dx = group_dim >> g.info.shift.unwrap().0;
+                        let bx = x as usize * dx;
+                        let dy = group_dim >> g.info.shift.unwrap().1;
+                        let by = y as usize * dy;
+                        ((chan_size.0 - bx).min(dx), (chan_size.1 - by).min(dy))
+                    }
+                };
+                ModularBuffer {
+                    data: None,
+                    auxiliary_data: None,
+                    remaining_uses: 0,
+                    used_by_transforms: vec![],
+                    size,
+                }
             })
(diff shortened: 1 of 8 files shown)


jxl/src/frame/modular/transforms.rs
```diff
@@ -463,11 +463,32 @@ pub fn make_grids(
...
-                used_by_transforms: vec![],
+            .map(|(x, y)| {
+                let chan_size = g.info.size;
+                let size = match g.grid_kind {
+                    ModularGridKind::None => chan_size,
+                    ModularGridKind::Lf => {
+                        let dx = frame_header.lf_group_dim() >> g.info.shift.unwrap().0;
+                        let bx = x as usize * dx;
+                        let dy = frame_header.lf_group_dim() >> g.info.shift.unwrap().1;
+                        let by = y as usize * dy;
```
mo271, line 474: I suppose one could write a function like
```
fn get_group_dim_for_grid_kind(grid_kind: ModularGridKind, frame_header: &FrameHeader) -> usize {
    match grid_kind {
        ModularGridKind::Lf => frame_header.lf_group_dim(),
        ModularGridKind::Hf => frame_header.group_dim(),
        _ => panic!()
    }
}
```
and then avoid having two almost identical branches here. Not sure if this makes is less readable....

jxl/src/frame/modular/tree.rs
```diff
@@ -167,4 +177,202 @@ impl Tree {
...
+        buffers: &mut [ModularBufferInfo],
+        buffer_indices: &[usize],
+        index: usize,
+        grid_index: usize,
+        wp_state: &mut WeightedPredictorState,
+        x: usize,
+        y: usize,
+        property_buffer: &mut [i32; 256],
+    ) -> PredictionResult {
+        let get_pixel = |x: usize, y: usize| -> i32 {
```
mo271, line 213: I think one could make the code below even more readable, if one that function was name `get_pixel_or_default` and a `condition` and `default` argument. Then one could save the last else {} in every case below, no?
e.g. the first calls would then be 
```
left = get_pixel_or_default(x - 1, y, x > 0, get_pixel_or_default(x, y - 1, y > 0, 0)
top = get_pixel_or_default(x, y - 1, y > 0, left)
```

mo271 (approved):
in general there are a few places where `unwrap` is used, I assume you have considered handling the potential panics there, and decided it is not needed.

jxl/src/frame/modular/tree.rs
mo271: ..it is also fine as is

veluca93:
> in general there are a few places where `unwrap` is used, I assume you have considered handling the potential panics there, and decided it is not needed.

I believe all of those only trigger in case of bugs...

jxl/src/frame/modular/transforms.rs
veluca93: Did something slightly different, PTAL

jxl/src/frame/modular/tree.rs
veluca93: not sure it would be more readable...

jxl/src/frame/modular/transforms.rs
mo271: nice!

PR #96, opened 2025-01-27
Add stage epf

inflation:
Just a quick POC for easier stage 0. It seems to have a lot of offset manipulations so fingers crossed 🤞 for the compiler to get rid of the bound checks. 

It's also awkward to work with the `row` type since it's behind a large `&mut[Channel]` reference.

### DIFF
--- a/jxl/src/render/stages/epf.rs
+++ b/jxl/src/render/stages/epf.rs
@@ -1,0 +1,132 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+use std::sync::Arc;
+
+use crate::{
+    image::Image,
+    render::{RenderPipelineInOutStage, RenderPipelineStage},
+    BLOCK_DIM, MIN_SIGMA, SIGMA_PADDING,
+};
+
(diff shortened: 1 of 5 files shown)


jxl/src/render/stages/epf.rs
```diff
@@ -0,0 +1,155 @@
...
+};
+
+/// 5x5 plus-shaped kernel with 5 SADs per pixel (3x3 plus-shaped). So this makes this filter a 7x7 filter.
+pub struct Epf0Stage {
+    /// Multiplier for sigma in pass 0
+    sigma_scale: f32,
+    /// (inverse) multiplier for sigma on borders
+    border_sad_mul: f32,
+    channel_scale: [f32; 3],
+    sigma: Arc<[f32]>,
```
veluca93, line 15: Why is this not an Arc<Image<f32>>?

veluca93:
> Just a quick POC for easier stage 0. It seems to have a lot of offset manipulations so fingers crossed 🤞 for the compiler to get rid of the bound checks.

I wouldn't worry about that yet, we'll speed things up after we have a complete decoder :-)

> 
> It's also awkward to work with the `row` type since it's behind a large `&mut[Channel]` reference.

How so?

inflation:
Just did some refactoring. The `row` structure is not so bad after rethinking.

veluca93 (approved):

PR #100, opened 2025-02-13
Added SplinesStage.

zond:
Also made the assert_almost_eq! macro handle relative errors.


veluca93:
Do you mind splitting this in 3 PRs? (pfm reading, math functions and splines)

jxl/src/util/test.rs
```diff
@@ -67,6 +79,17 @@ macro_rules! assert_almost_eq {
             _ => {}
         }
     };
+    ($left:expr, $right:expr, $max_error_abs:expr, $max_error_rel:expr $(,)?) => {
+        let (left_val, right_val, max_error_abs, max_error_rel) = (&$left, &$right, &$max_error_abs, &$max_error_rel);
```
veluca93, line 83: Why take a reference?
Also, I feel that there should be more concise ways to express `into` and/or `min` -- perhaps we could just decide to cast everything to f64 and call it a day, as I think this should be used just with f32 and f64

veluca93 (approved):

jxl/src/util/test.rs
```diff
@@ -48,63 +48,71 @@ impl From<JXLError> for Error {
     }
 }
 
-pub fn abs_delta<T: Num + std::cmp::PartialOrd>(left_val: T, right_val: T) -> T {
-    if left_val > right_val {
-        left_val - right_val
-    } else {
-        right_val - left_val
-    }
+pub fn safe_gt<T: std::cmp::PartialOrd>(left_val: T, right_val: T) -> bool {
```
veluca93, line 51: Please don't use `safe` for things that have nothing to do with `unsafe` code :-)

This doesn't need to be generic, right? could be done inline in a lambda or so.

veluca93 (approved):

jxl/src/util/test.rs
zond: Fixed

PR #118, opened 2025-04-23
2d idct

mo271:


jxl/src/var_dct/dct.rs
```diff
@@ -108,20 +108,28 @@ define_idct_1d!(64, 32);
 define_idct_1d!(128, 64);
 define_idct_1d!(256, 128);
 
-fn transpose<const N: usize, const M: usize>(_data: &mut [f32]) {
-    todo!();
+fn transpose<const N: usize, const M: usize>(data: &[[f32; M]]) -> Vec<[f32; N]> {
+    let mut transposed: Vec<[f32; N]> = vec![[0.0f32; N]; M];
```
veluca93, line 112: In the future we almost certainly do not want this allocation here.

veluca93 (approved):

jxl/src/var_dct/dct.rs
```diff
@@ -108,20 +108,28 @@ define_idct_1d!(64, 32);
...
+    for (i, data_row) in data.iter().enumerate() {
+        for (j, &input_element) in data_row.iter().enumerate() {
+            transposed[j][i] = input_element;
+        }
+    }
+    transposed
 }
 
-fn idct2d<const ROWS: usize, const COLS: usize>(_data: &mut [f32])
+fn idct2d<const ROWS: usize, const COLS: usize>(data: &mut [[f32; COLS]])
```
veluca93, line 121: Thinking about it, I am not sure this will work - IIRC the shape of the input block is always such that ROWS >= COLS in the bitstream

mo271: so, should we assume a flat array here?

veluca93 (approved):

jxl/src/var_dct/dct.rs
mo271: done

jxl/src/var_dct/dct.rs
mo271: changed to also use flat array here. Still doing one copy though.

PR #119, opened 2025-04-23
Patches

zond:


veluca93 (commented):
Seems reasonable to me.
I would possibly consider using fallible allocations (i.e. try_reserve) as the # of patches and patch coordinates can be somewhat high.

zond:
> Seems reasonable to me. I would possibly consider using fallible allocations (i.e. try_reserve) as the # of patches and patch coordinates can be somewhat high.

I sprinkled all the patches code with try_reserve - I basically tried to add a try_reserve for each push that didn't already have it, and when it was simple to see the number of elements we wanted to push I used that as "additional capacity".

WDYT?

jxl/src/features/patches.rs
```diff
@@ -325,6 +462,7 @@ impl PatchesDictionary {
                 positions.push(pos);
             }
 
+            ref_positions.try_reserve(1)?;
```
tirr-c, line 465: We don't need this because we've already allocated the memory using `new_with_capacity` at line 293.

jxl/src/features/patches.rs
```diff
@@ -333,11 +471,125 @@ impl PatchesDictionary {
...
+        Ok(patches_dict)
+    }
+
+    pub fn get_patches_for_row(&self, y: usize) -> Vec<usize> {
+        let mut result = vec![];
+        if self.num_patches.len() <= y || self.num_patches[y] == 0 {
+            return result;
+        }
+
+        result.reserve(self.num_patches[y]); //  Reserve space as in C++
```
tirr-c, line 495: You can use `new_with_capacity` here.

jxl/src/features/patches.rs
```diff
@@ -333,11 +471,125 @@ impl PatchesDictionary {
...
+        row: &mut [&mut [f32]],
+        position: (usize, usize),
+        xsize: usize,
+        extra_channel_info: &[ExtraChannelInfo],
+        reference_frames: &[ReferenceFrame],
+    ) {
+        let bg = row
+            .iter_mut()
+            .map(|s| s.to_vec())
+            .collect::<Vec<Vec<f32>>>();
```
tirr-c, line 550: This will allocate some buffers, not sure this would be okay... Maybe we need error checks here?

jxl/src/features/patches.rs
zond: Done. Thanks!

jxl/src/features/patches.rs
zond: Done

jxl/src/features/patches.rs
tirr-c: You can use `new_with_capacity` and `.extend_from_slice()` so that the closure returns `Result<Vec<f32>, _>`, and the iterator can be collected into `Result<Vec<T>, E>`.

zond: I just realized this function is used by the render pipeline, which doesn't handle errors.

Should I assume the allocations are small enough to not cause errors? If not, how should one handle render errors during render pipeline calls?

ping @veluca93 too

veluca93: Yes, you should assume so. Ideally, stages in the render pipeline allocate necessary temporary buffers in advance.

zond: I guess this could be done, but it would be messier, and I'll leave it as a TODO.

veluca93 (approved):

PR #120, opened 2025-04-23
Splines: take ColorCorrelationParams

sboukortt:


jxl/src/features/spline.rs
```diff
@@ -639,8 +638,12 @@ impl Splines {
...
                 &self.starting_points[index],
                 self.quantization_adjustment,
-                y_to_x,
-                y_to_b,
+                color_correlation_params.base_correlation_x
+                    + (color_correlation_params.ytox_lf as f32)
+                        / (color_correlation_params.color_factor as f32),
+                color_correlation_params.base_correlation_b
+                    + (color_correlation_params.ytob_lf as f32)
+                        / (color_correlation_params.color_factor as f32),
```
sboukortt, line 646: (Or would we prefer to have this as methods of `ColorCorrelationParams`?)

sboukortt:
Woops, clearly I did not test this sufficiently thoroughly.

jxl/src/features/spline.rs
veluca93: I think that would be better

veluca93 (approved):

PR #122, opened 2025-04-23
Initial version of transform application.

veluca93:


sboukortt (approved):

jxl/src/frame/modular/transforms.rs
```diff
@@ -398,6 +317,32 @@ pub struct TransformStepChunk {
...
 }
 
+impl TransformStepChunk {
+    // Mark that one dependency of this transform is ready, and potentially runs the transform,
+    // returning the new buffers that are now ready.
+    pub fn dep_ready(
+        &mut self,
+        buffer_info: &mut [ModularBufferInfo],
+    ) -> Result<Vec<(usize, usize)>> {
+        self.incomplete_deps = self.incomplete_deps.checked_sub(1).unwrap();
```
mo271, line 327: will this unwrap panic sometimes, or is it certain that `self.incomplete_deps` is never `0`?

jxl/src/frame/modular.rs
```diff
@@ -285,6 +309,42 @@ impl FullModularImage {
...
+                    bi.channel_id,
+                    g,
+                    bi.buffer_grid[grid].data.as_ref().unwrap(),
+                )?;
+                bi.buffer_grid[grid].mark_used();
+            }
+            Ok(())
+        };
+
+        let mut new_ready_transform_chunks = vec![];
```
mo271, line 328: is `vec![]` more idiomatic than `Vec::new()` (just a learning question for me)

jxl/src/frame/modular/transforms/rct.rs
```diff
@@ -0,0 +1,88 @@
...
+    for i in 0..3 {
+        assert_eq!(buffers[buf_in[0]].grid_kind, buffers[buf_in[i]].grid_kind);
+        assert_eq!(buffers[buf_in[0]].grid_kind, buffers[buf_out[i]].grid_kind);
+        assert_eq!(buffers[buf_in[0]].info.size, buffers[buf_in[i]].info.size);
+        assert_eq!(buffers[buf_in[0]].info.size, buffers[buf_out[i]].info.size);
+        assert!(buffers[buf_in[i]].buffer_grid[grid].data.is_some());
+        assert!(buffers[buf_out[i]].buffer_grid[grid].data.is_none());
+    }
+    if op != RctOp::YCoCg && perm != RctPermutation::Rgb {
+        unimplemented!("non-YCoCg not implemented");
```
mo271, line 43: the condition and the error message don't quite agree. Should this be 
"non-YCoCg with non rgb perm not implemented" or something?

jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -0,0 +1,108 @@
...
+            horizontal: false,
+            ..sp
+        });
+    }
+
+    const MAX_FIRST_PREVIEW_SIZE: usize = 8;
+
+    let sp = SqueezeParams {
+        begin_channel: num_meta_channels as u32,
+        num_channels: nc as u32,
```
mo271, line 77: should this be `nc - num_meta_channels` if it only starts at ben_channel?

mo271 (approved):
LGTM

jxl/src/frame/modular/transforms.rs
veluca93: It should never be 0 unless there's a logic bug somewhere.

jxl/src/frame/modular.rs
veluca93: I believe so...

jxl/src/frame/modular/transforms/rct.rs
veluca93: it could be, but I'm not sure it's _too_ important...

jxl/src/frame/modular/transforms/squeeze.rs
veluca93: (I only moved this code)
I *think* this was correct, I think it matches the C++ code and Jon checked it, but of course my confidence level is low...

PR #124, opened 2025-04-24
add DCT

mo271:
Analogue to the IDCT implementation.


veluca93 (approved):
Do we also want to test that idct(dct) does the expected thing?

mo271:
yes, but in a different pull request

PR #126, opened 2025-04-24
Made the blending code work with any inner arrays that can be converted to slices of floats.

zond:


veluca93 (commented):
Should you also delete the ndarray file?

zond:
> Should you also delete the ndarray file?

Oh, didn't it get deleted? I thought I did.

Anyway, it gets added back in the next PR, so..

PR #130, opened 2025-04-25
Incomplete palette implementation

sboukortt:


jxl/src/frame/modular/transforms/palette.rs
```diff
@@ -2,14 +2,244 @@
...
+// SMALL_CUBE ** 3
+const LARGE_CUBE_OFFSET: usize = SMALL_CUBE * SMALL_CUBE * SMALL_CUBE;
+const IMPLICIT_PALETTE_SIZE: usize = LARGE_CUBE_OFFSET + LARGE_CUBE * LARGE_CUBE * LARGE_CUBE;
+
+fn scale<const DENOM: usize>(value: usize, bit_depth: usize) -> i32 {
+    // return (value * ((1 << bit_depth) - 1)) / denom;
+    // We only call this function with SMALL_CUBE or LARGE_CUBE - 1 as denom,
+    // allowing us to avoid a division here.
+    const {
+        assert!(DENOM == 4, "denom must be 4");
```
veluca93, line 32: I'm still not convinced we need the const generic here :-) 

(also, stale comment)

veluca93 (approved):

jxl/src/frame/modular/transforms/palette.rs
sboukortt: Is it? Do you mean because of the case of `DENOM`?

sboukortt: Eh, it explains why we `>> 2` to start with. I think I would rather keep it.

sboukortt: (I have updated it to at least use the same naming convention as the actual variable.)

PR #131, opened 2025-04-25
Added tests to the blending code.

zond:


jxl/src/features/blending.rs
```diff
@@ -296,13 +323,1855 @@ pub fn perform_blending(
...
+
+#[cfg(test)]
+mod tests {
+    fn clamp(x: f32) -> f32 {
+        x.clamp(0.0, 1.0)
+    }
+
+    const MAX_DELTA: f32 = 1e-6;
+
+    mod perform_alpha_blending_layers_tests {
```
veluca93, line 350: Not sure I like the nested test mods...

jxl/src/features/blending.rs
```diff
@@ -296,13 +323,1855 @@ pub fn perform_blending(
...
+        use super::{super::*, *};
+        use crate::util::test::assert_all_almost_eq;
+
+        #[test]
```
veluca93, line 354: Remember to `use test_log::test`!

veluca93 (approved):
@sboukortt want to give this a look? ;-)

jxl/src/features/blending.rs
zond: Are you OK enough with them? There are so many tests and so much code that it felt meaningful to separate them a bit, to avoid risking naming collisions and keep things separated.

jxl/src/features/blending.rs
zond: Done

zond:
I'll just submit this - I can change stuff later if anyone has issues.

jxl/src/features/blending.rs
zond: I'll change it later if necessary - submitting now to get on with this :)

PR #132, opened 2025-04-25
Refactor modular buffer interfaces to be more (re)usable.

veluca93:
First step towards local transforms; also simplifies transform implementation.

### DIFF
--- a/jxl/src/frame/modular/decode.rs
+++ b/jxl/src/frame/modular/decode.rs
@@ -3,19 +3,19 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+use std::ops::DerefMut;
+
 use crate::{
     bit_reader::BitReader,
     entropy_coding::decode::Reader,
     error::{Error, Result},
     frame::quantizer::NUM_QUANT_TABLES,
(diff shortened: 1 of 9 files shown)


jxl/src/frame/modular/decode.rs
```diff
@@ -102,18 +89,31 @@ fn decode_modular_channel(
...
-    buffer_indices: &[usize],
+// This function will decode a header and apply local transforms if a header is not given.
+// The intended use of passing a header is for the DcGlobal section.
+pub fn decode_modular_subbitstream(
+    buffers: &mut [impl DerefMut<Target = Image<i32>>],
+    channel_indices: &[usize],
     grid_index: usize,
     stream_id: usize,
-    header: &GroupHeader,
+    header: Option<GroupHeader>,
```
sboukortt, line 99: (Not &?)

sboukortt (approved):

jxl/src/frame/modular/decode.rs
veluca93: No, & makes things more annoying for the `if` and doesn't actually help callers...

szabadka (approved):

PR #135, opened 2025-04-28
Add some more debug output and utilities.

veluca93:
Also fixes some incorrect accounting of bits in BitReader::split_at.

### DIFF
--- a/jxl/src/bit_reader.rs
+++ b/jxl/src/bit_reader.rs
@@ -3,7 +3,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-use crate::error::Error;
+use std::fmt::Debug;
+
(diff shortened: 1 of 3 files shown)


jxl/src/bit_reader.rs
```diff
@@ -167,9 +187,11 @@ impl<'a> BitReader<'a> {
             // Prevent the returned bitreader from over-reading.
             ret.data = &ret.data[..n - bytes_in_buf];
         } else {
-            ret.bits_in_buf -= n * 8;
```
szabadka, line 170: it should be mentioned in the commit message that this also fixes some bugs

szabadka (approved):

jxl/src/bit_reader.rs
veluca93: Good point...

PR #142, opened 2025-04-30
Pass shift and meta information down to all steps of modular decoding.

veluca93:


jxl/src/image.rs
```diff
@@ -268,6 +268,13 @@ impl<T: ImageDataType> Image<T> {
             image: self,
         }
     }
+
+    pub fn try_clone(&self) -> Result<Self> {
```
mo271, line 272: do we also want to add `clone` if it can never fail?

mo271 (approved):

jxl/src/image.rs
veluca93: Ugh, this implementation is not what I wanted to write... now it can fail.

jxl/src/frame/modular.rs
```diff
@@ -86,15 +86,50 @@ impl ModularGridKind {
...
+    // Shift of the channel (None if this is a meta-channel).
+    shift: Option<(usize, usize)>,
+}
+
+impl ModularChannel {
+    fn new(size: (usize, usize)) -> Result<Self> {
+        Self::new_with_shift(size, Some((0, 0)))
+    }
+
+    fn new_with_shift(size: (usize, usize), shift: Option<(usize, usize)>) -> Result<Self> {
```
sboukortt, line 106: Is this ever going to be called with `shift = None`? Would it make sense to take a naked tuple directly?

sboukortt (approved):

jxl/src/frame/modular.rs
veluca93: Not right now, but I expect it might be in the (near) future

PR #143, opened 2025-04-30
Implement the logic to apply local transforms.

veluca93:
Also finish implementing RCT, to allow for manual testing.


jxl/src/frame/modular/decode.rs
```diff
@@ -91,17 +91,27 @@ pub fn decode_modular_subbitstream(
...
+            // Note: reassigning to `buffers` here convinces the borrow checker that the borrow of
+            // `buffer_storage` ought to outlive `buffers[..]`'s lifetime, which obviously breaks
+            // applying transforms later.
+            let new_bufs;
+            (new_bufs, transform_steps) =
+                meta_apply_local_transforms(buffers, &mut buffer_storage, &h.transforms)?;
+            (h, new_bufs)
+        } else {
+            (h, buffers)
         }
```
sboukortt, line 112: Out of curiosity, why not a `match`?

(Something like:

```c++
    let (header, mut buffers) = match header {
        Some(h) => (h, buffers),
        None => {
            let h = GroupHeader::read(br)?;
            if !h.transforms.is_empty() {
                // Note: reassigning to `buffers` here convinces the borrow checker that the borrow of
                // `buffer_storage` ought to outlive `buffers[..]`'s lifetime, which obviously breaks
                // applying transforms later.
                let new_bufs;
                (new_bufs, transform_steps) =
                    meta_apply_local_transforms(buffers, &mut buffer_storage, &h.transforms)?;
                (h, new_bufs)
            } else {
                (h, buffers)
            }
        }
```
)

jxl/src/frame/modular/transforms/apply.rs
```diff
@@ -134,3 +141,484 @@ impl TransformStepChunk {
...
+                            return Err(Error::TooManySqueezes);
+                        }
+                        if horizontal {
+                            Some((shift.0 + 1, shift.1))
+                        } else {
+                            Some((shift.0, shift.1 + 1))
+                        }
+                    } else {
+                        None
+                    };
```
sboukortt, line 235: Isn’t this:

```suggestion
                    let new_shift = chan.shift.map(|shift| {
                        if shift.0 > 30 || shift.1 > 30 {
                            return Err(Error::TooManySqueezes);
                        }
                        if horizontal {
                            Some((shift.0 + 1, shift.1))
                        } else {
                            Some((shift.0, shift.1 + 1))
                        }
                    });
```

sboukortt (approved):

jxl/src/frame/modular/decode.rs
veluca93: I don't think it changes too much...

jxl/src/frame/modular/transforms/apply.rs
veluca93: No, the `return` inside `map` doesn't quite work right :-) this would need a `transpose()` or some other mildly painful additional step

PR #144, opened 2025-05-02
Wire up the render pipeline.

veluca93:
Also change the render pipeline interface to accept channel-by-channel data, and fix a few uncovered bugs.

### DIFF
--- a/jxl/src/headers/frame_header.rs
+++ b/jxl/src/headers/frame_header.rs
@@ -434,8 +434,8 @@ pub struct FrameHeader {
     #[default(Extensions::default())]
     extensions: Extensions,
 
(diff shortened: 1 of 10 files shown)


jxl/src/headers/frame_header.rs
```diff
@@ -434,8 +434,8 @@ pub struct FrameHeader {
     #[default(Extensions::default())]
     extensions: Extensions,
 
-    // The following 4 fields are not actually serialized, but just used as variables to help with
```
mo271, line 437: good thinking, removing the count, as it is cumbersome to keep in sync...

mo271 (approved):

PR #145, opened 2025-05-05
use width in lz77 decoding

mo271:


jxl/src/frame/modular/decode.rs
```diff
@@ -134,7 +134,7 @@ pub fn decode_modular_subbitstream(
         );
     }
 
-    let mut reader = tree.histograms.make_reader(br)?;
+    let mut reader = tree.histograms.make_reader_with_width(br, buffers.len())?;
```
veluca93, line 137: This is most certainly not the correct width to pass, right?

mo271: What is the correct width and how to access it here?

veluca93: See the loop that computes distance_multiplier here: https://github.com/libjxl/libjxl/blob/5f9ef79e5528ab8ec0b59c627e38d9fac3e97c68/lib/jxl/modular/encoding/encoding.cc#L572

mo271: ok, so it is the max of the width of all channels. 
This info on all channels is not available inside ` decode_modular_subbitstream`, right?
So do we need to pass whit max_width down to this function?

veluca93: No, it should be the maximum width of all the buffers.

mo271: make sense, done

veluca93 (approved):

PR #146, opened 2025-05-05
LF dequantization (not yet called)

sboukortt:


jxl/src/frame/modular.rs
```diff
@@ -440,6 +444,120 @@ impl FullModularImage {
     }
 }
 
+#[allow(dead_code)]
+#[allow(clippy::too_many_arguments)]
+fn dequant_lf<'a>(
+    r: Rect,
+    lf: (
```
veluca93, line 451: why not a [_; 3]?

sboukortt: It seems to make it a pain to mutably borrow all three at the same time like on lines 481..483.

veluca93: You can `let [a, b, c] = array ` :-)

sboukortt: Oh :D thanks.

PR #147, opened 2025-05-05
Call dequant_lf

sboukortt:


szabadka (approved):

jxl/src/frame/modular.rs
```diff
@@ -573,6 +582,18 @@ pub fn decode_vardct_lf(
         br,
     )?;
     // TODO(szabadka): Generate the f32 pixels of the LF image.
```
veluca93, line 584: Is this comment still true?

(as a side note, we might in the future want to avoid creating a full LF image)

PR #150, opened 2025-05-07
Weighted predictor

zond:


jxl/src/frame/modular/predict.rs
```diff
@@ -152,17 +155,156 @@ impl Predictor {
     }
 }
 
+const NUM_PREDICTORS: usize = 4;
+const PRED_EXTRA_BITS: i64 = 3;
+const PREDICTION_ROUND: i64 = ((1 << PRED_EXTRA_BITS) >> 1) - 1;
+const DIVLOOKUP: [u32; 64] = [
```
eustas, line 161: Please description on how this magic constants are obtained (even better, calculate them in compile time).

Original comment was `for (int i = 0; i < 64; i++) divlookup[i] = (1 << 24) / (i + 1);`

zond: I have no idea why they are what they are - I just copied them from libjxl.

Added a comment explaining that.

zond:
@veluca93 added a test, ready for review.

jxl/src/frame/modular/predict.rs
```diff
@@ -152,17 +155,247 @@ impl Predictor {
...
+        for (i, pred_err) in self.pred_errors.iter_mut().enumerate() {
+            let err =
+                (((self.prediction[i] - val).abs() + PREDICTION_ROUND) >> PRED_EXTRA_BITS) as u32;
+            pred_err[cur_row + pos.0] = err;
+            let idx = prev_row + pos.0 + 1;
+            pred_err[idx] = pred_err[idx].wrapping_add(err);
+        }
+    }
+
+    pub fn predict_and_property<const COMPUTE_PROPERTIES: bool>(
```
veluca93, line 238: Can you remove the generic and always compute properties for now?

jxl/src/frame/modular/predict.rs
```diff
@@ -152,17 +155,247 @@ impl Predictor {
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::headers::modular::{GroupHeader, WeightedHeader};
+
+    use super::{PredictionData, WeightedPredictorState};
+
+    struct SimpleRandom {
```
veluca93, line 321: Please add `rand` and say `rand_xoshiro` as a dev dependency instead.

jxl/src/frame/modular/predict.rs
zond: But I have no idea how the constants are obtained, all I can do is refer to the libjxl codebase.

jxl/src/frame/modular/predict.rs
zond: I generated the seed values for the libjxl golden data using this super simple method, and it doesn't feel useful to redo the work in libjxl _and_ add another dependency just for that.

jxl/src/frame/modular/predict.rs
zond: Oh, you mean copy the comment from libjxl about the DIVLOOKUP constant in particular. Sure, done.

jxl/src/frame/modular/predict.rs
```diff
@@ -152,17 +155,244 @@ impl Predictor {
...
+        res
+    }
+
+    #[test]
+    fn predict_and_update_errors() {
+        let mut rng = SimpleRandom::new();
+        let header = GroupHeader {
+            use_global_tree: false,
+            wp_header: WeightedHeader {
+                all_default: true,
```
veluca93, line 362: This seems inconsistent.

Also, these are clearly invalid values, we should be specifying values could be read from a bitstream (so up to 5 bits for p..., and up to 4 bits for w...), as otherwise we are testing behaviour we don't care about.

jxl/src/frame/modular/predict.rs
```diff
@@ -152,17 +155,244 @@ impl Predictor {
...
+            header,
+        }
+    }
 
-impl WeightedPredictorState {
-    pub fn new(_header: &GroupHeader) -> Self {
-        // TODO(veluca): implement the weighted predictor.
-        Self
+    #[allow(dead_code)]
+    pub fn update_errors(&mut self, correct_val: i64, pos: (usize, usize), xsize: usize) {
```
veluca93, line 222: Can you add the call to `update_errors` too?

jxl/src/frame/modular/predict.rs
```diff
@@ -152,17 +155,247 @@ impl Predictor {
...
+            self.error[cur_row + pos.0 - 1]
+        };
+        let te_n = self.error[pos_n];
+        let te_nw = self.error[pos_nw];
+        let sum_wn = te_n + te_w;
+        let te_ne = self.error[pos_ne];
+
+        let p = if COMPUTE_PROPERTIES {
+            let abs_te_w = te_w.abs();
+            if te_n.abs() > abs_te_w {
```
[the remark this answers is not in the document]
zond: Oh, nice catch - I somehow assumed _else_ where there weren't any.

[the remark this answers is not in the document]
zond: Better now?

jxl/src/frame/modular/predict.rs
zond: I'm not sure where it should be called - could you point it out?

zond: Done

jxl/src/frame/modular/predict.rs
veluca93: It matters, because it might not *stay* the same in the future when we do optimizations.
Also, Rust in debug mode panics on overflow, which these values would certainly cause at some point or another.

veluca93: It would help, yes.

zond: Done

jxl/src/frame/modular/predict.rs
zond: Ping

veluca93 (approved):

PR #108, opened 2025-04-04
Begin with dct implementation

mo271:

PR #151, opened 2025-05-07
use vectors in slow dct

mo271:
https://github.com/libjxl/jxl-rs/pull/108#discussion_r2054007056


mo271:
once reviewed, I plan to add the analogue changes to the slow idct implementation here.

veluca93 (approved):

PR #152, opened 2025-05-07
remove useless .take(foo)

mo271:


mo271:
those likely come from poorly interpreted clippy recommendations...

veluca93 (approved):

PR #153, opened 2025-05-08
test dct(idct(..)) scaling and vice versa

mo271:


mo271:
I considered using arbtest here, but decided the random values I picked might be sufficient.

veluca93 (approved):

PR #154, opened 2025-05-08
IO error should not cause panic

eustas:


veluca93 (commented):
You can also return `Result` from `main` and just use `?`, although I guess it doesn't really matter here :-)

veluca93 (approved):
You can also return `Result` from `main` and just use `?`, although I guess it doesn't really matter here :-)

eustas:
Thanks.

PR #156, opened 2025-05-08
Arbtest post merge

mo271:


mo271:
tested here:
https://github.com/mo271/jxl-rs/actions/runs/14908495558/job/41876634055

PR #170, opened 2025-05-14
Support block context maps with quant thresholds.

szabadka:
Drive-by: fix number of contexts when reading HF histograms.

### DIFF
--- a/jxl/src/frame/group.rs
+++ b/jxl/src/frame/group.rs
@@ -46,18 +46,21 @@ pub fn decode_vardct_group(
     debug!(?block_rect);
     let transform_map = hf_meta.transform_map.as_rect();
     let transform_map_rect = transform_map.rect(block_rect)?;
+    let raw_quant_map = hf_meta.raw_quant_map.as_rect();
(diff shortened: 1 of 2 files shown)


jxl/src/frame/group.rs
```diff
@@ -46,18 +46,22 @@ pub fn decode_vardct_group(
...
+    let raw_quant_map_rect = raw_quant_map.rect(block_rect)?;
     let mut num_nzeros: [Image<u32>; 3] = [
         Image::new(block_rect.size)?,
         Image::new(block_rect.size)?,
         Image::new(block_rect.size)?,
     ];
     let block_context_map = lf_global.block_context_map.as_mut().unwrap();
-    if block_context_map.num_lf_contexts > 1 || !block_context_map.qf_thresholds.is_empty() {
+    if block_context_map.num_lf_contexts > 1 {
+        // || !block_context_map.qf_thresholds.is_empty() {
```
veluca93, line 58: Remove the comment?

veluca93 (approved):

jxl/src/frame/group.rs
szabadka: oops, done

PR #171, opened 2025-05-14
Remove VarDCT exception from frame reading tests.

szabadka:


veluca93 (approved):

szabadka:
Since we can decode WP predictor and we do not have reference properties in any of the tests, I removed all exceptions now. Going forward, we can add new test streams for each new supported feature.

PR #173, opened 2025-05-15
add test files from conformance tests

mo271:


veluca93 (commented):
I think I would prefer to create a few separate PRs:
 - one that adds the files that work today
 - and one for each group of files that fails in the same way
 
In particular, I think we shouldn't introduce logic to special-case some images to not give errors.

mo271:
ok, your first point is now in https://github.com/libjxl/jxl-rs/pull/176

to the second point just so I understand: Do you proposed to have an open pr for each sort of failure until the reason for it is fixed and the tests pass? (i.e. not use `ignore` at all ?)

veluca93:
> ok, your first point is now in #176
> 
> to the second point just so I understand: Do you proposed to have an open pr for each sort of failure until the reason for it is fixed and the tests pass? (i.e. not use `ignore` at all ?)

Indeed.

PR #174, opened 2025-05-15
Computing quant tables

zond:
Completely untested, will add tests in next CL.


jxl/src/frame/quant_weights.rs
```diff
@@ -829,8 +864,8 @@ impl DequantMatrices {
             ]),
         }
     }
-    pub fn library() -> &'static [QuantEncoding; NUM_PREDEFINED_TABLES * NUM_QUANT_TABLES] {
-        static QUANTS: OnceLock<[QuantEncoding; NUM_PREDEFINED_TABLES * NUM_QUANT_TABLES]> =
+    pub fn library() -> &'static [QuantEncoding; NUM_PREDEFINED_TABLES * QuantTable::VALUES.len()] {
```
veluca93, line 867: Since `NUM_PREDEFINED_TABLES` is 1, I would get rid of the constant (the C++ has it because at some point we thought we might want it to be > 1)

I also think `NUM_QUANT_TABLES` is more readable; probably better to leave the constant around and `const _: () = assert!(QuantTable::VALUES.len() == NUM_QUANT_TABLES;` somewhere.

veluca93 (approved):

jxl/src/frame/quant_weights.rs
zond: Done

zond: Done

PR #175, opened 2025-05-15
Move output file writing from debug_tools to the cli.

szabadka:


jxl_cli/src/main.rs
```diff
@@ -27,6 +30,13 @@ fn parse_jxl_codestream(data: &[u8]) -> Result<(), jxl::error::Error> {
         let r = read_icc(&mut br)?;
         println!("found {}-byte ICC", r.len());
```
mo271, line 31: not super related, but I suppose we can at some point not print so much information here, but rather in jxl-inspect instead.

jxl/src/image.rs
```diff
@@ -498,8 +498,7 @@ impl<'a, T: ImageDataType> ImageRectMut<'a, T> {
     }
 }
 
-#[cfg(feature = "debug_tools")]
-pub mod debug_tools {
+pub mod output_tools {
```
veluca93, line 501: Move this to `enc`?

Do we even need the `debug_tools` feature anymore?

(in the long term, this functionality should probably go in a separate crate...)

jxl_cli/src/main.rs
```diff
@@ -59,22 +69,60 @@ fn parse_jxl_codestream(data: &[u8]) -> Result<(), jxl::error::Error> {
...
 }
 
 #[derive(Parser)]
 struct Opt {
     input: PathBuf,
+    output: PathBuf,
 }
 
-fn main() {
+fn main() -> ExitCode {
```
veluca93, line 125: You can also return a `Result<()>` from `main`.

veluca93 (approved):

mo271 (approved):

jxl_cli/src/main.rs
szabadka: when we add clap for CLI parsing, we can have a --quiet flag to control these as in libjxl

jxl/src/image.rs
szabadka: Moved these to enc, also removed the debug_tools feature

PR #176, opened 2025-05-15
add passing test files from conformance tests 

mo271:


jxl_macros/src/lib.rs
```diff
@@ -5,13 +5,14 @@
...
 
-use std::{fs, path::Path};
+use std::path::Path;
 
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use proc_macro_error::{abort, proc_macro_error};
 use quote::quote;
 use syn::{parse_macro_input, DeriveInput, Ident, Meta};
+use walkdir::WalkDir;
```
veluca93, line 15: Do we need to add a dependency?

mo271: No, but then I would re-implement walking the dir using `fs::read_dir`, or is there something like `fs::read_all_dir` that I could use? 
Do you propose that I walk the dir manually instead of having the extra dependency?

veluca93: I think I'd prefer that -- or alternatively, we can list the not-many directories that contain testdata manually :-)

mo271: done, spelling out that one dir manually...

PR #177, opened 2025-05-15
Cleaned up the quant table computation.

zond:


zond:
Something strange happened here - in the changes it looks like e.g. get_quant_weights didn't exist before this PR, but it was called from compute_quant_weights, and the code compiled. I'm confused.

jxl/src/frame/quant_weights.rs
```diff
@@ -136,7 +134,7 @@ impl QuantEncoding {
                 if predefined as usize >= 1 {
```
veluca93, line 134: Remove this logic too? The error can never be triggered anyway.

veluca93: Reading 0 bits from the bitstream does nothing, please remove that.

zond: Oh, it's actually really 0 bits?

I thought it read something.

Removing it.

PR #178, opened 2025-05-15
Pass the render pipeline callback to the vardct group decoder.

szabadka:
Since in vardct mode we pad the group size to 8x8 blocks, this requires a modification to ImageRect::copy_from to allow the source image to be bigger.

### DIFF
--- a/jxl/src/frame.rs
+++ b/jxl/src/frame.rs
@@ -16,7 +16,7 @@ use crate::{
         permutation::Permutation,
         FileHeader,
     },
-    image::Image,
+    image::{Image, Rect},
     render::{
         stages::SaveStage, RenderPipeline, RenderPipelineBuilder, SimpleRenderPipeline,
         SimpleRenderPipelineBuilder,
@@ -27,7 +27,7 @@ use block_context_map::BlockContextMap;
 use coeff_order::decode_coeff_orders;
 use color_correlation_map::ColorCorrelationParams;
 use group::decode_vardct_group;
-use modular::{decode_hf_metadata, decode_vardct_lf};
+use modular::{decode_hf_metadata, decode_vardct_lf, fill_in_modular_rect};
 use modular::{FullModularImage, ModularStreamId, Tree};
 use quant_weights::DequantMatrices;
 use quantizer::LfQuantFactors;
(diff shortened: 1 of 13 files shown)


szabadka:
I realized that VarDCT should fill in the render pipeline in f32 as in libjxl and for that to work the save stage and the image writers need to support it as well, so it will require some more plumbing.

jxl/src/frame.rs
```diff
@@ -480,32 +480,49 @@ impl Frame {
...
-        let mut pass_to_pipeline = |chan: usize, g: usize, img: &Image<i32>| {
+        let mut convert_and_pass_to_pipeline = |chan: usize, g: usize, img: &Image<i32>| {
             // TODO(veluca): figure out pass count and check group IDs. Also, investigate whether
             // reusing the `img` buffer would be possible.
             self.render_pipeline
                 .as_mut()
                 .unwrap()
-                .fill_input_channels(&[chan], g, 1, |rects| rects[0].copy_from(img.as_rect()))
+                .fill_input_channels(&[chan], g, 1, |rects| {
+                    fill_in_modular_rect(&mut rects[0], &img.as_rect())
```
veluca93, line 513: Ideally, the idea here would be to pass `i32` channels to the pipeline, and have the pipeline convert to f32 where needed.
That functionality should be supported, but the stage to do this might not be present yet.

jxl/src/image.rs
```diff
@@ -442,18 +442,18 @@ impl<'a, T: ImageDataType> ImageRectMut<'a, T> {
 
     #[instrument(skip_all)]
     pub fn copy_from(&mut self, other: ImageRect<'_, T>) -> Result<()> {
```
veluca93, line 444: Could also make the rect smaller instead of changing this :-)

jxl/src/frame.rs
szabadka: We only have a dummy render pipeline now that just stitches the groups together. I added a TODO to move this to the render pipeline.

jxl/src/image.rs
szabadka: Done.

veluca93 (approved):

PR #180, opened 2025-05-15
conformance fail: not yet implemented: Unsupported block context map in bench_oriented_brg

mo271:


mo271:
now the error is "not yet implemented: Unsupported block context map", whereas before #190 we had `BlockContextMapSizeTooBig`.

szabadka (approved):

PR #189, opened 2025-05-15
add conformance test file: patches

mo271:


mo271:
fixed accidentally by #190

PR #190, opened 2025-05-16
improve reading of BlockContextMap

mo271:


jxl/src/frame/block_context_map.rs
```diff
@@ -88,14 +88,14 @@ impl BlockContextMap {
             let mut qf_thresholds: Vec<u32> = vec![0; num_qf_thresholds];
             for val in qf_thresholds.iter_mut() {
                 *val = match br.read(2)? {
-                    0 => 2,
+                    0 => br.read(2)?,
                     1 => br.read(3)? + 4,
                     2 => br.read(5)? + 12,
                     _ => br.read(8)? + 44,
-                } as u32;
+                } as u32 + 1;
```
mo271, line 95: https://github.com/libjxl/libjxl/blob/b026924a6eab285bf5b14585f10dfdb8249f7387/lib/jxl/entropy_coder.cc#L45

mo271:
should fix the error in #180

mo271:
we should really use U32Coder::Select here and in all places...

veluca93 (approved):

szabadka (approved):

PR #201, opened 2025-05-19
check OutOfBounds in peek

mo271:


jxl/src/bit_reader.rs
```diff
@@ -44,16 +44,20 @@ impl<'a> BitReader<'a> {
     }
 
     /// Reads `num` bits from the buffer without consuming them.
-    pub fn peek(&mut self, num: usize) -> u64 {
-        debug_assert!(num <= MAX_BITS_PER_CALL);
+    pub fn peek(&mut self, num: usize) -> Result<u64, Error> {
+        if num <= MAX_BITS_PER_CALL {
```
zond, line 48: Shouldn't this be >?

mo271: yes, but closing this anyhow..

PR #202, opened 2025-05-19
All Palette predictors should be fine with local_apply. Right?

eustas:


veluca93 (approved):
I believe you're right :-)

PR #203, opened 2025-05-19
Some more progress in VarDCT (incomplete `dequant_block`)

sboukortt:


veluca93 (approved):
LGTM, but please check with Zoltan :-)

jxl/src/frame/group.rs
```diff
@@ -7,17 +7,220 @@ use crate::{
...
+    util::{tracing_wrappers::*, CeilLog2},
+    var_dct::{
+        dct::compute_scaled_dct,
+        dct_scales::{dct_total_resample_scale, DctResampleScales, HasDctResampleScales},
+        transform::*,
+    },
     BLOCK_DIM, BLOCK_SIZE,
 };
 
+// Computes the lowest-frequency LF_ROWSxLF_COLS-sized square in output, which
```
szabadka, line 24: where is LF_ROWS / LF_COLS defined? isn't that just ROWS / COLS below?

jxl/src/frame/group.rs
```diff
@@ -7,17 +7,220 @@ use crate::{
...
+
+#[allow(dead_code)]
+fn lowest_frequencies_from_lf(
+    hf_type: HfTransformType,
+    lf: &[f32],
+    lf_stride: usize,
+    llf: &mut [f32],
+    scratch: &mut [f32],
+) {
+    let mut warm_block = [0.0; 4 * 4];
```
szabadka, line 71: can't we use scratch everywhere?

szabadka (approved):

jxl/src/frame/group.rs
sboukortt: Oh, yes, they were additional parameters in the C++ which we `static_assert`ed to be equal to `ROWS` / `COLS` so I figured we might as well just not take them. Thanks.

PR #205, opened 2025-05-20
Fixed a Numpy format TODO.

zond:
Made the header dict string be evenly divisible by 16 as per https://github.com/numpy/numpy/blob/main/doc/neps/nep-0001-npy-format.rst:

The next HEADER_LEN bytes form the header data describing the arrays format. It is an ASCII string which contains a Python literal expression of a dictionary. It is terminated by a newline ("n") and padded with spaces ("x20") to make the total length of the magic string + 4 + HEADER_LEN be evenly divisible by 16 for alignment purposes.


mo271 (approved):
Have you checked if numpy can still parse these (as a sanity check)?
If not, no worries, we will find out soon enough...

zond:
I have! :D

PR #213, opened 2025-05-20
Reset the bit position of the per-section bit readers to zero.

szabadka:
This does not affect the actual decoding, but this way we get comparable bit positions to libjxl in the logs.


veluca93:
I'm not sure this is a good idea, I find the libjxl offsets significantly more confusing -- could we change those instead?

szabadka:
> I'm not sure this is a good idea, I find the libjxl offsets significantly more confusing -- could we change those instead?

Ok, I will revert this. I realized the situation is even worse in libjxl since it resets the bit position before the file header and the frame header.

PR #217, opened 2025-05-21
Fix frame origin reading.

szabadka:


jxl/src/headers/frame_header.rs
```diff
@@ -347,11 +348,21 @@ pub struct FrameHeader {
...
+    y0_raw: u32,
+
+    #[coder(Bits(0))]
+    #[default(unpack_signed(x0_raw))]
+    #[condition(false)]
+    pub x0: i32,
+
+    #[coder(Bits(0))]
+    #[default(unpack_signed(y0_raw))]
+    #[condition(false)]
```
mo271, line 365: funny that this pattern was never used before elsewhere in the header parsing

mo271: actually, is this even necessary? Isn't https://github.com/libjxl/jxl-rs/blob/4c123a36847b2fa0f6ec731028b3a86e219bfcb4/jxl/src/headers/encodings.rs#L95 doing the correct thing already

veluca93 (approved):

PR #220, opened 2025-05-21
Fix handling of extra channel upsampling.

szabadka:


mo271 (approved):

jxl/src/headers/frame_header.rs
```diff
@@ -640,6 +640,15 @@ impl FrameHeader {
         Rect { origin, size }
     }
 
+    pub fn postprocess(&mut self, nonserialized: &FrameHeaderNonserialized) {
+        if self.upsampling > 1 {
+            for i in 0..nonserialized.extra_channel_info.len() {
+                let dim_shift = nonserialized.extra_channel_info[i].dim_shift();
+                self.ec_upsampling[i] <<= dim_shift;
```
mo271, line 647: surprised that clippy is not complaining here.

PR #222, opened 2025-05-21
Support chroma subsampling in entropy decoding.

szabadka:


jxl/src/headers/frame_header.rs
```diff
@@ -679,7 +679,7 @@ impl FrameHeader {
             return Err(Error::NonPatchReferenceWithCrop);
         }
         if !self.is444()
-            && ((self.flags & Flags::SKIP_ADAPTIVE_LF_SMOOTHING) != 0)
+            && ((self.flags & Flags::SKIP_ADAPTIVE_LF_SMOOTHING) == 0)
```
sboukortt, line 682: Oh, is this flag for the missing feature we were discussing?

sboukortt (approved):

jxl/src/headers/frame_header.rs
szabadka: Yes, DC smoothing should be done after the last DC group is decoded.

PR #223, opened 2025-05-22
create path when writing files

mo271:


zond:
Are we sure we want the CLI tool to create the parent directory if it doesn't exist? It's not the way CLI tools usually work, so it might be surprising.

mo271:
ah, perhaps not, I had mistakenly thought that djxl did the same..

zond:
> ah, perhaps not, I had mistakenly thought that djxl did the same..

If djxl does it then there might be good reason, but from the "mistakenly" I assume it doesn't? :D

PR #225, opened 2025-05-22
Allow images with zero width to make delta palette decoding work.

szabadka:
This is needed since delta palette can have implicit deltas even when num_deltas is 0.

### DIFF
--- a/jxl/src/frame/modular.rs
+++ b/jxl/src/frame/modular.rs
@@ -347,6 +347,25 @@ impl FullModularImage {
         }
 
         trace!(?section_buffer_indices);
+        #[cfg(feature = "tracing")]
+        for (section, indices) in section_buffer_indices.iter().enumerate() {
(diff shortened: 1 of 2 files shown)


jxl/src/frame/modular.rs
```diff
@@ -347,6 +347,24 @@ impl FullModularImage {
         }
 
         trace!(?section_buffer_indices);
+        for (section, indices) in section_buffer_indices.iter().enumerate() {
```
veluca93, line 350: you could put this under a `#[cfg(feature(tracing))]` and skip the underscores (and skip compiling the code if not under tracing)

veluca93 (approved):

jxl/src/frame/modular.rs
szabadka: Done.

veluca93 (approved):

PR #227, opened 2025-05-23
Implement reference properties.

szabadka:


jxl/src/image.rs
```diff
@@ -215,10 +215,6 @@ impl<T: ImageDataType> Image<T> {
         let total_size = xsize
             .checked_mul(ysize)
             .ok_or(Error::ImageSizeTooLarge(xsize, ysize))?;
-        // We need images with xsize == 0 for delta palette decoding
```
veluca93, line 218: Can we explain why this is fine?

veluca93 (approved):

jxl/src/image.rs
szabadka: Added a comment.

veluca93 (approved):

PR #231, opened 2025-05-26
Fixes for decoding lossless_pfm conformance test image.

szabadka:
Uses wrapping_add and wrapping_sub for i32 operations.

Disable error for reading more than 29 bits for a symbol.

### DIFF
--- a/jxl/src/entropy_coding/hybrid_uint.rs
+++ b/jxl/src/entropy_coding/hybrid_uint.rs
@@ -55,7 +55,8 @@ impl HybridUint {
         let bits_in_token = self.lsb_in_token + self.msb_in_token;
         let nbits =
             self.split_exponent - bits_in_token + ((symbol - self.split_token) >> bits_in_token);
-        if nbits > 29 {
+        // To match the behaviour of libjxl, we limit nbits to 31.
+        if nbits > 31 {
             return Err(Error::IntegerTooLarge(nbits));
         }
         let low = symbol & ((1 << self.lsb_in_token) - 1);
--- a/jxl/src/frame/modular/predict.rs
+++ b/jxl/src/frame/modular/predict.rs
@@ -169,8 +169,8 @@ const DIVLOOKUP: [u32; 64] = [
     289262, 284359, 279620, 275036, 270600, 266305, 262144,
 ];
 
-fn add_bits<T: num_traits::PrimInt>(x: T) -> i64 {
-    x.to_i64().unwrap() << PRED_EXTRA_BITS
+fn add_bits(x: i32) -> i64 {
+    (x as i64) << PRED_EXTRA_BITS
 }
 
 fn error_weight(x: u32, maxweight: u32) -> u32 {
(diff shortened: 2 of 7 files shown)


jxl/src/entropy_coding/hybrid_uint.rs
```diff
@@ -53,11 +53,11 @@ impl HybridUint {
...
         let bits_in_token = self.lsb_in_token + self.msb_in_token;
-        let nbits =
+        let mut nbits =
             self.split_exponent - bits_in_token + ((symbol - self.split_token) >> bits_in_token);
-        if nbits > 29 {
-            return Err(Error::IntegerTooLarge(nbits));
-        }
+        // To match the behaviour of libjxl, instead of returning an error here, we limit the
+        // size of nbits.
+        nbits &= 31;
```
veluca93, line 60: Can you use hex?

jxl/src/frame/modular/predict.rs
```diff
@@ -268,12 +268,12 @@ impl<'a> WeightedPredictorState<'a> {
         let te_w = if pos.0 == 0 {
             0
         } else {
-            self.error[cur_row + pos.0 - 1]
+            self.error[cur_row + pos.0 - 1] as i64
```
veluca93, line 271: uhm, if we actually need 64-bit arithmetic here I suspect the image might not be conformant... @jonsneyers what's your take here?

szabadka: libjxl also uses 64-bit arithmetic here:

```
    pixel_type_w teW = x == 0 ? 0 : error[cur_row + x - 1];
    pixel_type_w teN = error[pos_N];
    pixel_type_w teNW = error[pos_NW];
    pixel_type_w sumWN = teN + teW;
    pixel_type_w teNE = error[pos_NE];
```

jxl/src/entropy_coding/hybrid_uint.rs
szabadka: Done.

jxl/src/frame/modular/predict.rs
szabadka: Also, without the as i64 it does not even compile, since all the rest of the quantities (n, w, ...) and the prediction[] array are in i64 (both here and in libjxl).

jxl/src/entropy_coding/hybrid_uint.rs
```diff
@@ -53,11 +53,11 @@ impl HybridUint {
             return Ok(symbol);
         }
         let bits_in_token = self.lsb_in_token + self.msb_in_token;
-        let nbits =
+        let mut nbits =
             self.split_exponent - bits_in_token + ((symbol - self.split_token) >> bits_in_token);
-        if nbits > 29 {
-            return Err(Error::IntegerTooLarge(nbits));
-        }
+        // To match the behaviour of libjxl, instead of returning an error here, we limit the
```
veluca93, line 58: What if we limit the size to 31 instead?

szabadka: That works for lossless_pfm image. Done.

PR #239, opened 2025-05-28
Call FromLinear stage when output is PPM.

szabadka:
The numpy output remains in linear RGB.

Drive-by: move the deocding loop into its own module and use that for the tests as well.


jxl/src/decode.rs
```diff
@@ -0,0 +1,137 @@
...
+    icc::read_icc,
+};
+
+use std::marker::PhantomData;
+
+#[allow(clippy::type_complexity)]
+pub struct DecodeOptions<'a> {
+    pub xyb_output_linear: bool,
+    #[cfg(test)]
+    pub frame_callback: Option<&'a mut dyn FnMut(&Frame) -> Result<(), Error>>,
```
veluca93, line 21: Why is this not generic on an owned callback type?
i.e.
```
pub struct DecodeOptions<F: FnMut(&Frame) -> Result<(), Error>> {
...
}
```

For that matter, why is it `#[cfg(test)]`?

It should be relatively easy to have something like
```
impl DecodeOptions<fn(&Frame) -> Result<(), Error>> {
  pub fn new() {
     xyb_output_linear: true,
     frame_callback: None,
  }
}
```
and then have `new_for_testing` on generic `F` which takes the callback.

jxl/src/decode.rs
```diff
@@ -0,0 +1,137 @@
...
+        ),
+        frames: vec![],
+    };
+    let mut decoder_state = DecoderState::new(file_header);
+    decoder_state.xyb_output_linear = options.xyb_output_linear;
+    loop {
+        let mut frame = Frame::new(&mut br, decoder_state)?;
+        let mut section_readers = frame.sections(&mut br)?;
+
+        println!("read frame with {} sections", section_readers.len());
```
veluca93, line 77: Do we want to keep this as a print?

veluca93 (approved):

jxl/src/frame.rs
```diff
@@ -604,181 +609,133 @@ impl Frame {
...
     use test_log::test;
 
-    use super::{DecoderState, Frame, Section};
+    use super::Frame;
 
+    #[allow(clippy::type_complexity)]
     fn read_frames(
         image: &[u8],
-        mut callback: impl FnMut(Frame) -> Result<Option<DecoderState>, Error>,
+        callback: Option<&mut dyn FnMut(&Frame) -> Result<(), Error>>,
```
veluca93, line 628: probably should not have the Option here :-)

veluca93 (approved):

jxl/src/frame.rs
szabadka: Done.

jxl/src/decode.rs
szabadka: Removed cfg(test), but kept the lifetime, because that is needed by the tests.

PR #249, opened 2025-06-04
Call the chroma upsampling stage

sboukortt:


jxl/src/frame.rs
```diff
@@ -497,7 +497,14 @@ impl Frame {
             pipeline = pipeline.add_stage(ConvertModularToF32Stage::new(i, metadata.bit_depth))?;
         }
 
-        // TODO: chroma upsampling
+        for c in 0..3 {
+            for _ in 0..self.header.hshift(c) {
```
veluca93: I think this is either 0 or 1, so can we make this an `if`?

PR #251, opened 2025-06-05
Added jxl-rs conformance tests.

zond:


.github/workflows/pull_request.yml
```diff
@@ -105,3 +105,76 @@ jobs:
...
+      uses: actions/checkout@master
+      with:
+        repository: libjxl/bench
+        path: bench      
+        fetch-depth: 1
+
+    - name: Checkout bench dependencies
+      run: |
+        ( cd bench && git submodule update --recursive --init --depth 1 --recommend-shallow third_party/conformance )
+        ( cd bench && git submodule update --recursive --init --depth 1 --recommend-shallow third_party/jxl-rs )
```
mo271: isn't the point of having the conformance tests in a pull request to use the branch of jxl-rs from that pull request? 
No it seems like you are just pulling the head of jxl-rs, no?

Perhaps I'm missing something?

.github/workflows/pull_request.yml
```diff
@@ -105,3 +105,76 @@ jobs:
...
+      uses: dtolnay/rust-toolchain@stable
+      with:
+        components: clippy, rustfmt
+
+    - name: Rust cache
+      uses: Swatinem/rust-cache@v2.7.3
+
+    - name: Build jxl_cli
+      run: |
+        ( cd bench/third_party/jxl-rs && cargo build --release --bin jxl_cli )
```
mo271: I think we perhaps can just build the jxl_cli from ./ here, instead and not even clone `bench/third_party/jxl-rs`?

ci/conformance_summary.py
```diff
@@ -0,0 +1,30 @@
...
+    icc_json_path = sys.argv[1]
+    png_json_path = sys.argv[2]
+    print('\n')
+    print('| Name | ICC result | PNG result |')
+    print('|------|------------|------------|')
+    results = {}
+    with open(icc_json_path) as icc_json_input:
+        for test in json.load(icc_json_input):
+            results[test['test_id']] = {
+                'icc_success': '🟢 SUCCESS' if test['success'] else '🔴 FAILURE'
```
mo271:
```suggestion
                'icc_success': '🟢 Success' if test['success'] else '🔴 Failure'
```
why the shouting?

ci/conformance_summary.py
```diff
@@ -0,0 +1,30 @@
...
+            }
+    with open(png_json_path) as png_json_input:
+        for test in json.load(png_json_input):
+            results[test['test_id']]['png_success'] = '🟢 SUCCESS' if test['success'] else '🔴 FAILURE'
```
mo271:
```suggestion
            results[test['test_id']]['png_success'] = '🟢 Success' if test['success'] else '🔴 Failure'
```

ci/conformance_summary.py
mo271, line 1: It would be even cooler, if we could distinguish between
"No frame data available", i.e. no image was written and 
other errors. I would potentially even useful to propagate the error, since it is not always "RMSE too large"

mo271: perhaps "yellow" = at least the image decoded and "red"=no image found

.github/workflows/pull_request.yml
zond: Ah no - I just followed along in how bench worked stand alone and didn't think enough. Fixed now.

.github/workflows/pull_request.yml
zond: Yes

ci/conformance_summary.py
zond: Done

ci/conformance_summary.py
zond: Done

ci/conformance_summary.py
zond: Done

mo271:
also good to keep this step not "Required", because we don't want to get slowed down by it.

PR #255, opened 2025-06-10
Add patches stage.

szabadka:


jxl/src/render/stages/save.rs
```diff
@@ -9,7 +9,13 @@ use crate::{
     render::{RenderPipelineInspectStage, RenderPipelineStage},
 };
 
+pub enum SaveStageType {
+    OUTPUT,
```
veluca93: I believe this goes against the usual naming convention for enums, `Output` and `Reference`

veluca93 (approved):

jxl/src/render/stages/save.rs
szabadka: Done.

PR #256, opened 2025-06-11
test jxl_inspect for all test images

mo271:


mo271:
TODO: fix failing tests before merging

PR #257, opened 2025-06-11
WIP: upsampling

sboukortt:


jxl/src/headers/frame_header.rs
```diff
@@ -591,6 +591,14 @@ impl FrameHeader {
         )
     }
 
+    pub fn size_upsampled_padded(&self) -> (usize, usize) {
```
veluca93: (maybe padded_upsampled is better?)

jxl/src/render/simple_pipeline.rs
```diff
@@ -51,13 +52,13 @@ impl SimpleRenderPipelineBuilder {
                     num_channels
                 ]],
                 input_size: size,
-                log_group_size,
-                xgroups: size.0.shrc(log_group_size),
+                log_group_size: log_group_size + downsampling_shift,
```
veluca93: Could also do `log_group_size += downsampling_shift`, it might be more readable (well, maybe with a comment...)

veluca93 (approved):

jxl/src/render/simple_pipeline.rs
sboukortt: Something like this?

sboukortt: (I thought rustfmt would line-wrap my comment, not sure why it didn’t.)

PR #258, opened 2025-06-13
Use unpadded frame size in render pipeline.

szabadka:
Handle the cropping inside the run_stage_on() function instead. This is necesarry because the extend-to-image-size stage needs the unpadded size to work correctly.

Drive-by: Include alpha channel in the upsampling stage.

### DIFF
--- a/jxl/src/frame.rs
+++ b/jxl/src/frame.rs
@@ -511,7 +511,7 @@ impl Frame {
         let metadata = &decoder_state.file_header.image_metadata;
         let mut pipeline = SimpleRenderPipelineBuilder::new(
             num_channels + num_temp_channels,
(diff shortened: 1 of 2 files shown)


jxl/src/frame.rs
```diff
@@ -576,7 +576,7 @@ impl Frame {
 
         if frame_header.upsampling > 1 {
             let transform_data = &decoder_state.file_header.transform_data;
-            for c in 0..3 {
+            for c in 0..num_channels {
```
veluca93: Alpha should use extra channel upsampling, no?

szabadka: Yes. Fixed it, and replicated the early/late ec_upsampling logic as well.

veluca93 (approved):

PR #259, opened 2025-06-16
Fix shift direction in render pipeline.

szabadka:
Fixes the cafe conformance test image.

### DIFF
--- a/jxl/src/render/simple_pipeline.rs
+++ b/jxl/src/render/simple_pipeline.rs
@@ -263,8 +263,8 @@ impl SimpleRenderPipeline {
                 current_size = stage.new_size(current_size);
                 for (c, info) in self.channel_info[i + 1].iter().enumerate() {
                     if stage.uses_channel(c) {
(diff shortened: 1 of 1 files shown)


veluca93:
Didn't @sboukortt change this recently? Or was that just similar-looking code?

Should it be shift or ceiling-shift?

sboukortt:
Oh, right, this might have been caught in the cross-fire of “getting upsampling to work”, and I do think it was a `.shrc` before.

szabadka:
> Didn't @sboukortt change this recently? Or was that just similar-looking code?
> 
> Should it be shift or ceiling-shift?

It was changed from shrc to left shift. I am assuming it was a typo and it should have been changed from shrc to right shift.

sboukortt:
More likely, it shouldn’t have been changed at all.

szabadka:
Ok, added back the .shrc()

szabadka:
Conformance test has 5 red instead of 6 before.

PR #260, opened 2025-06-17
Fix missing RCT transform in modular mode.

szabadka:


szabadka:
Makes two conformance test cases green, and reduces RMS of patches_lossless from 0.19 to 0.004.

PR #262, opened 2025-06-18
Reduce allocas

c410-f3r:
`DeriveInput` and `Expr` have more than 128 bytes but aren't fully used in the associates functions, as such, they are now passed as references rather than values. These modifications don't impact performance as they are restricted in the macro crate but they are nice to have regardless.

### DIFF
--- a/jxl/src/entropy_coding/context_map.rs
+++ b/jxl/src/entropy_coding/context_map.rs
@@ -18,8 +18,7 @@ fn move_to_front(v: &mut [u8], index: u8) {
 }
 
 fn inverse_move_to_front(v: &mut [u8]) {
-    use array_init::array_init;
-    let mut mtf: [u8; 256] = array_init(|x| x as u8);
+    let mut mtf: [u8; 256] = std::array::from_fn(|x| x as u8);
     for val in v.iter_mut() {
         let index = *val;
         *val = mtf[index as usize];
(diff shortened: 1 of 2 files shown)


jxl/src/entropy_coding/context_map.rs
```diff
@@ -18,8 +18,7 @@ fn move_to_front(v: &mut [u8], index: u8) {
 }
 
 fn inverse_move_to_front(v: &mut [u8]) {
-    use array_init::array_init;
-    let mut mtf: [u8; 256] = array_init(|x| x as u8);
+    let mut mtf: [u8; 256] = std::array::from_fn(|x| x as u8);
```
c410-f3r, line 21: `array::from_fn` exists since 1.57 and is already well optimized.

veluca93, line 21: Can you also remove array_init from Cargo.toml?

c410-f3r: Not entirely because of `array_init::try_array_init`.

https://github.com/libjxl/jxl-rs/blob/36ba4e8ceef778e0b514f7439447cfac7312cd2c/jxl/src/headers/encodings.rs#L196

There is a standard library alternative but it is nightly-only -> https://github.com/rust-lang/rust/issues/89379

veluca93: Ah, sad.

PR #265, opened 2025-06-20
Orientation

mo271:


jxl/src/frame.rs
```diff
@@ -682,6 +682,7 @@ impl Frame {
                     i,
                     image_size,
                     1.0,
+                    metadata.orientation,
```
veluca93: this should be Orientation::Identity

mo271: done

veluca93 (approved):

PR #267, opened 2025-06-20
Avoid usage of an opaque pointer

c410-f3r:
Opaque pointers usually hide from the compiler the underlying structure, which generally prevents code optimizations.

### DIFF
--- a/jxl_cli/src/main.rs
+++ b/jxl_cli/src/main.rs
@@ -124,7 +124,7 @@ fn main() -> Result<(), Error> {
     }
 
     let mut options = DecodeOptions::new();
-    options.xyb_output_linear = String::from(opt.output.to_string_lossy()).ends_with(".npy");
+    options = options.set_xyb_output_linear(opt.output.ends_with(".npy"));
     let (image_data, icc_bytes) = jxl::decode::decode_jxl_codestream(options, &codestream)?;
 
     let icc_result = save_icc(icc_bytes, opt.icc_out);
--- a/jxl/src/decode.rs
+++ b/jxl/src/decode.rs
@@ -14,15 +14,15 @@ use crate::{
 };
 
 #[allow(clippy::type_complexity)]
-pub struct DecodeOptions<'a> {
-    pub xyb_output_linear: bool,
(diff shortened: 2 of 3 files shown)


jxl_cli/src/main.rs
```diff
@@ -124,7 +124,7 @@ fn main() -> Result<(), Error> {
     }
 
     let mut options = DecodeOptions::new();
-    options.xyb_output_linear = String::from(opt.output.to_string_lossy()).ends_with(".npy");
```
c410-f3r, line 127: It is not necessary to perform a heap allocation to just check if the string ends with `.npy`.

veluca93: Can you make a separate PR with this change?

veluca93 (commented):
I'm not sure monomorphizing on the callback type is the correct tradeoff here, as I would expect it to possibly cause a meaningful increase in code size.

I would prefer to evaluate the code size vs performance tradeoff of these choices at a later point.

c410-f3r:
With these minor modifications the size of the `jxl_cli` binary had a cut of ~500 bytes.

Monomorphization allows the generalization of function pointers, function identifiers, closures and is actually nice in this scenario. For example, closures can be optimized at the calling site potentially erasing what was constructed at the definition site.

https://github.com/libjxl/jxl-rs/blob/46f8d4d7011320e75d2b7540025160262f8c5e9f/jxl/src/frame.rs#L952-L967

veluca93:
> With these minor modifications the size of the `jxl_cli` binary had a cut of ~500 bytes.
> 
> Monomorphizing allows the generalization of function pointers, function identifiers, closures and is actually nice in this scenario. For example, closures can be optimized at the calling site potentially erasing what was constructed at the definition site.
> 
> https://github.com/libjxl/jxl-rs/blob/46f8d4d7011320e75d2b7540025160262f8c5e9f/jxl/src/frame.rs#L952-L967

That's nice, but the decoder is far from ready and I can absolutely imagine things changing as we get close to that, hence I'd prefer to make those decisions later :-)

PR #268, opened 2025-06-20
Write 16-bit PNGs when warranted

sboukortt:


sboukortt:
(It worked on the first try, which is somewhat suspicious.)

veluca93 (approved):

PR #269, opened 2025-06-20
Implement delta palette decoding.

szabadka:


jxl/src/frame/modular/predict.rs
```diff
@@ -88,6 +100,96 @@ impl PredictionData {
             toprightright,
         }
     }
+
+    #[allow(clippy::too_many_arguments)]
+    #[allow(clippy::unnecessary_unwrap)]
```
veluca93: Can we not do this and instead do `if let Some(l) = rect_left { l.row(y)[xsize-1] }`?

jxl/src/frame/modular/transforms.rs
```diff
@@ -272,18 +272,7 @@ pub fn make_grids(
...
-                        | Predictor::Gradient
-                        | Predictor::NorthWest
-                        | Predictor::AverageWestAndNorth
-                        | Predictor::AverageWestAndNorthWest
-                        | Predictor::AverageNorthAndNorthWest => &[(0, -1), (-1, 0), (-1, -1)],
-                        Predictor::NorthEast | Predictor::AverageNorthAndNorthEast => {
-                            &[(0, -1), (1, 0), (1, -1)]
-                        }
-                        Predictor::AverageAll | Predictor::Weighted => unreachable!(),
+                        _ => &[(0, -1), (-1, 0), (-1, -1)],
```
veluca93, line 275: Why do we need all of those even for West/WestWest/North?

jxl/src/frame/modular/transforms/apply.rs
```diff
@@ -132,24 +148,112 @@ impl TransformStepChunk {
...
-                            &img_pal,
-                            &mut bufs,
-                            *num_colors,
-                            *num_deltas,
-                            *predictor,
-                            wp_header,
-                        );
-                        Ok(())
-                    })?;
+                    with_buffers(buffers, buf_out, out_grid, |_| Ok(()))?;
```
veluca93, line 153: Is this just meant to ensure that the buffers are present?

I am guessing you're not doing the thing in the callback, and I imagine that's because the interface is not powerful enough, but it might be nicer to extend that code instead

veluca93 (approved):

jxl/src/frame/modular/predict.rs
szabadka: I tried that but it complained that I can't use ```let``` in an if statement with && operators.

jxl/src/frame/modular/transforms.rs
szabadka: Even West can use Top if x==0, maybe we could spare (-1, -1) sometimes but that would make the delta palette code a bit more complex (now we always use a rectangle of groups).

jxl/src/frame/modular/transforms/apply.rs
szabadka: Yes, it is just for creating buffers. Added a comment and a TODO for extending the callback functionality.

jxl/src/frame/modular/predict.rs
veluca93: Ah that's annoying. I would probably still prefer nested `if`s over the unwrap, though.

szabadka: Ok, done, and I also found a bug in leftleft/toptop, so it was worth a second look. But now we have some scary looking nested ifs.

veluca93: I do expect to eventually rewrite most of this logic to instead copy rows, fwiw :-)

PR #274, opened 2025-06-23
Initial (unimplemented) version of an API.

veluca93:


szabadka (approved):

jxl/src/api.rs
```diff
@@ -0,0 +1,351 @@
...
+    ) -> Result<ProcessingResult<Self, Self>> {
+        todo!()
+    }
+
+    /// Obtains the image's basic information.
+    pub fn basic_info(&self) -> &JxlBasicInfo {
+        todo!()
+    }
+
+    /// Retrives the file's color profile.
```
mo271:
```suggestion
    /// Retrieves the file's color profile.
```

jxl/src/api.rs
```diff
@@ -0,0 +1,351 @@
...
+    pub fn embedded_color_profile(&self) -> &JxlColorProfile {
+        todo!()
+    }
+
+    /// Specifies the preferred color profile to be used for outputing data.
```
mo271:
```suggestion
    /// Specifies the preferred color profile to be used for outputting data.
```

jxl/src/api.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    pub fn set_pixel_format(&mut self, pixel_format: JxlPixelFormat) {
+        todo!()
+    }
+
+    pub fn process(
+        self,
+        input: &mut impl JxlBitstreamInput,
+    ) -> Result<ProcessingResult<JxlDecoder<WithImageInfo>, Self>> {
```
mo271:
```suggestion
    ) -> Result<ProcessingResult<JxlDecoder<WithFrameInfo>, Self>> {
```

jxl/src/api.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    /// Resets entirely a decoder, producing a new decoder with the same settings.
+    /// This is faster than creating a new decoder in some cases.
+    pub fn reset(self) -> JxlDecoder<Initialized> {
+        JxlDecoder::wrap_inner(self.inner)
+    }
+
+    /// Rewinds a decoder to the start of the file, allowing past frames to be displayed again.
+    pub fn rewind(self) -> JxlDecoder<Initialized> {
+        JxlDecoder::wrap_inner(self.inner)
```
mo271, line 177: Is `rewind` also producing a new decoder? If so perhaps perhaps we only need one of those?

veluca93: Rewind will optimize decoding assuming that _the same image_ will be decoded.

mo271: ok, perhaps worth mentioning in the doc string, but not super important

veluca93: Will do (and improve the documentation in general) as I implement it

PR #278, opened 2025-06-23
Add splines stage to render pipeline.

szabadka:


jxl/src/frame.rs
```diff
@@ -596,7 +596,13 @@ impl Frame {
             })?
         }
 
-        // TODO: splines
+        if frame_header.has_splines() {
+            pipeline = pipeline.add_stage(SplinesStage::new(
+                <Option<Splines> as Clone>::clone(&lf_global.splines).unwrap(),
```
veluca93: why not just `lf_global.splines.clone()`?

veluca93 (approved):

jxl/src/frame.rs
szabadka: .clone().unwrap() works

PR #283, opened 2025-06-23
Implement JxlOutputBuffer.

veluca93:


jxl/src/api/output.rs
```diff
@@ -0,0 +1,85 @@
...
+impl<'a> JxlOutputBuffer<'a> for [MaybeUninit<u8>] {
+    unsafe fn get_row_buffers(
+        &'a mut self,
+        shape: (usize, usize),
+        bytes_per_pixel: usize,
+    ) -> impl DerefMut<Target = [&'a mut [MaybeUninit<u8>]]> {
+        assert!(self.len() >= shape.0 * shape.1 * bytes_per_pixel);
+        let mut ret = vec![];
+        let mut slc = self;
+        for i in 0..shape.1 {
```
mo271: why isn't clippy complaining about the unused loop var `i` here?

jxl/src/api/output.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+impl<'a> JxlOutputBuffer<'a> for [&'a mut [MaybeUninit<u8>]] {
+    unsafe fn get_row_buffers(
+        &'a mut self,
+        shape: (usize, usize),
+        bytes_per_pixel: usize,
+    ) -> impl DerefMut<Target = [&'a mut [MaybeUninit<u8>]]> {
+        self
```
mo271, line 97:
```suggestion
        assert_eq!(self.len(), shape.1, "Incorrect number of rows provided");
        for row in self.iter() {
            assert!(row.len() >= shape.0 * bytes_per_pixel, "A row buffer is too short");
        }
        self
```

Perhaps add these kind of asserts here? You have a similar check for the flattened version.

jxl/src/api/output.rs
veluca93: That would be because of the `#[allow(unused_variables)]` in the parent module...

Fixed.

PR #284, opened 2025-06-24
implement api: check_signature

mo271:


mo271:
- ~~Not sure if we want tests for that~~ <-- tests added
 - Eventually we just want to call the api function from parse.rs, ~~at which point we might want to move the definitions of CO.*_SIGNATURE? Or should we move them now?~~ <-- moved now

jxl/src/api.rs
```diff
@@ -44,8 +45,49 @@ pub trait JxlBitstreamInput {
...
+/// # Returns
+///
+/// A `ProcessingResult` which is:
+/// - `Complete(Some(_))` if a full container or codestream signature is found.
+/// - `Complete(None)` if the prefix is definitively not a JXL signature.
+/// - `NeedsMoreInput` if the prefix matches a signature but is too short.
 pub fn check_signature(file_prefix: &[u8]) -> ProcessingResult<Option<JxlSignatureType>, ()> {
-    todo!()
+    // Check for the longer, more specific container signature first.
+    if CONTAINER_SIGNATURE.starts_with(file_prefix) {
```
veluca93: This doesn't look correct, did you try this out?

mo271: Added tests and fixed the code so that it does work..

mo271:
will add a test macro, one sec...

jxl/src/api.rs
```diff
@@ -54,36 +55,46 @@ pub trait JxlBitstreamInput {
...
                 result: Some(JxlSignatureType::Container),
-            };
+            }
         } else {
-            // It's a valid prefix of the container signature, but we need more data.
-            return ProcessingResult::NeedsMoreInput {
-                size_hint: CONTAINER_SIGNATURE.len() - file_prefix.len(),
+            // No, we need more bytes to complete the signature.
+            ProcessingResult::NeedsMoreInput {
+                size_hint: container_len - prefix_len,
```
veluca93: I believe this is incorrect if `file_prefix` is [] :-) (or in general if we are in the common prefix of the two signatures, but that's empty)

veluca93 (commented):
What do you think of moving all the signature stuff in a separate file and doing `pub use`?

jxl/src/api.rs
mo271: With the previous logic, we had the check:
```
    signature_test!(empty_prefix, &[], NeedsMoreInput(CONTAINER_SIGNATURE.len()));
```
So, for the emtpy `file_prefix`, we return `NeedsMoreInput(CONTAINER_SIGNATURE.len())`.
So the `NeedsMoreInput` seems correct, do you suggest to return only the length of the shorter `CODESTREAM_SIGNATURE`? I suppose we can give that size hint instead. Now changed to 
```
    signature_test!(empty_prefix, &[], NeedsMoreInput(CODESTREAM_SIGNATURE.len()));
```
for the empty prefix. The situation is a bit ambiguous. Is this what you had in mind?

veluca93: Yup

jxl/src/api/signature.rs
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+use crate::api::ProcessingResult;
+
+/// The magic bytes for a bare JPEG XL codestream.
+pub const CODESTREAM_SIGNATURE: [u8; 2] = [0xff, 0x0a];
```
veluca93: `pub`, or `pub(crate)`?

mo271: `pub(crate)`

PR #289, opened 2025-06-25
Add EPF stage.

szabadka:


jxl/src/image.rs
```diff
@@ -255,6 +255,16 @@ impl<T: ImageDataType> Image<T> {
         self.size
     }
 
+    pub fn fill(&mut self, v: T) {
```
mo271, line 258: Can be made 
`#[cfg(test)]` if I'm not mistaken.
```suggestion
    #[cfg(test)] 
    pub fn fill(&mut self, v: T) {
```

mo271: ah, sorry no

jxl/src/features/epf.rs
```diff
@@ -0,0 +1,72 @@
...
+                sigma_rect.row(sby)[SIGMA_PADDING + size_blocks.0 - 1];
+        }
+        for bx in 0..sigma_xsize {
+            sigma_rect.row(SIGMA_PADDING - 1)[bx] = sigma_rect.row(SIGMA_PADDING)[bx];
+            sigma_rect.row(SIGMA_PADDING + size_blocks.1)[bx] =
+                sigma_rect.row(SIGMA_PADDING + size_blocks.1 - 1)[bx];
+        }
+    } else {
+        let sigma = INV_SIGMA_NUM / rf.epf_sigma_for_modular;
+        sigma_image.fill(sigma);
```
veluca93, line 71: Might be better to not allocate an image but return an enum here (can also do it later)

szabadka: Added a TODO for now, we can fix this during performance / memory usage optimization.

PR #313, opened 2025-07-04
Move most operations to the public JxlColorEncoding

sboukortt:
They are safer and sometimes more convenient to write since chromaticity data and gamma values are colocated with the corresponding “custom” enum entries and accessible only if it makes sense to do so.


jxl/src/api/color.rs
```diff
@@ -3,19 +3,295 @@
...
+    let inv_det = 1.0 / det;
+
+    let adjugate_matrix: [[f64; 3]; 3] =
+        std::array::from_fn(|r_idx| std::array::from_fn(|c_idx| cofactor_matrix[c_idx][r_idx]));
+
+    // Inverse matrix = (1/det) * Adjugate matrix.
+    Ok(std::array::from_fn(|r_idx| {
+        std::array::from_fn(|c_idx| adjugate_matrix[r_idx][c_idx] * inv_det)
+    }))
+}
```
veluca93: All the linalg stuff is probably best in a separate module (probably not under `api`)

PR #315, opened 2025-07-06
Prepare for multithreaded pipeline runner

tirr-c:
- Receive `&self` at `process_row_chunk`.
- Add ability to create and receive thread-local state to `RenderPipelineStage`. Though I'm not sure if this is the best interface...
  - Rayon has [`.try_for_each_init()`][rayon] that allows creating thread-local state.

[rayon]: https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.try_for_each_init


veluca93 (approved):
I am fairly sure that `SaveStage` will need to be different in the end, but given that I am already aware of needing to change other things in it this is fine for now.

PR #316, opened 2025-07-08
Initial implementation of the main process() function.

veluca93:
Summary of the implementation choices:
- There are two different parsers, one for boxes and one for the codestream.
- The codestream parser pulls data from the box parser
- The codestream parser tries to minimize memory usage by putting data into buffers of the size of single sections.
- Data might be not read if it turns out not to be needed.

Still TODO:
- Integrating this code with jxl_cli and testing it
- Support output in different pixel formats
- Support output with different color profiles
- Implement flush_pixels()
- Generate BasicInfo and Frame information
- Integrate writing output in the render pipeline.


jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
-        bytes_per_pixel: usize,
-    ) -> impl DerefMut<Target = [&'a mut [MaybeUninit<u8>]]>;
+use num_traits::ToBytes;
+
+use crate::image::Image;
+
+pub struct JxlOutputBuffer<'a> {
+    // Safety invariants:
+    //  - uninit data is never written to `buf`.
+    //  - `buf` is valid for writes for all bytes in the range
```
szabadka, line 18: What is 'valid' in this context? Isn't all this just saying that 'buf' must have at least size E (as defined below)?

jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
+    //    `buf[i*bytes_between_rows..i*bytes_between_rows+bytes_per_row]` for all values of `i`
+    //    from `0` to `num_rows-1`.
+    //  - `self` has exclusive (write) access to the bytes in these ranges.
+    //  - all the bytes in those ranges (and in between) are part of the same allocated object.
+    //  - `num_rows > 0`, `bytes_per_row > 0`, `bytes_per_row <= bytes_between_rows`.
+    //  - The computation `E = bytes_between_rows * (num_rows-1) + bytes_per_row` does not
+    //    overflow and has a result that is at most `isize::MAX`.
+    buf: *mut MaybeUninit<u8>,
```
szabadka, line 26: wouldn't using ```&mut [V]```, where ```V: AsMut<[u8]>``` as in the render-pipeline be better? then it is up to the caller to initialize it or not

jxl/src/api/inner/codestream_parser.rs
```diff
@@ -0,0 +1,215 @@
...
+    ready_section_data: usize,
+    skip_sections: bool,
+
+    section_state: SectionState,
+    available_sections: Vec<SectionBuffer>,
+
+    pub(super) has_more_frames: bool,
+}
+
+#[allow(unsafe_code)]
```
szabadka: why do we need this?

jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
+    /// Creates a new JxlOutputBuffer from a slice of uninit data.
+    /// It is guaranteed that `buf` will never be used to write uninitalized data.
+    pub fn new_uninit(
+        buf: &'a mut [MaybeUninit<u8>],
+        num_rows: usize,
+        bytes_per_row: usize,
+    ) -> Self {
+        Self::check_vals(num_rows, bytes_per_row, bytes_per_row);
+        assert!(buf.len() >= bytes_per_row * num_rows);
+        // Safety: buf-related invariants follow from this function mutably borrowing a slice, and
```
szabadka: shouldn't here be an unsafe block?

jxl/src/api/output.rs
veluca93: It's more complicated than that, see https://doc.rust-lang.org/std/ptr/index.html#safety

jxl/src/api/output.rs
veluca93: `[MaybeUninit<u8>]` does not implement `AsMut<[u8]>` (or the other way round)

jxl/src/api/inner/codestream_parser.rs
veluca93: We don't, it's a leftover from earlier.

jxl/src/api/output.rs
veluca93: Unfortunately [unsafe fields](https://github.com/rust-lang/rfcs/pull/3458) are not stable yet, so upholding satisfying safety invariants doesn't need unsafe blocks :(

jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
+    /// - The bytes in these ranges must not be accessed as long as the returned `Self` is in scope.
+    /// - All the bytes in those ranges (and in between) must be part of the same allocated object.
+    pub unsafe fn new_from_ptr(
+        buf: *mut MaybeUninit<u8>,
+        num_rows: usize,
+        bytes_per_row: usize,
+        bytes_between_rows: usize,
+    ) -> Self {
+        Self::check_vals(num_rows, bytes_per_row, bytes_between_rows);
+        // Safety: caller guarantees the buf-related invariants, and other safety invariants are
```
szabadka: the first invariant is not guaranteed by the caller, but by the implementation of the struct, right? and it can only be guaranteed by keeping buf private (maybe it would be worth adding a comment that it must not be public) and adding # safety contracts to functions that expose it

jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
+            // Safety: `new_uninit` guarantees that no uninit data is ever written to the passed-in
+            // slice. Moreover, `T` and `MaybeUninit<T>` have the same memory layout.
+            unsafe { std::mem::transmute::<&'a mut [u8], &'a mut [MaybeUninit<u8>]>(buf) },
+            num_rows,
+            bytes_per_row,
+        )
+    }
+
+    /// Creates a new JxlOutputBuffer from a slice of uninit data.
+    /// It is guaranteed that `buf` will never be used to write uninitalized data.
```
szabadka, line 86: how can this function in itself guarantee this? only the entire implementation of this struct can, so this comment should go to some struct-level place (see also in the // Safety comment of new() above)

jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
-    ) -> impl DerefMut<Target = [&'a mut [MaybeUninit<u8>]]> {
-        assert_eq!(self.len(), shape.1, "Incorrect number of rows provided");
-        for row in self.iter() {
-            assert!(
-                row.len() >= shape.0 * bytes_per_pixel,
-                "A row buffer is too short"
-            );
+    pub fn new(buf: &'a mut [u8], num_rows: usize, bytes_per_row: usize) -> Self {
+        Self::new_uninit(
+            // Safety: `new_uninit` guarantees that no uninit data is ever written to the passed-in
```
szabadka, line 77: new_uninit in itself can not guarantee this

jxl/src/api/output.rs
```diff
@@ -5,95 +5,169 @@
...
+    /// It is guaranteed that `buf` will never be used to write uninitalized data.
+    pub fn new_uninit(
+        buf: &'a mut [MaybeUninit<u8>],
+        num_rows: usize,
+        bytes_per_row: usize,
+    ) -> Self {
+        Self::check_vals(num_rows, bytes_per_row, bytes_per_row);
+        assert!(buf.len() >= bytes_per_row * num_rows);
+        // Safety note: buf-related invariants follow from this function mutably borrowing a slice,
+        // and other safety invariants are checked in check_vals.
```
szabadka: ... or guaranteed by the struct

jxl/src/api/output.rs
veluca93: It does, transitively (as in, it promises this to the caller, and relies on promises in the struct safety invariant to make those promises).

jxl/src/api/output.rs
veluca93: (see above)

jxl/src/api/output.rs
veluca93: The first invariant, at struct construction time, is a promise, not a requirement. I changed the wording slightly.

On adding a comment saying that it should not be public: you should not do that if you see a big `Safety invariants` comment in the struct definition, IMO that counts as enough of a comment.

jxl/src/api/output.rs
veluca93: I changed the wording slightly here too.

szabadka (approved):

PR #321, opened 2025-08-05
Fix api tests

zond:


zond:
I believe we still have bugs, some images don't decode correctly (e.g. cafe.jxl decodes seemingly correctly, but bike.jxl is just black) - but all images decode without panics and errors now.

PR #326, opened 2025-08-11
Multiple improvements

zond:
- Re-introduced the option to run jxl_cli with the library decoder directly instead of the API.
- Fixed some container format parsing bugs.


veluca93 (commented):
On 1: why?
On 2: how did you find the bugs?

zond:
> On 1: why?

To enable finding bugs like this a while longer - it's super helpful to be able to compare code paths with some debug tracing and a command line argument :D

> On 2: how did you find the bugs?

An image on https://jpegxl.info/resources/jpeg-xl-test-page.html (`2x "Retina" Image`) that didn't render correctly. @sboukortt is working on finding an image with the same issue (possibly the same one) that we are sure we have license to put in the test images directory.

jxl_cli/src/main.rs
```diff
@@ -127,17 +130,106 @@ fn images_from_rgb_vec(vec: &[u8], size: (usize, usize)) -> Result<Vec<Image<f32
...
 
-fn main() -> Result<(), Error> {
-    #[cfg(feature = "tracing-subscriber")]
-    {
-        use tracing_subscriber::{EnvFilter, fmt, prelude::*};
-        tracing_subscriber::registry()
-            .with(fmt::layer())
-            .with(EnvFilter::from_default_env())
-            .init();
+fn with_lib(opt: Opt) -> Result<()> {
```
veluca93, line 136: Maybe add a TODO that explains that this is temporary to aid debugging?

zond: Done

PR #328, opened 2025-08-13
Added a simple HTML file to view all test images in the same tab.

zond:


mo271 (approved):
perhaps one could sorf lines 9 through 71 alphabetically?
Other than that LGTM!

PR #329, opened 2025-08-13
Added candle.jxl to test images.

zond:


zond:
@sboukortt ping

sboukortt:
How about copying https://github.com/AcademySoftwareFoundation/openexr-images/blob/main/LICENSE as `candle.license.txt` or `candle.txt` or something like that? Possibly with a link to the original (https://github.com/AcademySoftwareFoundation/openexr-images/blob/main/ScanLines/CandleGlass.exr).

zond:
OK, added the license.

PR #336, opened 2025-08-14
Replaced ExtraChannelInfo in JxlBasicInfo with JxlExtraChannel.

zond:


jxl/src/api/data_types.rs
```diff
@@ -107,3 +109,8 @@ impl JxlBitDepth {
         }
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct JxlExtraChannel {
+    pub ec_type: ExtraChannel,
```
sboukortt, line 115: Don’t we also need `alpha_associated` for Firefox?

zond: Ah yes. Updated.

PR #339, opened 2025-08-14
Initial implementation of multi-arch SIMD support.

veluca93:


jxl/src/simd/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    + Sub<Self, Output = Self>
+    + AddAssign<Self>
+    + MulAssign<Self>
+    + SubAssign<Self>
+{
+    type Descriptor: SimdDescriptor;
+
+    const LEN: usize;
+
+    fn splat(d: Self::Descriptor, v: f32) -> Self;
```
szabadka, line 61: could you add a comment here? I looked up the dictionary definition of splat: "a single flat thin often ornamental member of a back of a chair" which does not seem to be related...

jxl/src/render/simple_pipeline.rs
```diff
@@ -19,6 +19,13 @@ use super::{
     internal::RenderPipelineStageType,
 };
 
+const CACHE_LINE_BYTE_SIZE: usize = 64;
+
+pub fn round_up_size_to_two_cache_lines<T>(size: usize) -> usize {
```
szabadka: this seems a bit out of place here, probably should go to the simd or some util file?

szabadka (approved):

jxl/src/render/simple_pipeline.rs
```diff
@@ -19,6 +19,13 @@ use super::{
...
+    let elements_per_cache_line = CACHE_LINE_BYTE_SIZE / std::mem::size_of::<T>() * 2;
```
sboukortt: This is `(CACHE_LINE_BYTE_SIZE / std::mem::size_of::<T>()) * 2` (division first), right?

jxl/src/render/simple_pipeline.rs
veluca93: you are probably right. done.

jxl/src/simd/mod.rs
veluca93: TIL. I took the name from https://doc.rust-lang.org/std/simd/type.f32x4.html#method.splat, but I had no idea it didn't have a particularly related meaning (it "felt" right when I read it)

jxl/src/render/simple_pipeline.rs
veluca93: Yep

PR #341, opened 2025-08-14
Added JxlFrameHeader

zond:
JxlFrameHeader is returned from JxlDecoder<WithFrameInfo>.frame_header().

### DIFF
--- a/jxl/src/api/inner.rs
+++ b/jxl/src/api/inner.rs
@@ -71,25 +71,19 @@ impl JxlDecoderInner {
     }
 
     pub fn frame_header(&self) -> Option<JxlFrameHeader> {
-        if let Some(file_header) = &self.codestream_parser.file_header {
-            if let Some(frame) = &self.codestream_parser.frame {
-                let header = frame.header();
-                Some(JxlFrameHeader {
-                    name: header.name.clone(),
-                    duration: file_header
-                        .image_metadata
-                        .animation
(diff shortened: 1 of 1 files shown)


jxl/src/api/inner.rs
```diff
@@ -71,9 +70,24 @@ impl JxlDecoderInner {
...
+                        .animation
+                        .as_ref()
+                        .map(|anim| header.duration(anim)),
+                })
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
sboukortt:
```suggestion
        let Some(file_header) = &self.codestream_parser.file_header else {
            return None;
        };
        let Some(frame) = &self.codestream_parser.frame else {
            return None;
        };
        let header = frame.header();
        Some(JxlFrameHeader {
            name: header.name.clone(),
            duration: file_header
                .image_metadata
                .animation
                .as_ref()
                .map(|anim| header.duration(anim)),
        })
```

sboukortt (approved):

jxl/src/api/inner.rs
zond: Good catch on that one - I think I found an even nicer way to write it. Take a look!

sboukortt: Nice indeed 😁

sboukortt (approved):

PR #343, opened 2025-08-15
Support float samples in ConvertModularToF32Stage

sboukortt:


jxl/src/render/stages/convert.rs
```diff
@@ -108,28 +108,73 @@ impl RenderPipelineStage for ConvertModularXYBToF32Stage {
...
             f,
-            "convert modular data to F32 in channel {} with scale {}",
-            self.channel, self.scale
+            "convert modular data to F32 in channel {} with bit depth {:?}",
+            self.channel, self.bit_depth
         )
     }
 }
 
+fn int_to_float(input: &[i32], output: &mut [f32], bit_depth: &BitDepth) {
```
zond, line 133: I would love a comment about what this function actually does on a higher level, it's not trivial code :D

sboukortt: I have copied libjxl’s comment; does that help?

zond: It helps some :D 

Thanks!

jxl/src/render/stages/convert.rs
```diff
@@ -108,28 +108,76 @@ impl RenderPipelineStage for ConvertModularXYBToF32Stage {
...
-            self.channel, self.scale
+            "convert modular data to F32 in channel {} with bit depth {:?}",
+            self.channel, self.bit_depth
         )
     }
 }
 
+// Convert custom [bits]-bit float (with [exp_bits] exponent bits) stored as int
+// back to binary32 float.
+fn int_to_float(input: &[i32], output: &mut [f32], bit_depth: &BitDepth) {
```
veluca93, line 133: Maybe leave a comment for future-us to SIMD this?

sboukortt: (For what it’s worth, for an image where this is called, we currently spend 0.4% of the time in this stage.)

PR #344, opened 2025-08-15
With --without-api, restore the old “do as much as possible” behavior

sboukortt:


zond:
Oh, sorry! Maybe I should have asked why it looked that way - I just refactored it as simple as possible when I replaced the error types.

Why don't we want to break at first error?

zond (approved):

sboukortt:
I don’t remember exactly but it must have been along the lines of “let’s try to be as useful as we can even if something goes wrong” 🙂

Issue #330, opened 2025-08-13
Images incorrectly rendered by jxl_cli


zond:
If more images are found, they can be added so we don't forget them.

If they are incorrectly added to this list (they are actually correct, or as correct as we can make them) then feel free to remove them or just check the box.

- [x] jxl/resources/test/grayscale_patches_var_dct.jxl
- [x] jxl/resources/test/candle.jxl
- [x] https://github.com/tirr-c/jxl-oxide-testdata/blob/0dafe6807b8f26c605627c59f39cbb063d4225a7/issue_32/input.jxl (added by @tirr-c in https://github.com/libjxl/jxl-rs/issues/330#issuecomment-3194223061)

zond:
@sboukortt could you verify that these are actually incorrectly rendered? IIRC candle.jxl was as good as it could get with PNG?

sboukortt:
It seems the decoded alpha for `candle` is too low somehow.

tirr-c:
Found another incorrectly decoded image: https://github.com/tirr-c/jxl-oxide-testdata/blob/0dafe6807b8f26c605627c59f39cbb063d4225a7/issue_32/input.jxl

I remember this triggers some tricky chroma subsampling edge case. (Fixed by #347)

tirr-c:
Also this: https://github.com/tirr-c/jxl-oxide-testdata/blob/0dafe6807b8f26c605627c59f39cbb063d4225a7/squeeze_edge/input.jxl

This image triggers a Squeeze transform edge case. (Will be fixed by #346)

zond:
@tirr-c thanks!

I added the first image you mentioned to the list, but the second one seems to be fixed already, right?

tirr-c:
Yep, it's fixed by #346.

zond:
candle.jxl works now, I'm not certain which of the recent PRs fixed it.

sboukortt:
#343 got the alpha not to be ~a quarter of what it should be, and then #352 refined it a bit further for PNG output.

zond:
I'll close this now, we can reopen or create  new one once we see failing images again.

PR #347, opened 2025-08-18
Fix using wrong rects during LF dequant of chroma subsampled image

tirr-c:
The bug is triggered when there are multiple LF groups. cc https://github.com/libjxl/jxl-rs/issues/330#issuecomment-3194223061

I'm not sure about how to add tests for this...


veluca93:
> The bug is triggered when there are multiple LF groups. cc [#330 (comment)](https://github.com/libjxl/jxl-rs/issues/330#issuecomment-3194223061)
> 
> I'm not sure about how to add tests for this...

I imagine a file that would trigger the issue would be fairly large?

tirr-c:
Added an image and test that checks LF image buffer.

veluca93 (approved):

PR #348, opened 2025-08-19
Remove some allow(dead_code) annotations.

veluca93:


mo271 (approved):

jxl/src/frame/color_correlation_map.rs
```diff
@@ -12,6 +12,8 @@ use std::default::Default;
 
 pub const COLOR_TILE_DIM: usize = 64;
 
+const _: () = assert!(COLOR_TILE_DIM % BLOCK_DIM == 0);
```
sboukortt, line 15: Would it make any sense at all to `#[cfg(test)]` this so that it’s skipped in the final library / binaries? (Or is that not how it works?)

jxl/src/lib.rs
```diff
@@ -24,8 +24,6 @@ pub mod var_dct;
 // TODO: Move these to a more appropriate location.
 const BLOCK_DIM: usize = 8;
```
sboukortt, line 26: Apparently, we have this one in `FrameHeader` above. Should we reduce to just one copy?

jxl/src/frame/color_correlation_map.rs
veluca93: I don't believe `const`s actually exist as far as the final binary is concerned :-)

sboukortt: Ah, right :D I must have read it with my C++ glasses.

jxl/src/lib.rs
veluca93: Got rid of the ones in FrameHeader.

PR #352, opened 2025-08-24
Round sample values when writing to PNG

tirr-c:
Fixes slight mismatch on alpha channel when decoding `candle.jxl` to PNG, listed in #330.

Now `magick identify -verbose` shows the exact same statistics on alpha channel as the image decoded with libjxl. @sboukortt can you confirm that the alpha channels match?


sboukortt:
It does, thanks! And `grayscale_patches_var_dct.jxl` is much improved as well.

PR #357, opened 2025-08-27
SIMDified gaborish.

zond:


jxl/src/render/stages/gaborish.rs
```diff
@@ -32,6 +33,32 @@ impl std::fmt::Display for GaborishStage {
...
+    ) {
+        for idx in (0..xsize).step_by(D::F32Vec::LEN) {
+            let mut sum = D::F32Vec::splat(d, 0f32);
+            let row_and_kernel = std::iter::zip(
+                rows_in,
+                [kernel_top_bottom, kernel_center, kernel_top_bottom],
+            );
+            for (row_in, kernel) in row_and_kernel {
+                for (dx, weight) in kernel.iter().enumerate() {
+                    sum += D::F32Vec::load(d, &row_in[idx + dx..]) * D::F32Vec::splat(d, *weight);
```
veluca93: Should probably be a muladd.

jxl/src/render/stages/gaborish.rs
```diff
@@ -47,26 +74,14 @@ impl RenderPipelineStage for GaborishStage {
...
-
-            for (row_in, kernel) in row_and_kernel {
-                for (dx, weight) in kernel.into_iter().enumerate() {
-                    sum += row_in[idx + dx] * weight;
-                }
-            }
-
-            *out = sum;
-        }
+        gaborish_process_dispatch(
```
veluca93: Can we test that scalar and SIMD have the same behaviour?

jxl/src/render/stages/gaborish.rs
zond: Done

jxl/src/render/stages/gaborish.rs
zond: Done

jxl/src/render/stages/gaborish.rs
```diff
@@ -105,4 +117,42 @@ mod test {
 
         Ok(())
     }
+
+    fn gaborish_process_scalar_equivalent<D: SimdDescriptor>(d: D) {
+        arbtest::arbtest(|u| {
+            let stage = GaborishStage::new(0, 0.115169525, 0.061248592);
+            let xsize: usize = 1 + u.arbitrary::<usize>()? % 4096;
+            let mut in_out_rows = vec![(
+                vec![vec![0f32; round_up_size_to_two_cache_lines::<f32>(xsize) + 3]; 3],
```
veluca93: That +3 seems misplaced.

jxl/src/util/test.rs
```diff
@@ -196,9 +223,29 @@ pub fn read_pfm(b: &[u8]) -> Result<Vec<Image<f32>>, Error> {
     Ok(res)
 }
 
-pub(crate) use assert_all_almost_eq;
+macro_rules! with_in_out_rows {
+    ($data:expr, |$slice:ident| $body:expr) => {{
```
veluca93: I'm pretty sure you can do a macro that can be used like this:

```
in_out_rows!(name_of_the_variable, u_from_arbitrary, border_pixels, num_channels);

// do stuff with `name_of_the_variable`.
```

something like:

```
macro_rules! in_out_rows {
   ($name: ident, $u: expr, $border: expr, $chan: literal) => {
      // define intermediate variables
      let $name = [...]
   }
}
```

I think this would be easier to read, and names defined in macros don't collide with macro callers, so it shouldn't cause issues.

Also, this macro should very likely be in the render pipeline test utils.

veluca93: (maybe you could also add an optional maximum xsize, defaulting to 4096)

jxl/src/render/stages/gaborish.rs
zond: Ah, it should be a 2 due to the borders - not a 3 - IIUC.

veluca93: Even then, it should be *inside* the call to rounding.

zond: I kind of agree - but when it's _inside_ it always fails with `assertion failed: mem.len() >= Self::LEN`, but when it's outside it works... uh.

jxl/src/util/test.rs
zond: I did most of it, except Just set the max size to 4096. Making it optional can be done later.

jxl/src/render/stages/gaborish.rs
veluca93: mhhh, I'd say let's leave a TODO to investigate then.

veluca93:
Can you split this into two PRs, one for changing SIMD tests and one for gaborish?

jxl/src/render/stages/gaborish.rs
```diff
@@ -105,4 +114,29 @@ mod test {
...
+                .iter()
+                .map(|inner_slice| inner_slice.to_vec())
+                .collect();
+            gaborish_process(ScalarDescriptor::new().unwrap(), &stage, xsize, rows);
+            let scalar_result: Vec<Vec<f32>> = rows[0]
+                .1
+                .iter()
+                .map(|inner_slice| inner_slice.to_vec())
+                .collect();
+            for (index, scalar_row) in scalar_result.iter().enumerate() {
```
veluca93, line 134: Shouldn't this only go up to xsize?

jxl/src/render/stages/gaborish.rs
zond: Fixed as per offline discussion.

PR #358, opened 2025-08-28
Added tests for individual SIMD instructions.

zond:
Also added support for relative error in assert_all_almost_eq!.

### DIFF
--- a/jxl/src/simd/mod.rs
+++ b/jxl/src/simd/mod.rs
@@ -77,138 +77,145 @@ pub trait F32SimdVec:
 
 #[cfg(test)]
 mod test {
+    use arbtest::arbitrary::Unstructured;
+
     use crate::{
-        simd::{
-            F32SimdVec, ScalarDescriptor, SimdDescriptor, round_up_size_to_two_cache_lines,
-            test_all_instruction_sets,
-        },
+        simd::{F32SimdVec, ScalarDescriptor, SimdDescriptor, test_all_instruction_sets},
         util::test::assert_all_almost_eq,
     };
 
-    fn vec_add<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
-        assert!(a.len() == b.len());
-        let mut res = vec![0f32; a.len()];
-        for idx in (0..a.len()).step_by(D::F32Vec::LEN) {
-            (D::F32Vec::load(d, &a[idx..]) + D::F32Vec::load(d, &b[idx..])).store(&mut res[idx..]);
-        }
-        res
+    enum Distribution {
+        Floats,
+        NonZeroFloats,
     }
 
-    fn vec_sub<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
-        assert!(a.len() == b.len());
-        let mut res = vec![0f32; a.len()];
-        for idx in (0..a.len()).step_by(D::F32Vec::LEN) {
-            (D::F32Vec::load(d, &a[idx..]) - D::F32Vec::load(d, &b[idx..])).store(&mut res[idx..]);
-        }
-        res
-    }
-
-    fn vec_mul<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
-        assert!(a.len() == b.len());
-        let mut res = vec![0f32; a.len()];
-        for idx in (0..a.len()).step_by(D::F32Vec::LEN) {
-            (D::F32Vec::load(d, &a[idx..]) * D::F32Vec::load(d, &b[idx..])).store(&mut res[idx..]);
-        }
-        res
-    }
-
-    fn vec_add_assign<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
-        assert!(a.len() == b.len());
-        let mut res = vec![0f32; a.len()];
-        for idx in (0..a.len()).step_by(D::F32Vec::LEN) {
-            let mut a_vec = D::F32Vec::load(d, &a[idx..]);
-            a_vec += D::F32Vec::load(d, &b[idx..]);
-            a_vec.store(&mut res[idx..]);
-        }
-        res
-    }
-
-    fn vec_sub_assign<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
-        assert!(a.len() == b.len());
-        let mut res = vec![0f32; a.len()];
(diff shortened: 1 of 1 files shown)


jxl/src/simd/mod.rs
```diff
@@ -85,6 +87,70 @@ mod test {
         util::test::assert_all_almost_eq,
     };
 
+    fn arb_vec<D: SimdDescriptor>(_: D, u: &mut Unstructured) -> Vec<f32> {
+        let size = round_up_size_to_two_cache_lines::<f32>(D::F32Vec::LEN);
```
veluca93: Why are we rounding things up here? We only need a single vector.

jxl/src/simd/mod.rs
```diff
@@ -85,6 +87,70 @@ mod test {
...
+        (
+            $name:ident
+        ) => {
+            paste::paste! {
+                fn [<$name _test>]<D: SimdDescriptor>(d: D) {
+                    arbtest::arbtest(|u| {
+                        let a = arb_vec(d, u);
+                        assert_all_almost_eq!($name(ScalarDescriptor::new().unwrap(), &a), $name(d, &a), 1, 1e-8);
+                        Ok(())
+                    }).size_min(1<<16);
```
veluca93: And then we wouldn't need the size_min either.

jxl/src/simd/mod.rs
```diff
@@ -102,15 +179,17 @@ mod test {
         }
         res
     }
+    test_arity_2_instruction!(vec_sub);
 
-    fn vec_mul<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
+    fn vec_div<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
```
veluca93: Intuitively, it feels that we could be moving most of this into the macro. 

For example, the macro call could look like this:

```
test_binary_instruction!(vec_div, D, |a: D::F32Vec, b: D::F32Vec| { a / b });
```
which I think would overall be significantly more readable.

veluca93: Doing this would also allow us to get rid of the calls to paste! in the macro, as we could just pass `vec_div_test` to the macro.

jxl/src/util/test.rs
```diff
@@ -85,15 +84,43 @@ macro_rules! assert_all_almost_eq {
...
+            if matches!(error.partial_cmp(&max_error), Some(std::cmp::Ordering::Greater) | None) {
                 panic!(
                     "assertion failed: `(left ≈ right)`\n left: `{:?}`,\n right: `{:?}`,\n max_error: `{:?}`,\n left[{}]: `{}`,\n right[{}]: `{}`",
-                    left_val, right_val, max_error, index, left_val[index], index, right_val[index]
+                    left_val, right_val, max_error, index, left_elem, index, right_elem
+                )
+            }
+        }
+    };
+    ($left:expr, $right:expr, $max_error_abs:expr, $max_error_rel:expr $(,)?) => {
```
veluca93: I am not sure I am seeing why these are macros: wouldn't be able to do the same with a regular function call?

A macro could be helpful if we replaced `left` and `right` with `stringify!($left)` and `stringify!($right)`, but even then most of the code could be moved out of the macro and into a function.

jxl/src/simd/mod.rs
zond: Done

jxl/src/simd/mod.rs
zond: We still need size_min, the random allocation for an arbtest isn't even enough for 2 whole lanes :(

jxl/src/simd/mod.rs
zond: Done

jxl/src/util/test.rs
zond: Remove this change, will replace the eq assertion macros in a later PR.

jxl/src/simd/mod.rs
```diff
@@ -77,138 +77,125 @@ pub trait F32SimdVec:
...
-            let mut a_vec = D::F32Vec::load(d, &a[idx..]);
-            a_vec += D::F32Vec::load(d, &b[idx..]);
-            a_vec.store(&mut res[idx..]);
-        }
-        res
-    }
+    test_instruction!(add, |a, b| { a + b });
+    test_instruction!(mul, |a, b| { a * b });
+    test_instruction!(sub, |a, b| { a - b });
+    test_instruction!(div, |a, b| { a / b });
```
veluca93: This test will fail if a and b are both 0

jxl/src/simd/mod.rs
```diff
@@ -77,138 +77,125 @@ pub trait F32SimdVec:
...
-    fn vec_sub<D: SimdDescriptor>(d: D, a: &[f32], b: &[f32]) -> Vec<f32> {
-        assert!(a.len() == b.len());
-        let mut res = vec![0f32; a.len()];
-        for idx in (0..a.len()).step_by(D::F32Vec::LEN) {
-            (D::F32Vec::load(d, &a[idx..]) - D::F32Vec::load(d, &b[idx..])).store(&mut res[idx..]);
+    fn arb_vec<D: SimdDescriptor>(_: D, u: &mut Unstructured) -> Vec<f32> {
+        let mut res = vec![0.0; D::F32Vec::LEN];
+        for v in res.iter_mut() {
+            *v =
+                u.arbitrary::<i32>().unwrap() as f32 / (1.0 + u.arbitrary::<u32>().unwrap() as f32);
```
veluca93: Why are elements generated like this?

veluca93 (approved):

jxl/src/simd/mod.rs
zond: It's a way to generate arbitrary floats without using arbitrary::<f32>(). What would you prefer?

jxl/src/simd/mod.rs
zond: Since a and b are both generated floats, they should ~never be 0, or am I misunderstanding the question?

jxl/src/simd/mod.rs
veluca93: Good enough.

jxl/src/simd/mod.rs
veluca93: It will happen eventually :-)

zond: Alright, now b will not be zero :D

zond:
> FYI, you can most likely use macro argument repetition to handle an arbitrary number of arguments and reduce the code size, but this is fine for now.

Sounds good - this was more than enough macro horror for today :P

PR #359, opened 2025-08-28
Limiting comparison to xsize.

zond:


veluca93 (approved):

zond:
@veluca93 take another look - I added a double-rounding in the simple pipeline to avoid an arbtest failure caused by chunk size unfortunate enough to cause overread.

veluca93 (approved):

PR #361, opened 2025-08-30
Added test for decoding small chunks.

zond:


zond:
I'm not sure this is the best way to fix the issue - please take an extra look.

jxl/src/api/inner/codestream_parser.rs
```diff
@@ -168,6 +167,18 @@ impl CodestreamParser {
...
+                        .sections
+                        .front()
+                        .is_some_and(|s| s.len <= self.ready_section_data)
+                    {
+                        let s = self.sections.pop_front().unwrap();
+                        self.ready_section_data -= s.len;
+                        self.available_sections.push(s);
+                    }
+                    if self.available_sections.is_empty() {
+                        return Err(crate::error::Error::OutOfBounds(0));
```
veluca93: I assume the intention of this change is to return a request for more data to the user if no sections are available.

If so, what I think we should rather be doing is have process_sections return one of:
- Ok(None) ("I processed all the sections I could")
- Ok(Some(usize)) ("I need at least this much data to process the next section")
- Err(some error), mapping Error::OutOfBounds() to Error::SectionTooShort as is done now.

Then, we can discard Ok(None) and map Ok(Some(usize)) into a Error::OutOfBounds()

zond: I wanted to do that before calling process sections, but couldn't figure out how to compute the amount of available data - it wasn't ready_sections_data, and it wasn't sum of section data length. What's the correct calculation?

veluca93: The amount of available section data is ready_sections_data.

zond: Done

jxl/src/api/inner/codestream_parser/sections.rs
```diff
@@ -56,7 +56,9 @@ impl CodestreamParser {
             self.available_sections.push(s);
         }
         if self.available_sections.is_empty() {
-            return Ok(());
+            return Ok(Some(
+                self.sections.front().unwrap().len - self.ready_section_data,
```
veluca93, line 60: I believe this should be a saturating_sub.

zond: Really?
At this point we know sections isn't empty, and we have no sections with len <= ready section data, right?

veluca93: and not unwrap() but `.map(|x| x.len).unwrap_or_default(0)`

veluca93: (I could be wrong though, it could be that self.sections is not empty and self.ready_section_data is greater than len of the first section unless there is a bug)

zond: Yeah, that's what I think, but maybe it pays off to be defensive?

veluca93: I think you're right, let's leave this as is unless it does indeed crash sometimes

PR #365, opened 2025-09-03
Added a speedtest mode to jxl_cli.

zond:


jxl_cli/src/main.rs
```diff
@@ -353,25 +361,44 @@ fn with_api(opt: Opt) -> Result<()> {
...
+        let output_bit_depth = match opt.override_bitdepth {
+            None => original_bit_depth.bits_per_sample(),
+            Some(num_bits) => num_bits,
+        };
+        let image_result = Some(save_image(
+            image_data,
+            output_bit_depth,
+            &output_profile,
+            &opt.output,
+        ));
```
sboukortt: This shadows the previous `image_result` instead of updating it. It will stay `None` on line 392.

jxl_cli/src/main.rs
```diff
@@ -353,25 +361,44 @@ fn with_api(opt: Opt) -> Result<()> {
...
-    if let Err(ref err) = image_result {
-        println!("Failed to save image: {err}");
+    let image_result: Option<Result<()>> = None;
+    if !opt.speedtest {
+        let output_bit_depth = match opt.override_bitdepth {
+            None => original_bit_depth.bits_per_sample(),
+            Some(num_bits) => num_bits,
+        };
+        let image_result = Some(save_image(
+            image_data,
```
sboukortt: Something like this should probably do:

```suggestion
    let image_result: Option<Result<()>>;
    if opt.speedtest {
        image_result = None;
    } else {
        let output_bit_depth = match opt.override_bitdepth {
            None => original_bit_depth.bits_per_sample(),
            Some(num_bits) => num_bits,
        };
        image_result = Some(save_image(
            image_data,
```

jxl_cli/src/main.rs
zond: I refactored the whole thing to be much cleaner. Take another look.

veluca93 (approved):

PR #371, opened 2025-09-08
Improved render stage tests

zond:


jxl/src/render/stages/patches.rs
```diff
@@ -58,6 +58,65 @@ impl RenderPipelineStage for PatchesStage {
...
+        PatchBlendMode, PatchBlending, PatchPosition, PatchReferencePosition, PatchTreeNode,
+    };
+    use crate::util::test::read_headers_and_toc;
+
     #[test]
-    fn process_row_chunk() {}
+    fn patches_consistency() -> Result<()> {
+        let (file_header, _, _) =
+            read_headers_and_toc(include_bytes!("../../../resources/test/basic.jxl")).unwrap();
+        let patch_dict = PatchesDictionary {
```
veluca93: I would prefer if instead we had a cfg(test) PatchesDictionary::random constructor (and we didn't change the visibility), can we do that?

zond: How about this?

jxl/src/frame.rs
```diff
@@ -105,6 +105,23 @@ impl ReferenceFrame {
...
     }
+    #[cfg(test)]
+    pub fn random(
+        width: usize,
+        height: usize,
+        num_channels: usize,
+        saved_before_color_transform: bool,
+    ) -> Result<Self> {
+        use rand::SeedableRng;
+        let mut rng = rand_xorshift::XorShiftRng::seed_from_u64(0);
```
veluca93: Let's pass the RNG in too.

veluca93: (in particular, as of now, this will make the 4 reference frames identical, which wouldn't catch some errors)

veluca93 (approved):

jxl/src/frame.rs
zond: Done

PR #375, opened 2025-09-08
Hid the std::fs code in jxl_macros behind a feature.

zond:


jxl/Cargo.toml
```diff
@@ -26,6 +26,7 @@ paste = "1.0.15"
 rand = "0.8.5"
 rand_xorshift = "0.3.0"
 test-log = { version = "0.2.16", features = ["trace"] }
+jxl_macros = { path = "../jxl_macros", version = "=0.1.0", features = ["test"] }
```
veluca93: I think you can remove path and set version="*" here, can you try?

jxl_macros/src/lib.rs
```diff
@@ -5,13 +5,16 @@
...
+#[cfg(feature = "test")]
 use std::{fs, path::Path};
 
 use proc_macro::TokenStream;
 use proc_macro_error2::{abort, proc_macro_error};
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
-use syn::{DeriveInput, Ident, Meta, parse_macro_input};
+#[cfg(feature = "test")]
+use syn::Ident;
```
veluca93: Should we move these use declarations at function level?

zond: Done

jxl/Cargo.toml
zond: I could set version = "*", but I couldn't remove path.

```
error: failed to load manifest for workspace member `/usr/local/google/home/zond/projects/jxl-rs/jxl`
referenced by workspace at `/usr/local/google/home/zond/projects/jxl-rs/Cargo.toml`

Caused by:
  failed to parse manifest at `/usr/local/google/home/zond/projects/jxl-rs/jxl/Cargo.toml`

Caused by:
  Dependency 'jxl_macros' has different source paths depending on the build target. Each dependency must have a single canonical source path irrespective of build target.
```

jxl_macros/src/lib.rs
```diff
@@ -705,8 +703,13 @@ pub fn noop(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
 
+#[cfg(feature = "test")]
 #[proc_macro]
 pub fn for_each_test_file(input: TokenStream) -> TokenStream {
+    #[cfg(feature = "test")]
```
veluca93: Don't need these :-)

zond: Oops. Derp :P

PR #378, opened 2025-09-08
Use JxlOutputBuffer in SaveStage

veluca93:
- Make SaveStage no longer be generic on the input type.
- Pass down buffers to SaveStage.
- Change the interface to the pipeline to not need to name SaveStage.
- Implement writing the various supported datatypes in SaveStage.
- Improve tests.

### DIFF
--- a/jxl/src/api/output.rs
+++ b/jxl/src/api/output.rs
@@ -129,7 +129,7 @@ impl<'a> JxlOutputBuffer<'a> {
         }
     }
 
-    pub(crate) fn from_image<T: ImageDataType>(image: &'a mut Image<T>) -> Self {
+    pub fn from_image<T: ImageDataType>(image: &'a mut Image<T>) -> Self {
         let (width, height) = image.size();
         let buf = image.buf_mut();
         assert!(buf.len() >= width * height);
--- a/jxl_cli/src/enc/numpy.rs
+++ b/jxl_cli/src/enc/numpy.rs
@@ -68,7 +68,7 @@ fn numpy_bytes(image_data: ImageData<f32>, num_channels: usize) -> Vec<u8> {
                         (0..num_channels).map(move |c| channel_rects[c].row(y)[x])
                     })
                 })
-                .flat_map(|x| (x.clamp(0.0, 255.0) / 255.0f32).to_le_bytes()),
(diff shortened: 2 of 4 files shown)


jxl/src/api/output.rs
```diff
@@ -122,52 +120,43 @@ impl<'a> JxlOutputBuffer<'a> {
...
-            for (out, v) in row.chunks_exact_mut(4).zip(crow) {
-                let v = v.to_ne_bytes();
-                for i in 0..4 {
-                    out[i].write(v[i]);
-                }
-            }
+    pub fn from_image<T: ImageDataType>(image: &'a mut Image<T>) -> Self {
+        let (width, height) = image.size();
+        let buf = image.buf_mut();
+        assert!(buf.len() >= width * height);
```
szabadka, line 135: Shouldn't this assertion be somewhere in the Image class?

jxl/src/render/simple_pipeline/save.rs
```diff
@@ -3,374 +3,282 @@
...
-        let (w_src, h_src) = if self.orientation.is_transposing() {
-            (out_h, out_w) // Swapped
+        data: &[Image<f64>],
+        buffers: &mut [Option<JxlOutputBuffer>],
+    ) -> Result<()> {
+        for i in self.channels.iter().skip(1) {
+            assert_eq!(data[self.channels[0]].size(), data[*i].size());
+        }
+        let Some(buf) = buffers[self.output_buffer_index].as_mut() else {
+            return Ok(());
```
szabadka, line 72: why is it not an error?

szabadka (approved):

jxl/src/api/output.rs
veluca93: Even if it were there, having it here is useful to locally ensure correctness of the unsafe Self::new_from_ptr call below.

jxl/src/render/simple_pipeline/save.rs
veluca93: Because the user might decide to not provide buffers for some calls (i.e. because they are trying to get jxl-info-like functionality), and we already check elsewhere that they provide the correct number of buffers if they do provide them.

PR #389, opened 2025-09-18
Skip computing the output of the weighted predictor if unused

sboukortt:


veluca93 (approved):

jxl/src/frame/coeff_order.rs
```diff
@@ -77,7 +77,7 @@ pub fn natural_coeff_order(transform: HfTransformType) -> Vec<u32> {
         for j in 0..(i + 1) {
             let mut x = xsize - 1 - (i - j);
             let mut y = xsize - 1 - j;
-            if i % 2 != 0 {
+            if i.is_multiple_of(2) {
```
chicheng, line 80: changed from odd=True to even=True?

sboukortt: Oops… thanks. #390 

(Worrying that no test caught this, and odd that Clippy doesn’t detect the other instance, 24 lines above.)

PR #394, opened 2025-10-08
Prepare for SIMD DCT

sboukortt:


jxl/src/simd/x86_64/avx.rs
```diff
@@ -52,6 +52,15 @@ macro_rules! fn_avx {
 #[repr(transparent)]
 pub struct F32VecAvx(__m256, AvxDescriptor);
 
+impl F32VecAvx {
+    #[inline(always)]
+    unsafe fn get_mask(size: usize) -> __m256i {
+        const MASKS: [i32; 16] = [0, 0, 0, 0, 0, 0, 0, 0, -1, -1, -1, -1, -1, -1, -1, -1];
+        // SAFETY: this is only called from code that has avx available.
+        unsafe { _mm256_loadu_epi32(MASKS.as_ptr().add(size)) }
```
tirr-c: I think this should be reversed?

```suggestion
        const MASKS: [i32; 16] = [-1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0, 0];
        // SAFETY: this is only called from code that has avx available.
        unsafe { _mm256_loadu_epi32(MASKS.as_ptr().add(8 - size)) }
```

tirr-c: And [the guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_loadu_epi32&ig_expand=4061) says `_mm256_loadu_epi32` is added in AVX512, so you might need `_mm256_loadu_si256` which should work the same...

sboukortt: Right, thanks; that order was a remnant from when the mask was constructed using `_mm256_set_epi32`, which takes them in the opposite order. But we might eventually change to using a small intermediary buffer anyway.

jxl/src/simd/x86_64/avx.rs
```diff
@@ -26,6 +27,40 @@ impl AvxDescriptor {
     }
 }
 
+impl AvxDescriptor {
+    #[inline(always)]
+    fn transpose4x4(
```
veluca93: transpose4x4f32

jxl/src/simd/x86_64/avx.rs
```diff
@@ -26,6 +27,40 @@ impl AvxDescriptor {
...
+        self,
+        input: &[f32],
+        input_stride: usize,
+        output: &mut [f32],
+        output_stride: usize,
+    ) {
+        // SAFETY: AvxDescriptor can only be constructed if avx2 is available.
```
veluca93: we know avx is available from the safety invariant on `self`.

veluca93: I would move this into a "_impl" function with `#[target_feature(enable = "avx2")]`, and then justify the remaining unsafe operations individually.

jxl/src/simd/x86_64/avx.rs
```diff
@@ -26,6 +27,40 @@ impl AvxDescriptor {
...
+        unsafe {
+            let p0 = _mm_loadu_ps(input.as_ptr());
+            let p1 = _mm_loadu_ps(input[input_stride..].as_ptr());
+            let p2 = _mm_loadu_ps(input[2 * input_stride..].as_ptr());
+            let p3 = _mm_loadu_ps(input[3 * input_stride..].as_ptr());
```
veluca93: Can we assert that `input_stride >= 4` and that `input >= 3 * input_stride + 4` (and also that the intermediate calculation does not overflow), and then use `.as_ptr().add(...)`, with appropriate comments on the safety of the pointer arithmetic?

jxl/src/simd/x86_64/avx.rs
```diff
@@ -26,6 +27,40 @@ impl AvxDescriptor {
...
+
+            let r0 = _mm_unpacklo_ps(q0, q1);
+            let r1 = _mm_unpackhi_ps(q0, q1);
+            let r2 = _mm_unpacklo_ps(q2, q3);
+            let r3 = _mm_unpackhi_ps(q2, q3);
+
+            _mm_storeu_ps(output.as_mut_ptr(), r0);
+            _mm_storeu_ps(output[output_stride..].as_mut_ptr(), r1);
+            _mm_storeu_ps(output[2 * output_stride..].as_mut_ptr(), r2);
+            _mm_storeu_ps(output[3 * output_stride..].as_mut_ptr(), r3);
```
veluca93: Same here.

jxl/src/simd/x86_64/avx.rs
```diff
@@ -60,6 +114,15 @@ macro_rules! fn_avx {
 #[repr(transparent)]
 pub struct F32VecAvx(__m256, AvxDescriptor);
 
+impl F32VecAvx {
+    #[inline(always)]
+    unsafe fn get_mask(size: usize) -> __m256i {
+        const MASKS: [i32; 15] = [-1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0];
+        // SAFETY: this is only called from code that has avx available.
```
veluca93: I would make this just a `#[target_feature(enable = "avx")]` function; it does not need to be unsafe and probably should not be an associated function of F32VecAvx; alternatively, you can make it take a AvxDescriptor,

The safety comment needs to mention the pointer arithmetic below.

jxl/src/simd/x86_64/avx.rs
```diff
@@ -73,13 +136,33 @@ impl F32SimdVec for F32VecAvx {
         Self(unsafe { _mm256_loadu_ps(mem.as_ptr()) }, d)
     }
 
+    #[inline(always)]
+    fn load_partial(d: Self::Descriptor, size: usize, mem: &[f32]) -> Self {
+        assert!(Self::LEN >= size);
```
veluca93: (this *could* be a debug_assert, as it is not safety-critical)

jxl/src/simd/x86_64/avx.rs
sboukortt: Sorry, move what specifically? This function is already an implementation detail for the broader `transpose`.

jxl/src/simd/x86_64/avx.rs
sboukortt: Adding the asserts make sense. Why `.add` over slices though?

veluca93: Saves bounds checks :-)

jxl/src/simd/x86_64/avx.rs
veluca93: Then I'd just make this `#[target_feature(enable = "avx2")]`.

veluca93: (this implies that the unpack ops would not need unsafe anymore)

sboukortt:
Thanks, how is this?

jxl/src/simd/x86_64/avx.rs
```diff
@@ -60,6 +136,20 @@ macro_rules! fn_avx {
...
 pub struct F32VecAvx(__m256, AvxDescriptor);
 
+#[target_feature(enable = "avx2")]
+#[inline]
+fn get_mask(size: usize) -> __m256i {
+    assert!(size > 0);
+    const MASKS: [i32; 15] = [-1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0];
+    // SAFETY: this is only called from code that has avx available.
+    // The pointer arithmetic is safe because `size` is at least one, so:
+    // `size - 1` does not underflow;
```
veluca93: We don't actually need size-1 to not underflow (underflow is not UB, and even if it does (size-1) & 7 is still between 0 and 7).

I would remove the assert and the sentence.

jxl/src/simd/x86_64/avx.rs
```diff
@@ -60,6 +136,20 @@ macro_rules! fn_avx {
 #[repr(transparent)]
 pub struct F32VecAvx(__m256, AvxDescriptor);
 
+#[target_feature(enable = "avx2")]
+#[inline]
+fn get_mask(size: usize) -> __m256i {
+    assert!(size > 0);
+    const MASKS: [i32; 15] = [-1, -1, -1, -1, -1, -1, -1, -1, 0, 0, 0, 0, 0, 0, 0];
+    // SAFETY: this is only called from code that has avx available.
```
veluca93: If we actually had this as a safety requirement, this function should be unsafe -- fortunately we don't require it because of the `target_feature` attribute, so we can as well remove this part of the comment.

jxl/src/simd/x86_64/avx.rs
sboukortt: Done, thanks.

PR #396, opened 2025-10-16
NOT UPSTREAMABLE: move to APIs supported by MSRV of 1.85

ErichDonGubler:


ErichDonGubler:
Oh, yikes, I'd meant to file this as a PR against my own fork, so that Mozilla would have a stable point from which it could experiment with it. I'll move this over there, sorry for the noise!

EDIT: Moved to https://github.com/erichdongubler-mozilla/jxl-rs/pull/1.

PR #398, opened 2025-10-20
SIMDfy dequant_lanes

sboukortt:
Before: `dequant_and_transform_to_pixels_dispatch` took 2.2-2.3% of the total time according to `perf`

After: 1.7-1.8%

(Both figures with the simple pipeline)

### DIFF
--- a/jxl/src/simd/mod.rs
+++ b/jxl/src/simd/mod.rs
@@ -30,6 +30,10 @@ pub(crate) use scalar::ScalarDescriptor;
 pub trait SimdDescriptor: Sized + Copy + Debug + Send + Sync {
     type F32Vec: F32SimdVec<Descriptor = Self>;
 
+    type I32Vec: I32SimdVec<Descriptor = Self, F32Vec = Self::F32Vec, Mask = Self::Mask>;
+
+    type Mask: SimdMask<Descriptor = Self, F32Vec = Self::F32Vec, I32Vec = Self::I32Vec>;
+
     fn new() -> Option<Self>;
 
     fn transpose<const ROWS: usize, const COLS: usize>(self, input: &[f32], output: &mut [f32]);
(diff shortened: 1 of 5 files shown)


jxl/src/simd/x86_64/avx512.rs
```diff
@@ -222,3 +227,98 @@ impl DivAssign<F32VecAvx512> for F32VecAvx512 {
...
+    fn splat(d: Self::Descriptor, v: i32) -> Self {
+        // SAFETY: We know avx512f is available from the safety invariant on `d`.
+        unsafe { Self(_mm512_set1_epi32(v), d) }
+    }
+
+    fn_avx!(this: I32VecAvx512, fn as_f32() -> F32VecAvx512 {
+         F32VecAvx512(_mm512_cvtepi32_ps(this.0), this.1)
+    });
+
+    fn_avx!(this: I32VecAvx512, fn abs() -> I32VecAvx512 {
```
veluca93, line 260: I don't believe this is correct :-) But there's _mm512_abs_epi32 & friends.

jxl/src/simd/mod.rs
```diff
@@ -96,6 +100,50 @@ pub trait F32SimdVec:
...
+
+    fn as_f32(self) -> Self::F32Vec;
+
+    fn gt(self, other: Self) -> Self::Mask;
+}
+
+pub trait SimdMask: Sized + Copy + Debug + Send + Sync {
+    type Descriptor: SimdDescriptor;
+
+    type F32Vec: F32SimdVec;
```
veluca93, line 134: Why do we have those? Couldn't we use `Self::Descriptor::F32Vec` instead?

(and/or have `type Descriptor: SimdDescriptor<Mask = Self>`)

veluca93: (same for I32)

veluca93 (approved):

jxl/src/simd/x86_64/avx512.rs
sboukortt: Oops 😂 not sure what I was thinking. (I’m guessing “not”.) Thanks.

jxl/src/simd/mod.rs
sboukortt: Sadly not, as far as I can tell:

        fn as_f32(self) -> Self::Descriptor::F32Vec;

    ambiguous associated type [E0223]
    Help: if there were a trait named `Example` with associated type `F32Vec` implemented for `<Self as I32SimdVec>::Descriptor`, you could use the fully-qualified path

 (I don’t understand what `Example` has to do with it in the help message)

veluca93: Ah, you'd need to use the fully qualified path, which gets fairly verbose fairly quickly... I think it's fine to leave it as is, at least for now.

PR #399, opened 2025-10-23
Less copying around in DCT

sboukortt:


jxl/src/var_dct/dct.rs
```diff
@@ -313,48 +284,20 @@ where
 {
     assert_eq!(data.len(), ROWS * COLS, "Data length mismatch");
 
-    // Create a temporary flat buffer for the transposed data.
-    let mut transposed_data = vec![0.0f32; ROWS * COLS];
+    // Create a temporary buffer for the transposed data.
```
veluca93, line 287: we should probably pass a temporary buffer in as we do in libjxl, btw

veluca93 (approved):

jxl/src/var_dct/dct.rs
sboukortt: Ah, you mean a preallocated “scratch” buffer like we have in various places?

sboukortt:
(I managed to remove a couple more instances thanks to `.as_chunks`.)

jxl/src/var_dct/dct.rs
veluca93: yup

PR #405, opened 2025-10-27
Use padded frame size in PatchesDictionary::read

mo271:
This aligns patch boundary checks with libjxl by using header.size_padded() to prevent PatchesOutOfBounds errors on images with non-aligned dimensions.


mo271:
waiting to merge this until I have added a test image that triggers this in another pull request (in order to see failing ci there)

PR #406, opened 2025-10-28
test image for patches out of bound

mo271:
should exhibit a failure when decoding that is fixed in #405.

This was produced with a manually modified version of `cjxl`:
```
diff --git a/lib/jxl/enc_patch_dictionary.cc b/lib/jxl/enc_patch_dictionary.cc
index e76d7cc7..bb2bb77b 100644
--- a/lib/jxl/enc_patch_dictionary.cc
+++ b/lib/jxl/enc_patch_dictionary.cc
@@ -631,10 +631,25 @@ Status FindBestPatchDictionary(const Image3F& opsin,
     Rect rect(0, 0, state->shared.frame_dim.xsize,
               state->shared.frame_dim.ysize);
     JXL_ASSIGN_OR_RETURN(info,
+\
                          FindDotDictionary(state->cparams, opsin, rect,
                                            state->shared.cmap.base(), pool));
   }
 
+  // Force a patch at (0, 13) 
+  if (state->cparams.patches == Override::kOn) {
+    info.emplace_back();
+    info.back().second.emplace_back(0, 13);
+    QuantizedPatch& patch = info.back().first;
+    patch.xsize = 1;
+    patch.ysize = 1;
+    for (size_t c = 0; c < 3; ++c) {
+      patch.fpixels[c][0] = 0.1f;  // Arbitrary non-zero value
+      patch.pixels[c][0] = 1;
+    }
+  }
+  
+
   if (info.empty()) return true;
 
```
and then calling 
```
cjxl test_image16x13.ppm patch_y_out_of_bounds.jxl --patches=1 -e 9
```

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -159,8 +159,8 @@ impl Frame {
             info!("decoding patches");
             Some(PatchesDictionary::read(
                 br,
-                self.header.width as usize,
-                self.header.height as usize,
+                self.header.size_padded().0,
+                self.header.size_padded().1,
                 self.decoder_state.extra_channel_info().len(),
                 &self.decoder_state.reference_frames,
             )?)


mo271:
fails as expected before merging #405 with
```
---- api::decoder::tests::decode_test_file_patch_y_out_of_bounds stdout ----

thread 'api::decoder::tests::decode_test_file_patch_y_out_of_bounds' panicked at jxl/src/api/decoder.rs:279:39:
called `Result::unwrap()` on an `Err` value: PatchesOutOfBounds("y", 13, 1, 13)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    api::decoder::tests::decode_test_file_patch_y_out_of_bounds

```

PR #409, opened 2025-10-29
Unify the JxlOutputBuffer type with the Image type.

veluca93:
They now belong to the same hierarchy of Image types.

This introduces new RawImage types, that store data independently of the actual data type used. This will be useful as a replacement for RowBuffer and for storing pipeline input without needing to generate code specific to every data type.

The underlying implementation is somewhat more efficient, giving an approximate 5% speed increase.


jxl/src/image/data_type.rs
```diff
@@ -0,0 +1,185 @@
...
+// license that can be found in the LICENSE file.
+
+use std::fmt::Debug;
+
+mod private {
+    pub trait Sealed {}
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum DataTypeTag {
```
szabadka, line 13: maybe add a comment here that for any new data type tag the ImageDataType  trait safety requirements must be asserted below.

jxl/src/image/data_type.rs
```diff
@@ -0,0 +1,185 @@
...
+};
+
+/// # Safety
+/// Any type implementing this trait must be a "bag-of-bits" type with no padding.
+/// Moreover, Self::DATA_TYPE_ID.size() must match the size and be a multiple of the alignment of
+/// `Self`.
+pub unsafe trait ImageDataType:
+    private::Sealed + Copy + Default + 'static + Debug + PartialEq + Send + Sync
+{
+    /// ID of this data type. Different types *must* have different values.
```
anforowicz: nit: Can the comment clarify if this is a _safety_ requirement, or merely a correctness requirement?

jxl/src/image/data_type.rs
```diff
@@ -0,0 +1,185 @@
...
+    assert!(std::mem::size_of::<u16>() == DataTypeTag::U16.size());
+    assert!(std::mem::size_of::<half::f16>() == DataTypeTag::F16.size());
+    assert!(std::mem::size_of::<i32>() == DataTypeTag::I32.size());
+    assert!(std::mem::size_of::<u32>() == DataTypeTag::U32.size());
+    assert!(std::mem::size_of::<f32>() == DataTypeTag::F32.size());
+    assert!(std::mem::size_of::<f64>() == DataTypeTag::F64.size());
+};
+
+/// # Safety
+/// Any type implementing this trait must be a "bag-of-bits" type with no padding.
```
anforowicz, line 51: I wonder if using `bytemuck::Pod` would be one way to avoid 1) introducing an `unsafe trait` and 2) providing `unsafe impl`s for primitive types.

OTOH, `bytemuck::Pod` may not cover all of your requirements (matching `Self::DATA_TYPE_ID.size()`).  Still, maybe removal of _some_ of the safety requirements is desirable?

jxl/src/image/data_type.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Moreover, Self::DATA_TYPE_ID.size() must match the size and be a multiple of the alignment of
```
anforowicz: The Rust Reference [says](https://doc.rust-lang.org/reference/type-layout.html#r-layout.properties.size:~:text=The%20size%20of%20a%20value%20is%20always%20a%20multiple%20of%20its%20alignment) that "the size of a value is always a multiple of its alignment".

szabadka: Shouldn't the size be equal to the alignment of `Self` ?

In the implementation of ImageRect::row() it says: "sizeof(T) == alignof(T) == T::DATA_TYPE_ID.size() by the requirements of ImageDataType"

jxl/src/image/internal.rs
```diff
@@ -0,0 +1,283 @@
...
+        assert!(rect.origin.1 < self.num_rows);
+        assert!(rect.origin.1.checked_add(rect.size.1).unwrap() <= self.num_rows);
+        assert!(rect.origin.0 < self.bytes_per_row);
+        assert!(rect.origin.0.checked_add(rect.size.0).unwrap() <= self.bytes_per_row);
+        // SAFETY: the safety invariant of `self`, together with the above check, guarantees that
+        // the calculation does not overflow and that the new pointer stays within the bounds of
+        // the allocation.
+        let start_ptr = unsafe {
+            self.buf
+                .add(rect.origin.1 * self.bytes_between_rows + rect.origin.0)
```
anforowicz, line 179: Is it expected that `...origin.1` is expressed in pixels, but `...origin.0` is expressed in bytes?  Should `...origin.0` be multiplied by `bytes_between_pixels` or `bytes_per_pixel`?

veluca93:
> Just some shallow, drive-by feedback.
> 
> I don't quite like that `RawImageBuffer` doesn't have an associated lifetime, but there is probably a good reason for that + I wouldn't know how to avoid that anyhow.

I am also not super enthusiastic about that, but given that it's a private type and that all usages are encapsulated in a type with proper lifetimes, I think it should be fine.

jxl/src/image/data_type.rs
veluca93: Clarified :-)

jxl/src/image/data_type.rs
veluca93: I have thought about bytemuck, but I think that in this case (in which we just handle primitive types) the best tradeoff is to not add a dependency, given that (a) it is fairly simple to verify the safety requirements and (b) it wouldn't reduce code much.

jxl/src/image/data_type.rs
veluca93: Oh, good point, that gets rid of some code.

jxl/src/image/internal.rs
veluca93: Yep, that is expected - the intention is for a RawImageBuf to be equivalent to an image of bytes.

There will be some places in which I'll use direct byte access regardless of the original type in follow-up PRs :-)

PR #410, opened 2025-10-29
test image for alpha with squeeze

mo271:


mo271:
should be rebased after the fix (#415) is merged

PR #411, opened 2025-10-29
Type-erased low_memory_pipeline and start SIMD save.

veluca93:
Faster saving is only implemented for 4-byte-per-pixel output and 3 channels, or 1 channel with any size.

Roughly 15-20% speedup.

### DIFF
--- a/jxl/src/render/low_memory_pipeline/save/x86_64.rs
+++ b/jxl/src/render/low_memory_pipeline/save/x86_64.rs
@@ -1,0 +1,168 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#![allow(unsafe_code)]
+
+use std::{
+    arch::x86_64::{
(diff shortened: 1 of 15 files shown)


jxl/src/render/low_memory_pipeline/save/x86_64.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    processed
+}
+
+/// Safety note: does not write uninit data in `out`.
+pub(super) fn interleave3_32b(inp: &[&[u8]; 3], out: &mut [MaybeUninit<u8>]) -> usize {
+    if is_x86_feature_detected!("avx512f") {
+        // SAFETY: we just checked for avx512f.
+        unsafe { interleave3_32b_avx512(inp, out) }
```
szabadka, line 161: why do we need the unsafe block here, since the function it calls is not unsafe

veluca93: `#[target_feature(enable = "...")]` functions are a bit special -- they are unsafe to be called from a function that does not have the same attribute.

PR #412, opened 2025-10-30
Speed up Gaborish.

veluca93:


sboukortt:
Out of curiosity, roughly how much is the speedup, if you have that at hand?

veluca93:
IIRC it went from 3-4% of the runtime to 1%.

PR #418, opened 2025-10-30
SIMDfy the gamma tf in From/ToLinear

sboukortt:
The remaining transfer functions are not done yet.

### DIFF
--- a/jxl/src/render/stages/from_linear.rs
+++ b/jxl/src/render/stages/from_linear.rs
@@ -6,6 +6,7 @@
 use crate::color::tf;
 use crate::headers::color_encoding::CustomTransferFunction;
 use crate::render::RenderPipelineInPlaceStage;
+use jxl_simd::{F32SimdVec, simd_function};
 
 /// Apply transfer function to display-referred linear color samples.
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


jxl/src/render/stages/from_linear.rs
```diff
@@ -106,12 +101,30 @@ impl RenderPipelineInPlaceStage for FromLinearStage {
             }
             TransferFunction::Gamma(g) => {
                 for row in row {
-                    for v in &mut row[..xsize] {
-                        *v = crate::util::fast_powf(v.abs(), g).copysign(*v)
+                    for x in (0..xsize).step_by(D::F32Vec::LEN) {
```
veluca93: probably better to use `.chunks_exact_mut(D::F32Vec::LEN).take(xsize.div_ceil(D::F32Vec::LEN))` (similar to Gaborish)

jxl/src/util/rational_poly.rs
```diff
@@ -13,3 +15,25 @@ pub fn eval_rational_poly<const P: usize, const Q: usize>(x: f32, p: [f32; P], q
     let yq = q.into_iter().rev().reduce(|yq, q| yq * x + q).unwrap();
     yp / yq
 }
+
+#[inline(always)]
+pub fn eval_rational_poly_simd<D: SimdDescriptor, const P: usize, const Q: usize>(
```
veluca93, line 20: (making this a loop might work better)

sboukortt: I see your point. :D

Issue #422, opened 2025-10-31
range end index X out of range for slice of length Y


monad0:
Tried to decode this file to PNG with jxl_cli:
![Image](https://github.com/user-attachments/assets/57c352ad-6302-4c75-98d8-4f7102806e97)

More cases: [range end index 36 out of range for slice of length 32.zip](https://github.com/user-attachments/files/23262271/range.end.index.36.out.of.range.for.slice.of.length.32.zip)

Example output:
```
thread 'main' (redacted) panicked at jxl/src/render/stages/extend.rs:82:42:
range end index 2884 out of range for slice of length 2048
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/redacted/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/redacted/library/core/src/panicking.rs:80:14
   2: core::slice::index::slice_index_fail::do_panic::runtime
             at /rustc/redacted/library/core/src/panic.rs:173:21
   3: core::slice::index::slice_index_fail
             at /rustc/redacted/library/core/src/panic.rs:178:9
   4: <core::ops::range::Range<usize> as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/redacted/library/core/src/slice/index.rs:438:13
   5: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/redacted/library/core/src/slice/index.rs:18:15
   6: jxl::render::stages::extend::ExtendToImageDimensionsStage::process_row_chunk
             at /redacted/jxl-rs/jxl/src/render/stages/extend.rs:82:42
   7: jxl::render::low_memory_pipeline::render_group::<impl jxl::render::low_memory_pipeline::LowMemoryRenderPipeline>::render_outside_frame
             at /redacted/jxl-rs/jxl/src/render/low_memory_pipeline/render_group.rs:352:24
   8: jxl::render::low_memory_pipeline::LowMemoryRenderPipeline::render_padding
             at /redacted/jxl-rs/jxl/src/render/low_memory_pipeline/mod.rs:205:18
   9: <jxl::render::low_memory_pipeline::LowMemoryRenderPipeline as jxl::render::RenderPipeline>::do_render
             at /redacted/jxl-rs/jxl/src/render/low_memory_pipeline/mod.rs:442:18
  10: jxl::frame::render::<impl jxl::frame::Frame>::render_frame_output
[...]

PR #426, opened 2025-10-31
add jxl art test image

mo271:


mo271:
currently the same error as in #422

PR #440, opened 2025-11-01
Add some #[inline] annotations.

veluca93:
~3% speedup.


jxl/src/image/internal.rs
```diff
@@ -121,6 +122,7 @@ impl RawImageBuffer {
         }
     }
 
+    #[inline]
```
sboukortt, line 125: Why is this one not `(always)`?

veluca93: Mostly because it's not *too* important that it gets inlined.

PR #446, opened 2025-11-03
Added --num-reps to jxl_cli speedtest.

zond:
Extracted the actual decoding in jxl_cli into it's own function, repeatedly calling it num-reps times.

### DIFF
--- a/jxl_cli/src/enc/numpy.rs
+++ b/jxl_cli/src/enc/numpy.rs
@@ -3,7 +3,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-use crate::ImageData;
+use crate::DecodeOutput;
 use jxl::error::{Error, Result};
 use jxl::image::ImageRect;
 use std::io::Write;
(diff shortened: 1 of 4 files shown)


veluca93:
Seems reasonable, but I wouldn't mind if @sboukortt gave it a second look :-)

jxl_cli/src/enc/numpy.rs
```diff
@@ -54,18 +54,16 @@ fn numpy_header<Writer: Write>(
...
+    image_data: &ImageData<f32>,
     num_channels: usize,
     writer: &mut Writer,
 ) -> Result<()> {
-    let size = image_data.size;
-    let (width, height) = size;
+    let (width, height) = image_data.size;
 
-    for frame in image_data.frames {
-        assert_eq!(frame.size, size);
```
sboukortt, line 65: Should we really drop this?

jxl_cli/src/main.rs
```diff
@@ -272,35 +251,73 @@ fn main() -> Result<()> {
...
+            duration_sum += iteration_duration;
+            Ok(Some(iteration_image_data))
+        })?
+        .unwrap();
 
     if opt.speedtest {
-        let duration = start.elapsed().as_nanos() as f64 / 1e9;
+        let num_pixels = image_data.size.0 * image_data.size.1;
+        let nanos = duration_sum.as_nanos() as f64 / reps as f64;
+        let seconds = nanos / 1e9;
```
sboukortt: Maybe naming these `average_nanos` / `average_seconds` (or `mean_`) would make it a bit clearer what is going on.

jxl_cli/src/main.rs
```diff
@@ -5,70 +5,66 @@
...
 }
 
 pub struct ImageData<T: ImageDataType> {
     pub size: (usize, usize),
     pub frames: Vec<ImageFrame<T>>,
+    pub original_bit_depth: JxlBitDepth,
+    pub output_profile: JxlColorProfile,
+    pub jxl_animation: Option<JxlAnimation>,
+    pub original_icc_result: Result<()>,
+    pub data_icc_result: Result<()>,
```
sboukortt, line 40: Having these as part of `ImageData` feels slightly strange, but I’m not sure whether the best approach is:

- rename the struct (for example to `DecodingOutput`);

- move these two outside of the struct;

- not care; it’s internal to `jxl_cli` anyway.

jxl_cli/src/enc/numpy.rs
zond: The size of the frame is always full image size in the API, and was even coded to just be set to image size in jxl_cli.

jxl_cli/src/main.rs
zond: Rename sounds good.

jxl_cli/src/main.rs
zond: Done

PR #450, opened 2025-11-04
Made frame indices seed the rng

### LINKED ISSUE #424: Discrepancy between libjxl concerning decode to APNG
JXL:
Image

libjxl (djxl) decode:
Image

jxl-rs (jxl_cli) decode (original file fails upload, but all frames appear identical):
[hbo.jxl-rs_20251030.zip](https://github.com/user-attachments/files/23248890/hbo.jxl-rs_20251030.zip)

Image source unknown.

zond:
Fixes #424

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -370,7 +370,6 @@ impl Frame {
             let upsampling = self.header.upsampling;
             let x0 = gx * upsampling * group_dim;
             let y0 = gy * upsampling * group_dim;
-            // TODO(sboukortt): actual frame indices for the first two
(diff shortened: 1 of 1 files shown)


jxl/src/frame/decode.rs
```diff
@@ -365,7 +371,12 @@ impl Frame {
             let x0 = gx * upsampling * group_dim;
             let y0 = gy * upsampling * group_dim;
             // TODO(sboukortt): actual frame indices for the first two
```
sboukortt: This change obsoletes the TODO, doesn’t it?

PR #453, opened 2025-11-04
Change IDCT impl and add a Sse4 SIMD descriptor.

veluca93:
Also adds a downgrade() method to SIMD descriptor, switches to fixed shifts of integers as sse4 has no variable shifts, uses block-by-block transposes, removes the usage of scratch space in IDCT, moves the transforms to a new crate, and merges the match {} blocks for LF and DCT.


veluca93:
Review notes: the files idct*.rs in jxl_transform are generated, no need to look at them if not to understand the generator better.

Speed-wise, the end-to-end speed didn't change too much (1-2%), but the DCT themselves got significantly faster (30-70% in most cases) for AVX and AVX512 (except 2x2), with AVX512 now being faster than AVX; scalar got a bit slower, but I don't think it matters too much (and if it did, we can probably improve the situation).

I plan to next do something similar to reinterpreting_dct, and get rid of a few parts of code (such as scales.rs, which can become test-only, and the generic non-inplace SIMD transpose)

PR #459, opened 2025-11-06
Get rid of SIGMA_PADDING.

veluca93:
It is not necessary, and removing it simplifies the logic.

### DIFF
--- a/jxl/src/features/epf.rs
+++ b/jxl/src/features/epf.rs
@@ -4,7 +4,6 @@
 // license that can be found in the LICENSE file.
 
 use crate::{
-    SIGMA_PADDING,
     error::{Error, Result},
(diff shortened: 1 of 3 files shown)


jxl/src/features/epf.rs
```diff
@@ -34,9 +37,9 @@ pub fn create_sigma_image(
         let epf_map = hf_meta.epf_map.as_rect();
         let mut sigma_rect = sigma_image.as_rect_mut();
         for by in 0..size_blocks.1 {
-            let sby = SIGMA_PADDING + by;
+            let sby = by;
```
sboukortt: Should we keep these at all?

veluca93: Removed.

PR #471, opened 2025-11-11
Avoiding br.peek when unnecessary

zond:
According to perf, this made AnsHistogram::read go from 12.62% to 12.14%.


zond:
> IIRC from libjxl this was very CPU-dependent. We might re-evaluate this later (especially if we make peek no longer implicitly refill)

:+1:

PR #476, opened 2025-11-12
Made bit reader refill conditional

zond:
Before: Decoded 9088204800 pixels in 515.156071204 seconds: 17641653.293067962 pixels/s
After:  Decoded 9088204800 pixels in 506.305871868 seconds: 17950028.441245105 pixels/s

~1.7% faster


zond:
> Why not `if self.bits_in_buf < num` instead?
> 
> Ideally, we would have a separate function for cases in which we _know_ a refill is not necessary, but we can do that later.

Done

Issue #445, opened 2025-11-02
Non-section buffer of 4096 bytes is not enough for some huge images


tirr-c:
Image: [windwaker.jxl](https://files.catbox.moe/2kdb4f.jxl) (23164x18480, 14 MiB)

4096 bytes are just not enough to hold all of the TOC entries, it tries to decode TOC and fails to do so infinitely. I was able to decode the image with buffer of 16384 bytes, with a patch to use `Vec<u8>` instead of `[u8; SIZE]` in `SmallBuffer` (without it the compilation takes too much time).

veluca93:
It's worse than that: some headers might be actually multiple megabytes (admittedly, you'd have to construct such an image on purpose -- i.e. by having 1k channels each of which has a 1kB name).

I suspect the best way to fix this is to make header parsing incremental, but perhaps there are other options...

PR #477, opened 2025-11-12
Incremental TOC reader

tirr-c:
Related to #445

Now it actually decodes windwaker.jxl.


tirr-c:
UnconditionalCoder derive is still there mainly because I wasn't sure about whether they can be safely removed. Sounds like a great idea to use those in tests, will work on that.

veluca93 (approved):

PR #483, opened 2025-11-15
Make `SymbolReader` optimistic

tirr-c:
`SymbolReader` keeps read error in its internal state, and reports it when `.check_final_state()` is called.

About 1% to 10% speedups, depending on the image.

### DIFF
--- a/jxl/src/entropy_coding/hybrid_uint.rs
+++ b/jxl/src/entropy_coding/hybrid_uint.rs
@@ -54,22 +54,19 @@ impl HybridUint {
     }
 
     #[inline]
-    pub fn read(&self, symbol: u32, br: &mut BitReader) -> Result<u32, Error> {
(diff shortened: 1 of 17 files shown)


jxl/src/entropy_coding/hybrid_uint.rs
```diff
@@ -71,6 +71,21 @@ impl HybridUint {
...
     }
+
+    #[inline]
+    pub fn read_optimistic(&self, symbol: u32, br: &mut BitReader) -> u32 {
+        if symbol < self.split_token {
+            return symbol;
+        }
+        let bits_in_token = self.lsb_in_token + self.msb_in_token;
+        let nbits =
+            self.split_exponent - bits_in_token + ((symbol - self.split_token) >> bits_in_token);
```
tirr-c, line 63: Didn't have an idea how to report `nbits > 31` error here...

PR #487, opened 2025-11-20
Added criterion bench for decoding

zond:
Defaults to benchmarking the files in the jxl/resourced directory, but can be controlled via JXL_FILES env variable.

Also moved the decode_bytes function in jxl_cli out of main.rs to import it into the benchmark.

### DIFF
--- a/jxl_cli/benches/decode.rs
+++ b/jxl_cli/benches/decode.rs
@@ -1,0 +1,68 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+use criterion::{BenchmarkId, Criterion, SamplingMode, criterion_group, criterion_main};
+use jxl::api::JxlDecoderOptions;
+use jxl_cli::dec::{decode_bytes, decode_header};
+use std::fs;
+use std::path::{Path, PathBuf};
+
+fn decode_benches(c: &mut Criterion) {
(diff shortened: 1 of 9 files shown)


jxl_cli/benches/decode.rs
```diff
@@ -0,0 +1,65 @@
...
+use jxl_cli::dec::decode_bytes;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+fn decode_benches(c: &mut Criterion) {
+    let mut group = c.benchmark_group("decode");
+    group.sampling_mode(SamplingMode::Flat);
+
+    let paths: Vec<PathBuf> = std::env::var("JXL_FILES").map_or(
+        {
```
sboukortt: Maybe `map_or_else` so that this is not evaluated when not needed?

zond: Good point, done

jxl_cli/benches/decode.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#![allow(clippy::identity_op)]
```
veluca93: Why this?

zond: Copypasta from your benches :P Removed it.

camachojos995-lab:
Hola soy José Luis Camacho Riascos jc1323433@gmail.com

PR #489, opened 2025-11-21
Reader instead of slice

zond:
Using a JxlBitstreamInput instead of bytes slice in the decode function, which allows jxl_cli to pass it a BufReader instead of a slice.


zond:
This is on top of https://github.com/libjxl/jxl-rs/pull/488, but it would of course work independently as well.

PR #492, opened 2025-11-23
Remove zerocopy from production dependencies

hjanuschka:
This removes zerocopy from jxl-rs production dependencies, so downstream users don't need `allow_unsafe = true` in their zerocopy configuration.

**Changes:**

1. **Custom f16 implementation** - Replace `half` crate with ~100 lines of pure safe Rust in `jxl/src/util/float16.rs`. The `half` crate pulled in zerocopy.

2. **Fix jxl_transforms Cargo.toml** - Move `rand`/`rand_chacha` to dev-dependencies (they're only used in `#[cfg(test)]` code).

**Result:**
```
cargo tree -p jxl --edges no-dev | grep zerocopy
# (empty - no zerocopy in production deps)
```

The `jxl_cli` still uses `half` for EXR interop, but that's a dev/CLI tool.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,3 +28,4 @@ Jacob Abel <jacobabel@nullpo.dev>
 Zoltan Szabadka
 Caio <c410.f3r@gmail.com>
 Louis Dispa <louis.dispa@outlook.fr>
+Helmut Januschka <helmut@januschka.com>


veluca93:
Please add yourself to the AUTHORS file!

hjanuschka:
@veluca93  awesome, thanks for merging, could you name ETA when a new release happens? would simplify alot on my side, thanks in advance

veluca93:
> @veluca93 awesome, thanks for merging, could you name ETA when a new release happens? would simplify alot on my side, thanks in advance

That's a good question - on one hand, I can see the advantage of releasing the dependency reductions. OTOH, there's a bunch of unsafe code that would need to be reviewed and might change once we're "ready" with optimizations, so I am not sure what's the best option...

hjanuschka:
ok, would it really be bad? not really fully understanding the issue, when all tests pass, anyway, i am going to use the commit id for now, and update once there is a crate.io release

PR #491, opened 2025-11-23
feat: Implement HDR tone mapping for ICC profile generation

hjanuschka:
Implements HDR tone mapping for ICC profile generation (A2B0/B2A0 tags).

Previously, `maybe_create_profile()` returned `Ok(None)` for HDR content. Now it generates proper ICC profiles with tone-mapped 3D LUTs.

**What's added:**
- BT.2408 tone mapper for PQ (10000→250 nits)
- HLG OOTF
- Gamut mapping
- mft1 A2B0 tag (9³ LUT) and mBA B2A0 tag

Based on libjxl's `tone_mapping.h` / `jxl_cms_internal.h`.

**Testing:**
- Unit tests for all components (tone mapper, OOTF, gamut mapping, ICC generation)
- Verified generated profiles have correct Lab PCS, A2B0/B2A0 tags, and CICP tags

PR #493, opened 2025-11-23
Fix rect bounds check for padded images + add alpha boundary test

hjanuschka:
## Summary
- Fixes bounds check in shift_rect that was incorrectly checking against content size instead of allocation size
- The shifted rect extends into padding area which is valid (offset <= padding is guaranteed)
- Adds test that validates alpha channel decoding at group boundaries (x=256, 512)

Split from #491 per reviewer request.

## Test plan
- cargo test --package jxl --test alpha_boundary


jxl/src/image/raw.rs
```diff
@@ -64,7 +64,10 @@ impl OwnedRawImage {
             size: rect.size,
         };
         if cfg!(debug_assertions) {
-            ret.check_within(self.byte_size());
+            // Check against the allocation size (data.byte_size()), not the content size
```
veluca93: Checking against the content size was intentional - do you have cases in which this causes issues or failures?

hjanuschka: https://jpegxl.info/images/dice.jxl from the demo page failed
> Rect out of bounds: 28x5+8+2 rect in 28x5 view

veluca93: Ok, I checked the code again and this check is complete nonsense - instead we should check that the *original* rect is within the byte size without any extra padding (the returned rect will be checked elsewhere anyway).

veluca93 (approved):

jxl/tests/alpha_boundary.rs
veluca93, line 1: Do we need this explicit test? I'd have thought that using the compare_pipelines check on this image would catch all the issues.

hjanuschka: good catch! removing the test

PR #494, opened 2025-11-23
Fix spline DISTANCE_EXP to match libjxl default

### LINKED ISSUE #431: Discrepancy between libjxl concerning spline rendering in XYB
JXL:
Image

djxl decode:
Image

jxl_cli decode:
Image

Source:
[map.txt](https://github.com/user-attachments/files/23263113/map.txt)

hjanuschka:
Use DISTANCE_EXP=3.0 by default instead of 5.0 to match libjxl's non-high-precision mode.

Adds `high_precision` option to `JxlDecoderOptions` for integrators who need the higher precision behavior.

Fixes #431

### DIFF
--- a/jxl/src/api/options.rs
+++ b/jxl/src/api/options.rs
@@ -26,9 +26,12 @@ pub struct JxlDecoderOptions {
     pub xyb_output_linear: bool,
     pub enable_output: bool,
     pub cms: Option<Box<dyn JxlCms>>,
-    /// Use high precision mode for spline rendering.
+    /// Use high precision mode for decoding.
     /// When false (default), uses lower precision settings that match libjxl's default.
     /// When true, uses higher precision at the cost of performance.
+    ///
+    /// This affects multiple decoder decisions including spline rendering precision
+    /// and potentially intermediate buffer storage (e.g., using f32 vs f16).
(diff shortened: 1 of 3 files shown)


jxl/src/api/options.rs
```diff
@@ -26,6 +26,10 @@ pub struct JxlDecoderOptions {
     pub xyb_output_linear: bool,
     pub enable_output: bool,
     pub cms: Option<Box<dyn JxlCms>>,
+    /// Use high precision mode for spline rendering.
+    /// When false (default), uses lower precision settings that match libjxl's default.
+    /// When true, uses higher precision at the cost of performance.
+    pub high_precision: bool,
```
veluca93, line 35: Just a nit, but in libjxl JXL_HIGH_PRECISION is used also for other decisions, not just for spline rendering - we might want to have the documentation reflect that (i.e. I plan to have high_precision control whether some intermediate buffers could be saved as f16)

hjanuschka: documentation updated, should i give it a try to add control as you mentioned (or should this be a separate PR)?

veluca93: No, this is fine for now.

PR #495, opened 2025-11-23
Add preview frame API support

hjanuschka:
- Add `preview_size: Option<(usize, usize)>` to `JxlBasicInfo`
- Implement the `skip_preview` option (was defined but never used)
- Track preview frame state to properly skip when requested


hjanuschka:
~~conformance fails on all of my PRs 🙈 is it broken or my fault?~~ (was on my end; atleast i hope bots succeed now)
what else is needed to get it merged?

zond:
> ~conformance fails on all of my PRs 🙈 is it broken or my fault?~ (was on my end; atleast i hope bots succeed now) what else is needed to get it merged?

You can run the conformance tests locally if you want. Even if it's a pain it may save you time now and then. See https://github.com/libjxl/jxl-rs/blob/main/.github/workflows/pull_request.yml#L149 for how to do it.

PR #496, opened 2025-11-23
Implement remaining decoder API methods

hjanuschka:
## Summary

Implements the remaining decoder API methods that were previously stubs:

- **`num_completed_passes()`**: Returns the minimum number of passes completed across all groups. Useful for progressive decoding progress reporting.

- **`set_pixel_format()`**: Allows overriding the output pixel format (color type, data format, extra channels).

- **`set_output_color_profile()`**: Allows specifying the target color space for output. Validates that CMS is available when using ICC profiles.

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -4,8 +4,8 @@
 // license that can be found in the LICENSE file.
 
 use super::{
-    JxlBasicInfo, JxlBitstreamInput, JxlColorProfile, JxlDecoderInner, JxlDecoderOptions,
-    JxlOutputBuffer, JxlPixelFormat, ProcessingResult,
+    JxlBasicInfo, JxlBitstreamInput, JxlColorProfile, JxlDataFormat, JxlDecoderInner,
+    JxlDecoderOptions, JxlOutputBuffer, JxlPixelFormat, ProcessingResult,
 };
 #[cfg(test)]
 use crate::frame::Frame;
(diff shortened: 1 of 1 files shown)


hjanuschka:
todos removed and implemented

jxl/src/render/simple_pipeline/save.rs
```diff
@@ -49,11 +49,15 @@ impl SaveStage {
                     }
 
                     match self.data_format {
-                        JxlDataFormat::U8 { .. } => {
-                            write_pixel!(px as u8, Endianness::LittleEndian);
+                        JxlDataFormat::U8 { bit_depth } => {
+                            let max = ((1u32 << bit_depth) - 1) as f64;
```
veluca93: I don't believe this change is correct - bit depth is already handled by conversions, no?

jxl/src/render/builder.rs
```diff
@@ -132,13 +133,58 @@ impl<Pipeline: RenderPipeline> RenderPipelineBuilder<Pipeline> {
 
     #[instrument(skip_all, err)]
     pub fn add_save_stage(
```
veluca93, line 134: I think adding an implicit conversion here is surprising (if subsequent stages want to use that data). Instead, I think we should add conversion stages as needed in frame/render.rs.

jxl/src/render/builder.rs
```diff
@@ -132,13 +133,58 @@ impl<Pipeline: RenderPipeline> RenderPipelineBuilder<Pipeline> {
...
+                        channel_types[i].is_none_or(|ty| ty == crate::image::DataTypeTag::F32);
+
+                    if needs_conversion {
+                        self = self
+                            .add_inout_stage(ConvertF32ToU16Stage::new(channel, bit_depth))?;
+                    }
+                }
+            }
+            // F16 and F32 don't need conversion - the pipeline already uses f32
+            JxlDataFormat::F16 { .. } | JxlDataFormat::F32 { .. } => {}
```
veluca93: F16 *does* need conversion, no?

jxl/src/api/decoder.rs
```diff
@@ -253,8 +253,6 @@ pub(crate) mod tests {
         assert!(buffer_height > 0);
 
         // Get pixel format info
-        // TODO(veluca): this relies on the default pixel format using floats. We should not do
-        // this, and instead call set_pixel_format, but that is currently not implemented.
         let pixel_format = decoder_with_image_info.current_pixel_format().clone();
```
veluca93, line 269: The comment is still partially true - now that set_pixel_format is implemented, can we call that instead?

veluca93 (approved):

veluca93 (approved):

PR #497, opened 2025-11-23
implement flush_pixels for progressive decoding

hjanuschka:
Adds `flush_pixels()` to output partially decoded pixels during progressive decoding.

- Added `flush` method to RenderPipeline trait
- Implemented for both SimpleRenderPipeline and LowMemoryRenderPipeline
- Added `flush_to_buffers` on Frame
- Wired through CodestreamParser to JxlDecoderInner
- 6 tests covering edge cases, both pipelines, modular/vardct, idempotency

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -385,4 +385,442 @@ pub(crate) mod tests {
     }
 
     for_each_test_file!(compare_pipelines);
+
+    /// Test that flush_pixels works when no frame is currently being decoded.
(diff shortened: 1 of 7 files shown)


jonnyawsom3:
I was going to say, isn't that functionality similar to the Eager progressive mode?
https://github.com/libjxl/jxl-rs/blob/9b5a9141f51fd96cfc841a5a5573957e766e5b93/jxl/src/api/options.rs#L8-L15

PR #501, opened 2025-11-26
bt709 based on sRGB

zond:
Refactored the bt709 conversion based on the sRGB conversion. SIMD path is ~40% faster according to perf.

### DIFF
--- a/jxl/src/color/tf.rs
+++ b/jxl/src/color/tf.rs
@@ -6,71 +6,8 @@
 use crate::util::{eval_rational_poly, eval_rational_poly_simd};
 use jxl_simd::{F32SimdVec, SimdDescriptor, SimdMask};
 
-const SRGB_POWTABLE_UPPER: [u8; 16] = [
-    0x00, 0x0a, 0x19, 0x26, 0x32, 0x41, 0x4d, 0x5c, 0x68, 0x75, 0x83, 0x8f, 0xa0, 0xaa, 0xb9, 0xc6,
-];
-
-const SRGB_POWTABLE_LOWER: [u8; 16] = [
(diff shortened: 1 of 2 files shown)


jxl/src/color/tf.rs
```diff
@@ -171,41 +179,106 @@ pub fn srgb_to_linear_simd<D: SimdDescriptor>(d: D, samples: &mut [f32]) {
 }
 
 /// Converts the linear samples with the BT.709 transfer curve.
+// Rational polynomial approximation of 1.099 * x^0.45 - 0.099 on sqrt(x).
+// Max error ~3e-7
 pub fn linear_to_bt709(samples: &mut [f32]) {
```
veluca93: I believe this is only used in tests now, no? Can we move it to the test module and use the "naive" implementation?

jxl/src/color/tf.rs
```diff
@@ -171,41 +179,106 @@ pub fn srgb_to_linear_simd<D: SimdDescriptor>(d: D, samples: &mut [f32]) {
...
+            .if_then_else_f32(
+                a * D::F32Vec::splat(d, 4.5),
+                eval_rational_poly_simd(d, a.sqrt(), P, Q),
+            )
             .copysign(x)
             .store(vec);
     }
 }
 
+/// Converts the linear samples with the BT.709 transfer curve.
```
veluca93: Let's just use the SIMD code also for scalar.

jxl/src/color/tf.rs
zond: Done

jxl/src/color/tf.rs
zond: Done

PR #503, opened 2025-11-26
WIP: Add XYB planar output support

hjanuschka:
## Summary
Adds XYB planar output support - outputs raw XYB color space data in separate buffers instead of converting to interleaved RGB.

## Why
Eliminates ~5% CPU overhead from XYB→RGB conversion by letting Skia do color conversion on GPU, similar to how JPEG/AVIF use YUV planar output.

## Changes
- New color types: `JxlColorType::Xyb` and `JxlColorType::XybAlpha`
- Skip XYB→RGB conversion stage when XYB output requested
- Output to 3 separate buffers (X, Y, B) instead of 1 interleaved RGB buffer
- Update buffer count validation for XYB planar (expects 3 buffers not 1)
- Block XYB planar for animations (blending not supported)

## Test
`cargo test test_decode_xyb_planar` - decodes real JXL file to XYB planar format

### DIFF
--- a/jxl/src/api/data_types.rs
+++ b/jxl/src/api/data_types.rs
@@ -13,6 +13,8 @@ pub enum JxlColorType {
     Rgba,
     Bgr,
     Bgra,
+    Xyb,
+    XybAlpha,
 }
 
 impl JxlColorType {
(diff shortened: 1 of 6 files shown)


veluca93:
I don't think we should be doing this any time soon - it adds a lot of complexity, and there are more significant performance improvements that we should do first.

In particular, I don't think we should do this before we reach performance parity with libjxl, and even not before we add support for having relevant parts of the computation happen in float16 instead of f32 on platforms where it's beneficial.

hjanuschka:
@veluca93 thanks! can we keep it as WIP?

veluca93:
Sure, no problem, but I expect the likelihood of it getting out of date to be high :-)

PR #504, opened 2025-11-26
Add ISO 21496-1 gain map support

hjanuschka:
## What this adds

The decoder can now read gain maps from JXL files. Gain maps let images store both SDR and HDR versions with conversion metadata - pretty useful for cross-device HDR support.

## Implementation

- Added `GainMapBundle` struct matching libjxl's API
- Parses `jhgm` boxes from containers  
- Binary format is byte-for-byte compatible with libjxl
- Includes example showing how to use it
- Test file generator + integration tests

## Testing

All tests pass - verified against libjxl's test vectors to make sure the binary format matches exactly.

```bash
cargo test --package jxl --lib api::gain_map     # unit tests
cargo test --package jxl --test gain_map_parsing # integration tests
cargo run --example gain_map_info -- file.jxl   # try it out
```

The only thing not implemented yet is ColorEncoding serialization (would need BitWriter), but that's rarely used since ICC profiles work fine.

Let me know if you'd like any changes!


jxl/src/api/gain_map.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// The format is defined by the ISO 21496-1 standard.
+    pub gain_map_metadata: Vec<u8>,
+
+    /// Optional color encoding for the gain map.
+    ///
+    /// If present, this describes the color space of the gain map image.
+    /// If None, the gain map uses the same color encoding as the base image.
+    pub color_encoding: Option<ColorEncoding>,
+
+    /// Alternative ICC profile for the gain map (may be Brotli-compressed).
```
jonsneyers: No, ICC profiles cannot be Brotli-compressed, they use the specific ICC compression also used in the jxl image header.

jxl/src/api/gain_map.rs
```diff
@@ -0,0 +1,310 @@
...
+            // For now, write 0 to indicate no color encoding
+            output.push(0);
+        } else {
+            output.push(0);
+        }
+
+        // Write alt_icc_size and alt_icc
+        let icc_size = self.alt_icc.len() as u32;
+        output.extend_from_slice(&icc_size.to_be_bytes());
+        output.extend_from_slice(&self.alt_icc);
```
jonsneyers, line 133: The ICC profile is not just stored as a blob, it uses the same syntax as in the jxl image header.
The code as it is here is fine, but with the understanding that `alt_icc` has to be an already-compressed representation of the ICC profile. The `alt_icc_size` field does correspond to the compressed length so that's fine.

jxl/src/api/gain_map.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// If None, the gain map uses the same color encoding as the base image.
+    pub color_encoding: Option<ColorEncoding>,
+
+    /// Alternative ICC profile for the gain map (may be Brotli-compressed).
+    ///
+    /// This is used when the color encoding cannot be fully described by the
+    /// JPEG XL ColorEncoding structure.
+    pub alt_icc: Vec<u8>,
+
+    /// The gain map image data as a JPEG XL naked codestream.
```
jonsneyers: Doesn't have to be naked. It can be an arbitrary embedded JPEG XL file, also one using the container (but it is not allowed to itself contain a gain map box). Originally it was required to be a naked codestream, but allowing embedded containers makes it possible to have a gain map with `jbrd`, in case the gain map is itself a recompressed JPEG and you want to be able to reconstruct that JPEG bitstream exactly.

tools/create_gain_map_test.py
```diff
@@ -0,0 +1,186 @@
...
+    Create a minimal valid JPEG XL naked codestream for a 1x1 pixel image.
+    This is a very simplified version just for testing purposes.
+    """
+    # JXL signature for naked codestream
+    sig = bytes([0xff, 0x0a])
+
+    # Extremely minimal bitstream for 1x1 image
+    # This is a simplified representation - a real encoder would be more complex
+    # For testing, we'll use a known-good minimal codestream
+    minimal_stream = sig + bytes([
```
jonsneyers: I don't think this is a conforming bitstream, is it?

If you want a small valid bitstream, the smallest I could come up with is here (12 bytes):
https://github.com/mathiasbynens/small/pull/125

jxl/src/api/gain_map.rs
```diff
@@ -0,0 +1,310 @@
...
+        let color_encoding_size = data[offset] as usize;
+        offset += 1;
+
+        // Read color_encoding if present
+        let color_encoding = if color_encoding_size > 0 {
+            if data.len() < offset + color_encoding_size {
+                return Err(Error::OutOfBounds(color_encoding_size));
+            }
+
+            // TODO: Phase 4 - parse ColorEncoding from bitstream
```
veluca93: why can't we do that already?

jxl/src/api/inner/box_parser.rs
```diff
@@ -152,6 +185,10 @@ impl BoxParser {
                             };
                             self.state = ParseState::CodestreamBox(content_len);
                         }
+                        b"jhgm" => {
+                            // Gain map box - collect the data
+                            self.state = ParseState::GainMapBox(content_len);
```
tirr-c, line 198: I think `try_reserve(content_len)?` on `gain_map_data` would be good here.

jxl/src/api/inner/box_parser.rs
```diff
@@ -87,6 +93,33 @@ impl BoxParser {
...
+                    // All data collected, parse the gain map
+                    match GainMapBundle::from_bytes(&self.gain_map_data) {
+                        Ok(bundle) => {
+                            self.gain_map = Some(bundle);
+                        }
+                        Err(_) => {
+                            // Invalid gain map data, ignore it
+                        }
+                    }
+                    self.gain_map_data.clear(); // Free memory
```
tirr-c: `clear()` doesn't free underlying buffer, it just sets `len` to zero. We need `std::mem::take` to actually free memory.

jxl/src/api/inner/box_parser.rs
```diff
@@ -87,6 +93,33 @@ impl BoxParser {
...
+                        return Err(Error::OutOfBounds(remaining as usize));
+                    }
+
+                    // All data collected, parse the gain map
+                    match GainMapBundle::from_bytes(&self.gain_map_data) {
+                        Ok(bundle) => {
+                            self.gain_map = Some(bundle);
+                        }
+                        Err(_) => {
+                            // Invalid gain map data, ignore it
```
tirr-c: It would be nice to leave a warning log here.

jxl/src/api/gain_map.rs
hjanuschka: updated comment

tools/create_gain_map_test.py
```diff
@@ -0,0 +1,170 @@
...
+    box_size = len(content) + 8  # 4 bytes size + 4 bytes type + content
+    return struct.pack('>I', box_size) + box_type + content
+
+def create_minimal_jxl_codestream():
+    """
+    Create a minimal valid JPEG XL naked codestream.
+    This is the smallest valid JXL codestream (12 bytes), from:
+    https://github.com/mathiasbynens/small/pull/125
+
+    It encodes a 1x1 grayscale image.
```
Galaxy4594: Small correction, the smallest JXL bitstream is actually a 512x256 RGB image.

zone117x:
@hjanuschka @veluca93 is there anything needed to before this PR can be merged? Would love to have this feature available.

gregbenz:
Hi Helmut,

Thanks for putting #504 together. I saw that #926 just landed, and I’m interested in helping carry the gain-map work forward for use in Chromium.

I tested #504 against Adobe’s official set of 20 JXL gain-map samples, covering both SDR and HDR base images. libjxl finds the `jhgm` box in every file, but the `gain_map_info` example currently reports no gain map for all 20. Their embedded gain maps range from roughly 340 KB to 1.8 MB, so it looks like #926 may give us a cleaner way to handle these larger boxes.

My thought is to build the decode side on top of #926: make auxiliary-box capture bounded and safe for fragmented and trailing input, then add a borrowed `jhgm` parser with alternate-ICC support and solid tests. I’d focus the jxl-rs work on decoding first, then handle the Chromium integration as a separate follow-up. Writing support could come later if there’s interest.

I can validate locally against the full Adobe reference set, while using small redistributable fixtures for upstream tests.

Are you planning to continue work on #504 now that #926 has landed? If so, I’d be happy to help. Otherwise, would you be comfortable with me preparing a focused successor that builds on and credits #504, and asking you to review it? Also, please let me know if there’s any upcoming jxl-rs 0.8 or Chromium API work we should align with.

Thanks!

hjanuschka:
@gregbenz  sounds like you are in a good position to run a successor! happy to review.

once we have something ready, we will see how this will fit into chromium

gregbenz:
Perfect, thanks @hjanuschka! I appreciate the quick feedback. I’ll prepare a focused decode-only successor on top of #926 and tag you when it’s ready for review.

I’ll include redistributable test fixtures and validate it locally. We can look at the Chromium integration once the upstream shape is settled.

gregbenz:
@hjanuschka, this is ready for review now. Thanks again for offering to take a look. I’d appreciate your thoughts on the scope and API.

PR #510, opened 2025-12-01
Made render pipeline use SmallVec

zond:
~1.5% improvement according to the benchmark.


zond:
Nah, I don't trust the benchmarks, I'm not sure this is actually faster.

PR #511, opened 2025-12-01
Fewer allocations

zond:
Created scratch memory for group decoding.

In total seems to be ~1-2% faster on bike.jxl.

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -10,7 +10,7 @@ use super::{
     block_context_map::BlockContextMap,
     coeff_order::decode_coeff_orders,
     color_correlation_map::ColorCorrelationParams,
-    group::decode_vardct_group,
+    group::{VarDctBuffers, decode_vardct_group},
     modular::{FullModularImage, ModularStreamId, Tree, decode_hf_metadata, decode_vardct_lf},
     quant_weights::DequantMatrices,
     quantizer::{LfQuantFactors, QuantizerParams},
@@ -138,6 +138,7 @@ impl Frame {
             reference_frame_data,
             lf_frame_data,
             lf_global_was_rendered: false,
+            vardct_buffers: None,
         })
     }
     /// Given a bit reader pointing at the end of the TOC, returns a vector of `BitReader`s, each
@@ -434,6 +435,7 @@ impl Frame {
                 pipeline!(self, p, p.get_buffer(1))?,
                 pipeline!(self, p, p.get_buffer(2))?,
             ];
+            let buffers = self.vardct_buffers.get_or_insert_with(VarDctBuffers::new);
             decode_vardct_group(
                 group,
                 pass,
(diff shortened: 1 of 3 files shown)


veluca93:
I am not convinced we should be using smallvec, for a few reasons:
- https://github.com/google/rust-crate-audits/blob/3f4f841688a9ee9637863299573f536862a7804f/manual-sources/google3-audits.toml#L2438 declares it to be ub-risk-3, which I would prefer to avoid.
- The smallvec maintainers are working towards smallvec 2.0, which has a very different API
- I don't think we need the full flexibility that SmallVec offers
- I am not entirely convinced that collect-ing into a smallvec is what we'd want to do here, since most of our loops come from filtering a range or start with a pretty strong expectation of the number of number of elements we will have.

Thus, for now I would rather only merge the scratch memory for group decoding part.

zond:
> I am not convinced we should be using smallvec, for a few reasons:
> 
> * https://github.com/google/rust-crate-audits/blob/3f4f841688a9ee9637863299573f536862a7804f/manual-sources/google3-audits.toml#L2438 declares it to be ub-risk-3, which I would prefer to avoid.
> * The smallvec maintainers are working towards smallvec 2.0, which has a very different API
> * I don't think we need the full flexibility that SmallVec offers
> * I am not entirely convinced that collect-ing into a smallvec is what we'd want to do here, since most of our loops come from filtering a range or start with a pretty strong expectation of the number of number of elements we will have.
> 
> Thus, for now I would rather only merge the scratch memory for group decoding part.

IIUC we still depend on smallvec, via tracing-subscriber (but that is perhaps not a reason to add it as a direct dep).

Could I add a local tiny smallvec implementation as replacement untill smallvec 2 is released?

How would you prefer the loops with strong expectations do this instead?

veluca93:
> > I am not convinced we should be using smallvec, for a few reasons:
> > 
> > * https://github.com/google/rust-crate-audits/blob/3f4f841688a9ee9637863299573f536862a7804f/manual-sources/google3-audits.toml#L2438 declares it to be ub-risk-3, which I would prefer to avoid.
> > * The smallvec maintainers are working towards smallvec 2.0, which has a very different API
> > * I don't think we need the full flexibility that SmallVec offers
> > * I am not entirely convinced that collect-ing into a smallvec is what we'd want to do here, since most of our loops come from filtering a range or start with a pretty strong expectation of the number of number of elements we will have.
> > 
> > Thus, for now I would rather only merge the scratch memory for group decoding part.
> 
> IIUC we still depend on smallvec, via tracing-subscriber (but that is perhaps not a reason to add it as a direct dep).

tracing-subscriber is not a dependency of the main library.

> 
> Could I add a local tiny smallvec implementation as replacement untill smallvec 2 is released?
> 
> How would you prefer the loops with strong expectations do this instead?

I will think about this more deeply in the future.

zond:
Updated to remove the smallvecs.

PR #515, opened 2025-12-02
Made pipeline stages use flat buffers

zond:
Made the stages use flat buffers for row data. ~3% faster on bike.jxl.

### DIFF
--- a/jxl/src/render/channels.rs
+++ b/jxl/src/render/channels.rs
@@ -1,0 +1,146 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/// Multi-row channel accessor for immutable access.
+///
+/// Provides 2D indexing where `channels[ch]` returns `&[&[T]]` (all rows for a channel),
+/// and `channels[ch][row]` returns `&[T]` (pixels for a specific row).
+///
+/// This eliminates nested Vec collections while maintaining the same indexing syntax.
+pub struct Channels<'a, T> {
+    pub(crate) row_data: Vec<&'a [T]>,
+    num_channels: usize,
+    pub(crate) rows_per_channel: usize,
+}
+
+impl<'a, T> Channels<'a, T> {
+    /// Create a new Channels accessor.
+    ///
+    /// # Arguments
+    /// * `row_data` - Flat vector of all rows for all channels (length = num_channels * rows_per_channel)
+    /// * `num_channels` - Number of channels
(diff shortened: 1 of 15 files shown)


jxl/src/render/stages/epf/epf2.rs
```diff
@@ -81,17 +78,17 @@ fn epf2_process_row_chunk(
         let sad_mul = D::F32Vec::load(d, &sad_mul_storage[x % 8..]);
 
         if D::F32Vec::splat(d, MIN_SIGMA).gt(sigma).all() {
-            D::F32Vec::load(d, &input_x[1][1 + x..]).store(&mut output_x[0][x..]);
-            D::F32Vec::load(d, &input_y[1][1 + x..]).store(&mut output_y[0][x..]);
-            D::F32Vec::load(d, &input_b[1][1 + x..]).store(&mut output_b[0][x..]);
+            D::F32Vec::load(d, &input_rows[1][1][1 + x..]).store(&mut output_rows[1][0][x..]);
```
veluca93: I believe this is wrong, fwiw.

jxl/src/render/mod.rs
```diff
@@ -81,9 +82,9 @@ pub trait RenderPipelineInOutStage: Any + std::fmt::Display {
         position: (usize, usize),
         xsize: usize,
         // channel, row, column
-        input_rows: &[&[&[Self::InputT]]],
+        input_rows: &Channels<Self::InputT>,
```
veluca93, line 88: Since this is used not just in the low memory pipeline, its implementation should be outside the lm pipeline.

jxl/src/render/stages/epf/epf2.rs
zond: Fixed

jxl/src/render/mod.rs
zond: Done

jxl/src/render/channels.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /// Returns true if there are no channels.
+    pub fn is_empty(&self) -> bool {
+        self.num_channels == 0
+    }
+
+    /// Splits the first `n` channels into separate mutable slices.
+    /// Returns a Vec containing mutable references to each channel's rows.
+    pub fn split_first_n_mut(&mut self, n: usize) -> Vec<&mut [&'a mut [T]]> {
```
veluca93: Let's not return a vec here. Since this is only ever used with n = 3, let's specialize it for that case.

zond: Done, returning tuple.

PR #520, opened 2025-12-03
Add CXX FFI wrapper for Chromium integration

hjanuschka:
## Summary
- Add optional CXX-based FFI wrapper behind the `cxx` feature flag
- Provides C++-compatible API for integration with Chromium's Blink image decoder


hjanuschka:
goes to the stash of ideas that dont work 
/close

Issue #521, opened 2025-12-03
Speed improvements from #506


veluca93:
This issue is a checklist of which speed improvements from https://github.com/libjxl/jxl-rs/pull/506 we want to merge roughly as-is (the rest should probably be left to subsequent investigations)

- [ ] SIMDfication of remaining transfer curves
- [ ] Precomputed cosines in spline rendering
- [ ] Weighted predictor speedup
- [ ] Flattened modular trees
- [ ] Caching of natural coefficient orders
- [ ] Caching of default quantization tables
- [ ] SIMD Chroma upsampling
- [ ] SIMD F32 to U8 conversion
- [ ] SIMD other data type conversions (not in PR)
- [ ] SIMD noise application/convolution
- [ ] SIMD upsampling
- [ ] SIMD YCbCr conversion

PR #506, opened 2025-11-27
road to 1x or less

hjanuschka:
# Road to 1x or Less 🚀

### Benchmark Results (2025-12-01)

| Test Case | Rust (ms) | C++ (ms) | Ratio | Status |
|-----------|-----------|----------|-------|--------|
| bench_oriented_brg | 12.14 | 13.92 | **0.87x** | 🚀 |
| bench_oriented_brg_5 | 12.84 | 13.48 | **0.95x** | 🚀 |
| progressive_5 | 540.97 | 433.79 | 1.25x | 🟢 |
| lossless_pfm | 57.48 | 44.94 | 1.28x | 🟢 |
| progressive | 556.01 | 427.77 | 1.30x | 🟢 |
| bike_5 | 223.37 | 163.83 | 1.36x | 🟢 |
| grayscale_public_university | 356.70 | 258.81 | 1.38x | 🟢 |
| bike | 226.15 | 163.10 | 1.39x | 🟢 |
| lz77_flower | 73.53 | 52.50 | 1.40x | 🟢 |
| delta_palette | 62.60 | 40.48 | 1.55x | 🟡 |
| alpha_nonpremultiplied | 88.90 | 56.43 | 1.58x | 🟡 |
| alpha_premultiplied | 72.11 | 45.49 | 1.59x | 🟡 |
| opsin_inverse | 14.83 | 9.10 | 1.63x | 🟡 |
| grayscale_jpeg_5 | 2.25 | 1.24 | 1.81x | 🟡 |
| cafe_5 | 53.05 | 28.51 | 1.86x | 🟡 |
| cafe | 53.85 | 28.82 | 1.87x | 🟡 |
| opsin_inverse_5 | 16.70 | 8.93 | 1.87x | 🟡 |
| grayscale_5 | 2.82 | 1.50 | 1.88x | 🟡 |
| grayscale | 2.78 | 1.45 | 1.92x | 🟡 |
| alpha_triangles | 100.83 | 52.46 | 1.92x | 🟡 |
| bicycles | 86.33 | 43.67 | 1.98x | 🟡 |
| grayscale_jpeg | 2.34 | 1.18 | 1.98x | 🟡 |
| noise_5 | 23.60 | 11.54 | 2.05x | 🔴 |
| animation_icos4d_5 | 85.04 | 40.73 | 2.09x | 🔴 |
| animation_icos4d | 85.70 | 40.72 | 2.10x | 🔴 |
[...]

PR #525, opened 2025-12-04
Cache natural coefficient orders using OnceLock

hjanuschka:
## Summary
Use `[OnceLock<Vec<u32>>; NUM_ORDERS]` to lazily cache coefficient orders per transform type. This avoids computing orders for large transforms (128x128, 256x256) if they are never used.

## Changes
- Use array of `OnceLock` instead of single `OnceLock` containing array
- Each transform order is computed lazily on first access
- Simplify `get_natural_coeff_order` to take index directly

## Benchmark Results
Using `--speedtest --num-reps 100` on small VarDCT images:

| Test Case | Main (MP/s) | PR (MP/s) | Change |
|-----------|-------------|-----------|--------|
| grayscale | 22.27 | 29.30 | **+32%** |
| opsin_inverse | 25.46 | 27.21 | **+7%** |
| alpha_premultiplied | 18.93 | 19.75 | +4% |

The benefit is most visible on small VarDCT images where coefficient order computation is a larger fraction of total decode time.

Part of #521 (ported from #506)


jxl/src/frame/coeff_order.rs
```diff
@@ -37,6 +38,27 @@ pub const TRANSFORM_TYPE_LUT: [HfTransformType; NUM_ORDERS] = [
 
 pub const NUM_PERMUTATION_CONTEXTS: usize = 8;
 
+/// Cached natural coefficient orders for all transform types.
+/// Computed once on first access to avoid recomputation.
+static NATURAL_COEFF_ORDERS: OnceLock<[Vec<u32>; NUM_ORDERS]> = OnceLock::new();
```
veluca93: Have you tried doing a `[OnceLock<Vec<u32>>; NUM_ORDERS]` instead? that would avoid ever computing coefficient orders for large transforms that are not used.

veluca93:
Can you mentioned in the PR description the speedup and the image(s) the speedup was obtained in?

veluca93:
> Can you mentioned in the PR description the speedup and the image(s) the speedup was obtained in?

For this PR, I would expect the benefits to be mostly visible on small VarDCT images. Can you try one of those out?

hjanuschka:
done thanks, would be nice if you update the master issue and add the images you'd expect most change

zond:
```
MULTI-FILE BENCHMARK RESULTS (4 files, 2 revisions)
  CPU architecture: x86_64

[ 1] c31fec57 Cache natural coefficient orders using per-transform Once...
     alpha_premultiplied.jxl   |                    ├────│─────┤          | 1.030 / prev ▲ faster (1.03 / prev)
     grayscale.jxl             |                    ┃          ├│┤        | 1.071 / prev ▲ faster (1.07 / prev)
     green_queen_vardct_e3.jxl |                  ├─│┤                    | 0.999 / prev 
     opsin_inverse.jxl         |               ├─│┤ ┃                     | 0.981 / prev ▼ slower (0.98 / prev)

```

jxl/src/frame/coeff_order.rs
```diff
@@ -94,10 +120,10 @@ pub fn natural_coeff_order(transform: HfTransformType) -> Vec<u32> {
 }
 
 pub fn decode_coeff_orders(used_orders: u32, br: &mut BitReader) -> Result<Vec<Permutation>> {
```
veluca93, line 123: What do you think of making `Permutation` a `Cow<'static, [u32]>`? This would avoid copying a potentially fairly large `Vec`, and should be a minimal change code-wise.

veluca93 (approved):

PR #527, opened 2025-12-04
Add simple fuzzers via cargo-fuzz (decode, decode_header)

Mrmaxmeier:
Hi,

this adds a basic fuzzing setup based on [`cargo-fuzz`](https://rust-fuzz.github.io/book/cargo-fuzz.html).
To run the fuzzer, install Rust nightly and `cargo-fuzz`, and run the fuzzers with `cargo fuzz run decode` in `jxl/`. (NB: I used `cargo fuzz run decode -D -snone -- -timeout=5` for a quicker build)

This setup is primitive (-> not aware of the JPEG-XL file format) but still finds a bunch of issues quickly:

- `unreachable!()` on `JxlColorEncoding:::XYB` in `non_section.rs`
- assert in `huffman.rs`: `al_size < (1 << HUFFMAN_MAX_BITS)`
- debug assert: `bit_reader.rs` `num <= MAX_BITS_PER_CALL`
- debug panic on integer overflow in `huffman.rs:466`: `decode_varint16(br)? + 1`
- memory limit after allocating an unchecked size (u32 `num_toc_entries`)
- timeout in infinite / non-progressing loop
    
Note: The fuzzer is not effective if the target is slow or crashy, so I applied workarounds for the panics to get to the list of issues above.

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -524,4 +524,20 @@ pub(crate) mod tests {
         let result = decoder.set_output_color_profile(icc_profile);
         assert!(result.is_err());
     }
+
+    #[test]
+    #[should_panic = "al_size < 1 << HUFFMAN_MAX_BITS"]
+    // FIXME: This is a fuzz-bug from #527 and should be fixed.
+    fn fixme_invalid_huffman_table() {
+        let data = [255, 10, 75, 10, 75, 161, 254, 255, 255];
+        let _ = decode(&data, usize::MAX, false, None);
+    }
+
+    #[test]
+    #[should_panic = "internal error: entered unreachable code"]
+    // FIXME: This is a fuzz-bug from #527 and should be fixed.
+    fn fixme_xyb_color_encoding() {
+        let data = [255, 10, 98, 0, 0, 10, 37, 141, 0, 255];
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ and [PATENTS](PATENTS) files. Before we can accept your contributions, even for
 small changes, there are just a few small guidelines you need to follow:
 
 Please fill out either the individual or corporate Contributor License Agreement
-(CLA) with Google. JPEG XL Project is an an effort by multiple individuals and
+(CLA) with Google. JPEG XL Project is an effort by multiple individuals and
```
veluca93, line 12: This seems unrelated to the fuzzer, do you mind reverting the change and sending it as a separate PR?

veluca93 (approved):
Fuzzer looks good to me!

A couple of questions:
- Could you open issues for the issues you found? ideally, with a .jxl file that reproduces the problem independently of the fuzzer PR.
- Is there any chance you might manage to set up miri tests too? :-D

Mrmaxmeier:
I've added `#[should_panic]` tests for the first few issues and have opened an issue for the hanging input. After these are fixed, the fuzzer should be able to test the program more effectively.

PR #531, opened 2025-12-05
Add SIMD F32 to U8 conversion

hjanuschka:
## Summary
- Use the `simd_function!` macro for SIMD-accelerated F32 to U8 conversion
- Clamping to [0, max] range using SIMD max operations
- Upper clamp computed as `max - max(0, max - val)` (equivalent to min)

## Benchmark

| Image | Before (MP/s) | After (MP/s) | Change |
|-------|---------------|--------------|--------|
| cafe | 30.5 | 30.8 | +1% |
| bike | 28.1 | 28.8 | +2% |

Part of #521.

### DIFF
--- a/jxl/src/render/stages/convert.rs
+++ b/jxl/src/render/stages/convert.rs
@@ -259,6 +259,7 @@ simd_function!(
         let simd_width = D::F32Vec::LEN;
 
         let zero = D::F32Vec::splat(d, 0.0);
+        let half = D::F32Vec::splat(d, 0.5);
         let max_vec = D::F32Vec::splat(d, max);
 
         // Temporary buffer for storing SIMD results before converting to u8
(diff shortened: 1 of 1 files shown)


zond:
```
MULTI-FILE BENCHMARK RESULTS (2 files, 2 revisions)
  CPU architecture: x86_64

[ 1] f1c517b2 Add SIMD F32 to U8 conversion
     bike.jxl |                  ├─╂─│────┤              | 1.013 / prev 
     cafe.jxl |                  ├─╂│─────┤              | 1.006 / prev 

```

jxl/src/render/stages/convert.rs
```diff
@@ -250,6 +251,45 @@ impl std::fmt::Display for ConvertF32ToU8Stage {
...
+            let clamped_low = val.max(zero);
+            let scaled = clamped_low * max_vec;
+            // Clamp upper bound using max(0, max - scaled) trick
+            // min(a, b) = b - max(0, b - a)
+            let diff = max_vec - scaled;
+            let clamped = max_vec - diff.max(zero);
+            clamped.store(&mut temp[..simd_width]);
+
+            for i in 0..simd_width {
+                output[x + i] = temp[i].round() as u8;
```
sboukortt: Since the values are all non-negative after the clamping, we could also skip the `.round()` and just add `.5` (in SIMD) instead. Might that be faster?

hjanuschka: done, lets see what jxl-performance will report

veluca93:
Seems to be strictly less functionality than #537, no?

hjanuschka:
clsoing in favor of #537 - it covers u8 and u16 and also has min()

PR #533, opened 2025-12-05
Add min and store_interleaved functions to jxl_simd

hjanuschka:
## Summary

This PR adds several new SIMD functions to prepare for upcoming optimizations (related to #521):

- `min(other)`: Element-wise minimum of two vectors
- `store_interleaved_2(a, b, mem)`: Store two vectors interleaved as `[a0, b0, a1, b1, ...]`
- `store_interleaved_4(a, b, c, d, mem)`: Store four vectors interleaved as `[a0, b0, c0, d0, a1, ...]`
- `store_interleaved_8(a, b, c, d, e, f, g, h, mem)`: Store eight vectors interleaved

All functions are implemented for:
- Scalar
- SSE4.2
- AVX/AVX2
- AVX512
- NEON (aarch64)

### DIFF
--- a/jxl_simd/src/aarch64/neon.rs
+++ b/jxl_simd/src/aarch64/neon.rs
@@ -154,8 +154,6 @@ impl F32SimdVec for F32VecNeon {
         h: Self,
         dest: &mut [f32],
     ) {
-        assert!(dest.len() >= 8 * Self::LEN);
-
         #[target_feature(enable = "neon")]
         #[inline]
         fn store_interleaved_8_impl(
@@ -169,6 +167,7 @@ impl F32SimdVec for F32VecNeon {
             h: float32x4_t,
             dest: &mut [f32],
         ) {
+            assert!(dest.len() >= 8 * F32VecNeon::LEN);
             // NEON doesn't have vst8, so we use manual interleaving
             // For 4-wide vectors, output is 32 elements: [a0,b0,c0,d0,e0,f0,g0,h0, a1,...]
 
@@ -226,7 +225,7 @@ impl F32SimdVec for F32VecNeon {
                 vreinterpretq_f64_f32(cgdh_3),
             ));
 
-            // SAFETY: dest is guaranteed to have enough space by the caller's assert.
+            // SAFETY: we just checked that dest has enough space.
             unsafe {
                 let ptr = dest.as_mut_ptr();
                 vst1q_f32(ptr, out0);
--- a/jxl_simd/src/x86_64/avx.rs
+++ b/jxl_simd/src/x86_64/avx.rs
@@ -199,11 +199,10 @@ impl F32SimdVec for F32VecAvx {
 
     #[inline(always)]
     fn store_interleaved_2(a: Self, b: Self, dest: &mut [f32]) {
-        assert!(dest.len() >= 2 * Self::LEN);
-
         #[target_feature(enable = "avx2")]
         #[inline]
         fn store_interleaved_2_impl(a: __m256, b: __m256, dest: &mut [f32]) {
+            assert!(dest.len() >= 2 * F32VecAvx::LEN);
             // a = [a0, a1, a2, a3, a4, a5, a6, a7], b = [b0, b1, b2, b3, b4, b5, b6, b7]
             // Output: [a0, b0, a1, b1, a2, b2, a3, b3, a4, b4, a5, b5, a6, b6, a7, b7]
             let lo = _mm256_unpacklo_ps(a, b); // [a0, b0, a1, b1, a4, b4, a5, b5]
@@ -211,7 +210,7 @@ impl F32SimdVec for F32VecAvx {
             // Need to permute to get correct order
             let out0 = _mm256_permute2f128_ps::<0x20>(lo, hi); // lower halves: [a0,b0,a1,b1, a2,b2,a3,b3]
             let out1 = _mm256_permute2f128_ps::<0x31>(lo, hi); // upper halves: [a4,b4,a5,b5, a6,b6,a7,b7]
-            // SAFETY: dest is guaranteed to have enough space by the caller's assert.
+            // SAFETY: we just checked that dest has enough space.
             unsafe {
                 _mm256_storeu_ps(dest.as_mut_ptr(), out0);
                 _mm256_storeu_ps(dest.as_mut_ptr().add(8), out1);
@@ -224,11 +223,10 @@ impl F32SimdVec for F32VecAvx {
 
     #[inline(always)]
     fn store_interleaved_4(a: Self, b: Self, c: Self, d: Self, dest: &mut [f32]) {
-        assert!(dest.len() >= 4 * Self::LEN);
-
         #[target_feature(enable = "avx2")]
         #[inline]
         fn store_interleaved_4_impl(a: __m256, b: __m256, c: __m256, d: __m256, dest: &mut [f32]) {
+            assert!(dest.len() >= 4 * F32VecAvx::LEN);
             // First interleave pairs
             let ab_lo = _mm256_unpacklo_ps(a, b);
             let ab_hi = _mm256_unpackhi_ps(a, b);
@@ -259,7 +257,7 @@ impl F32SimdVec for F32VecAvx {
             let out2 = _mm256_permute2f128_ps::<0x31>(abcd_0, abcd_1);
             let out3 = _mm256_permute2f128_ps::<0x31>(abcd_2, abcd_3);
 
-            // SAFETY: dest is guaranteed to have enough space by the caller's assert.
+            // SAFETY: we just checked that dest has enough space.
             unsafe {
                 _mm256_storeu_ps(dest.as_mut_ptr(), out0);
                 _mm256_storeu_ps(dest.as_mut_ptr().add(8), out1);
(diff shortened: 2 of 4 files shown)


jxl_simd/src/aarch64/neon.rs
```diff
@@ -121,6 +121,113 @@ impl F32SimdVec for F32VecNeon {
...
+            // vst4q_f32 stores four vectors interleaved
+            vst4q_f32(
+                base.as_mut_ptr().add(offset),
+                float32x4x4_t(a.0, b.0, c.0, d.0),
+            );
+        }
+    }
+
+    #[inline(always)]
+    fn store_interleaved_8(
```
veluca93, line 146: In general, can we avoid having big unsafe blocks?

In particular, I would prefer having one target_feature function that implements the body of each (more-than-one-line) store-interleave method, and have a single unsafe block for all the `store`s at the end inside it + an unsafe block for calling into the wrapper function.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -115,6 +115,150 @@ impl F32SimdVec for F32VecAvx {
...
+        e: Self,
+        f: Self,
+        g: Self,
+        h: Self,
+        base: &mut [f32],
+        offset: usize,
+    ) {
+        assert!(base.len() >= offset + 8 * Self::LEN);
+        // For 8-wide vectors storing 8 interleaved, output is 64 elements
+        // This is essentially an 8x8 transpose
```
veluca93: Can we share code with the 8x8 transpose implemented somewhere else in this file?

hjanuschka:
## jxl-perfhistory (bike 2048x2560)

```
================================================================================
Performance Graph (with credible intervals):
------------------------------------------------------------------------------------------------------------------------------------------------------
[ 1] 895068e7 Address review comments: reduce unsafe blocks a... |      ├─────────────────────│─────────────────────┤           | 28398939.12 
[ 2] a48cb4a7 Add min and store_interleaved functions to jxl_... |                       ├─────────────────│──────────────────┤ | 29169923.72 ▲ MAX (1.04 / min)
[ 3] 430c9e6c Bump version to 0.1.4. (#541)                      |          ├─────────────────────│──────────────────────┤      | 28641549.76 
[ 4] 86d5e907 Reduce code size with a generic transform impl ... |     ├──────────────│───────────────┤                         | 27952988.15 ▼ MIN (0.96 / max)
[ 5] 46cb3a6c Made pipeline stages use flat buffers (#515)       | ├─────────────────────│─────────────────────┤                | 28097733.57 
------------------------------------------------------------------------------------------------------------------------------------------------------
Scale: 26786267 to 30238673 pixels/s
```

Statistics: Min 27.95 MP/s, Max 29.17 MP/s, Improvement 4.35%

jxl_simd/src/aarch64/neon.rs
hjanuschka: done

jxl_simd/src/x86_64/avx.rs
hjanuschka: done

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,140 @@ impl F32SimdVec for F32VecNeon {
...
+        g: Self,
+        h: Self,
+        base: &mut [f32],
+        offset: usize,
+    ) {
+        assert!(base.len() >= offset + 8 * Self::LEN);
+
+        #[target_feature(enable = "neon")]
+        #[inline]
+        unsafe fn store_interleaved_8_impl(
```
veluca93: We don't need the `unsafe` here, as long as we move the checks on `base` inside the function (and thus pass in a slice and not a pointer).

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,140 @@ impl F32SimdVec for F32VecNeon {
         unsafe { vst1q_f32(mem.as_mut_ptr(), self.0) }
     }
 
+    #[inline(always)]
+    fn store_interleaved_2(a: Self, b: Self, base: &mut [f32], offset: usize) {
+        assert!(base.len() >= offset + 2 * Self::LEN);
+        // SAFETY: we just checked that `base` has enough space.
```
veluca93: These comments should also address `neon` being present.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,140 @@ impl F32SimdVec for F32VecNeon {
         unsafe { vst1q_f32(mem.as_mut_ptr(), self.0) }
     }
 
+    #[inline(always)]
+    fn store_interleaved_2(a: Self, b: Self, base: &mut [f32], offset: usize) {
```
veluca93: I just realized this now, but why are we taking a `offset` argument? IMO it'd be easier to do slice arithmetic in the caller.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,140 @@ impl F32SimdVec for F32VecNeon {
...
+        assert!(base.len() >= offset + 2 * Self::LEN);
```
veluca93: If we keep the offset, we also need to check that the computation does not overflow.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,140 @@ impl F32SimdVec for F32VecNeon {
...
+            let out6 = vreinterpretq_f32_f64(vzip1q_f64(
+                vreinterpretq_f64_f32(aebf_3),
+                vreinterpretq_f64_f32(cgdh_3),
+            ));
+            let out7 = vreinterpretq_f32_f64(vzip2q_f64(
+                vreinterpretq_f64_f32(aebf_3),
+                vreinterpretq_f64_f32(cgdh_3),
+            ));
+
+            vst1q_f32(ptr, out0);
```
veluca93: We should have a safety comment for this block of calls.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,140 @@ impl F32SimdVec for F32VecNeon {
...
+            vst1q_f32(ptr.add(4), out1);
+            vst1q_f32(ptr.add(8), out2);
+            vst1q_f32(ptr.add(12), out3);
+            vst1q_f32(ptr.add(16), out4);
+            vst1q_f32(ptr.add(20), out5);
+            vst1q_f32(ptr.add(24), out6);
+            vst1q_f32(ptr.add(28), out7);
+        }
+
+        // SAFETY: bounds checked above, neon available from safety invariant on descriptor
```
veluca93: This comment should only address NEON, not the bounds. Also, there is no explicit descriptor, so we should name what descriptor is providing that invariant.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -14,6 +14,54 @@ use std::{
     },
 };
 
+/// Core 8x8 transpose algorithm for AVX2.
+/// Takes 8 __m256 vectors representing rows and returns 8 transposed vectors.
+/// Used by both store_interleaved_8 and transpose_square.
+#[target_feature(enable = "avx2")]
+#[inline]
+unsafe fn transpose_8x8_core(
```
veluca93: No unsafe needed.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -116,6 +164,125 @@ impl F32SimdVec for F32VecAvx {
         unsafe { _mm256_storeu_ps(mem.as_mut_ptr(), self.0) }
     }
 
+    #[inline(always)]
+    fn store_interleaved_2(a: Self, b: Self, base: &mut [f32], offset: usize) {
+        assert!(base.len() >= offset + 2 * Self::LEN);
+        // SAFETY: we just checked that `base` has enough space.
+        unsafe {
```
veluca93: Same comment on smaller unsafe blocks from `neon` apply here, too (and also for avx512 and sse4.2)

jxl_simd/src/x86_64/avx.rs
```diff
@@ -116,6 +164,125 @@ impl F32SimdVec for F32VecAvx {
...
+        g: Self,
+        h: Self,
+        base: &mut [f32],
+        offset: usize,
+    ) {
+        assert!(base.len() >= offset + 8 * Self::LEN);
+
+        #[target_feature(enable = "avx2")]
+        #[inline]
+        unsafe fn store_interleaved_8_impl(
```
veluca93: As for Neon, we should make this not unsafe.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -116,6 +197,129 @@ impl F32SimdVec for F32VecAvx {
         unsafe { _mm256_storeu_ps(mem.as_mut_ptr(), self.0) }
     }
 
+    #[inline(always)]
+    fn store_interleaved_2(a: Self, b: Self, dest: &mut [f32]) {
+        assert!(dest.len() >= 2 * Self::LEN);
```
veluca93: We should move these asserts in the _impl functions.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -122,6 +122,128 @@ impl F32SimdVec for F32VecNeon {
...
+        b: Self,
+        c: Self,
+        d: Self,
+        e: Self,
+        f: Self,
+        g: Self,
+        h: Self,
+        dest: &mut [f32],
+    ) {
+        assert!(dest.len() >= 8 * Self::LEN);
```
veluca93: Missed one ;-)

PR #535, opened 2025-12-05
Add SIMD upsampling for 2x, 4x, and 8x

hjanuschka:
## Summary

This implements SIMD-accelerated upsampling using a two-pass algorithm matching the libjxl approach (related to #521):

- **Pass 1**: Precompute min/max values to temp buffers for better cache locality across the 5x5 kernel window
- **Pass 2**: Apply the 5x5 convolution kernel with FMA using 3-way ILP, then clamp results using precomputed min/max values

Key optimizations:
- Pre-broadcast kernel weights to SIMD vectors (avoids repeated splats in inner loop)
- Chunked processing (1024 elements) for L1 cache efficiency
- Use `store_interleaved_2/4/8` for efficient output of interleaved results
- 3-way accumulator pattern for instruction-level parallelism

> **Note**: This PR depends on #533 (min and store_interleaved functions). Please merge #533 first.

## Benchmarks

Tested with `--speedtest --num-reps 10` on conformance test images:

| Image | main (MP/s) | PR (MP/s) | Change |
|-------|-------------|-----------|--------|
| upsampling | 5.5 | 38.9 | **+7.0x** |
| cafe_5 | 24.8 | 26.6 | +7% |
| bicycles | 8.0 | 7.5 | -6% |
| progressive | 22.6 | 22.4 | ~0% |

The `upsampling` test case specifically exercises the upsampling code path and shows a 7x speedup. Other images that don't use upsampling show no significant regression.

### DIFF
--- a/jxl/src/render/stages/upsample.rs
+++ b/jxl/src/render/stages/upsample.rs
@@ -4,6 +4,7 @@
 // license that can be found in the LICENSE file.
 
 #![allow(clippy::needless_range_loop)]
+#![allow(clippy::too_many_arguments)]
 
 use std::any::Any;
 


zond:
```
MULTI-FILE BENCHMARK RESULTS (4 files, 2 revisions)
  CPU architecture: x86_64

[ 1] ccb8361e Add SIMD upsampling for 2x, 4x, and 8x
     vs a8b168de Add min and store_interleaved functions to jxl_simd
     bicycles.jxl    |                    │                     | 0.998 / prev 
     cafe_5.jxl      |                    │                     | 1.002 / prev 
     progressive.jxl |                    │                     | 0.999 / prev 
     upsampling.jxl  |                    ┃                │┤   | 6.114 / prev ▲ faster (6.11 / prev)
                       Scale: -4.76 to 6.76 (1.0 = same speed)

[ 2] a8b168de Add min and store_interleaved functions to jxl_simd (oldest, baseline for comparisons)

```

jxl/src/render/stages/upsample.rs
veluca93: Instead of allocating a row on the stack and proceeding in chunks, we probably should use the `_state` parameter to allocate an appropriate number of `Vec`s that we can then resize to the length of the row buffer.

jxl/src/render/stages/upsample.rs
```diff
@@ -57,6 +76,289 @@ impl<const N: usize, const SHIFT: u8> std::fmt::Display for Upsample<N, SHIFT> {
...
+                    }
+                }
+
+                cx += simd_width;
+            }
+
+            chunk_start = chunk_end;
+        }
+
+        // Scalar remainder for any pixels not covered by SIMD chunks
```
veluca93: We shouldn't need this.

jxl/src/render/stages/upsample.rs
```diff
@@ -57,6 +76,289 @@ impl<const N: usize, const SHIFT: u8> std::fmt::Display for Upsample<N, SHIFT> {
...
 }
 
+// Chunk size for processing - matches libjxl kChunkSize
+const CHUNK_SIZE: usize = 1024;
+
+// Two-pass SIMD upsampling matching libjxl algorithm
+// Pass 1: Precompute min/max to temp buffers (better cache behavior)
+// Pass 2: Kernel convolution using precomputed min/max
+simd_function!(
+    upsample_simd_dispatch,
```
veluca93: I suspect it'd be better (and faster) to have an implementation for 2x, one for 4x, and one for 8x (which share the min/max logic). Can we do that?

(also, usual comment on windows/chunks :-))

jxl/src/render/stages/upsample.rs
```diff
@@ -57,6 +80,410 @@ impl<const N: usize, const SHIFT: u8> std::fmt::Display for Upsample<N, SHIFT> {
...
+            self.maxs.resize(needed, 0.0);
+        }
+    }
+}
+
+// Shared min/max computation for all upsampling variants
+simd_function!(
+    compute_minmax_simd_dispatch,
+    d: D,
+    fn compute_minmax_simd(
```
veluca93: I don't think this needs its own dispatch - you can make a single function that gets dispatched, which then calls a function that is generic on D (or two -- one for minmax and one for the actual upsampling).

This makes the body in the macro small, and is somewhat more efficient than doing two dynamic dispatches.

jxl/src/render/stages/upsample.rs
```diff
@@ -57,6 +80,410 @@ impl<const N: usize, const SHIFT: u8> std::fmt::Display for Upsample<N, SHIFT> {
...
+            let v4 = D::F32Vec::load(d, &r4[offset..]);
+
+            let col_min_v = v0.min(v1).min(v2).min(v3).min(v4);
+            let col_max_v = v0.max(v1).max(v2).max(v3).max(v4);
+
+            col_min_v.store(&mut col_min[offset..]);
+            col_max_v.store(&mut col_max[offset..]);
+        }
+
+        // Handle remainder elements with scalar code
```
veluca93: We shouldn't need this if we use div_ceil above.

jxl/src/render/stages/upsample.rs
```diff
@@ -57,6 +80,410 @@ impl<const N: usize, const SHIFT: u8> std::fmt::Display for Upsample<N, SHIFT> {
...
+        for ((mins_chunk, maxs_chunk), x) in mins_iter
+            .zip(maxs_iter)
+            .zip((0..xsize).step_by(D::F32Vec::LEN))
+            .take(xsize.div_ceil(D::F32Vec::LEN))
+        {
+            let minval = D::F32Vec::load(d, mins_chunk);
+            let maxval = D::F32Vec::load(d, maxs_chunk);
+            let out_x = x * 4;
+
+            // Row 0
```
veluca93: This starts to be a lot of repeated code - maybe it's best do make a for loop for the rows (also for 8x -- 2x is fine as is).

veluca93 (approved):

veluca93:
Please silence the clippy warnings :-) (`#![allow(clippy::too_many_arguments)]`)

PR #537, opened 2025-12-05
Add SIMD F32 to U8/U16 conversions

hjanuschka:
## Summary

Add SIMD-accelerated versions of F32 to U8 and F32 to U16 conversions (related to #521):

- `f32_to_u8_simd` - vectorized clamp to [0,1], scale, and convert
- `f32_to_u16_simd` - vectorized clamp to [0,1], scale, and convert

Uses `simd_function!` macro for automatic dispatch across instruction sets.

> **Note**: This PR depends on #533 (min function). Please merge #533 first.

## Benchmarks

The data conversion stage is not a significant bottleneck - benchmarks show results within noise (~1%). This PR adds SIMD support for consistency with other stages.

### DIFF
--- a/jxl/src/render/stages/convert.rs
+++ b/jxl/src/render/stages/convert.rs
@@ -8,6 +8,7 @@ use crate::{
     headers::bit_depth::BitDepth,
     render::{Channels, ChannelsMut, RenderPipelineInOutStage},
 };
+use jxl_simd::{F32SimdVec, simd_function};
 
 pub struct ConvertU8F32Stage {
     channel: usize,
@@ -250,6 +251,31 @@ impl std::fmt::Display for ConvertF32ToU8Stage {
     }
 }
 
+// SIMD F32 to U8 conversion
+simd_function!(
+    f32_to_u8_simd_dispatch,
+    d: D,
+    fn f32_to_u8_simd(input: &[f32], output: &mut [u8], max: f32, xsize: usize) {
+        let simd_width = D::F32Vec::LEN;
+        let zero = D::F32Vec::splat(d, 0.0);
+        let one = D::F32Vec::splat(d, 1.0);
+        let scale = D::F32Vec::splat(d, max);
+
+        // Process SIMD vectors using div_ceil (buffers are padded)
+        for (input_chunk, output_chunk) in input
+            .chunks_exact(simd_width)
+            .zip(output.chunks_exact_mut(simd_width))
+            .take(xsize.div_ceil(simd_width))
+        {
+            let val = D::F32Vec::load(d, input_chunk);
+            // Clamp to [0, 1] and scale
+            let clamped = val.max(zero).min(one);
+            let scaled = clamped * scale;
+            scaled.round_store_u8(output_chunk);
+        }
+    }
+);
+
 impl RenderPipelineInOutStage for ConvertF32ToU8Stage {
     type InputT = f32;
     type OutputT = u8;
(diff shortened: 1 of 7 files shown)


zond:
```
MULTI-FILE BENCHMARK RESULTS (2 files, 3 revisions)
  CPU architecture: x86_64

[ 1] 9b04cd04 Add SIMD F32 to U8/U16 conversions
     vs a8b168de Add min and store_interleaved functions to jxl_simd
     bike.jxl  |                ├──│╂──────┤              | 0.997 / prev 
     noise.jxl |                ├───╂│────┤               | 1.005 / prev 

[ 2] a8b168de Add min and store_interleaved functions to jxl_simd
     bike.jxl  |                   ├╂───│────┤            | 1.024 / prev 
     noise.jxl |           ├──────│─┤                     | 0.989 / prev 

[ 3] 4127ca64 Precompute cosines in spline rendering (#524) (oldest, baseline for comparisons)

```

Changes seem pretty small, more or less within the noise - what files did you test with?

sboukortt:
The `store_interleaved` function seems unrelated. Perhaps it should be reverted until it’s actually used?

hjanuschka:
@sboukortt  Done! Removed the `store_interleaved_*` functions and split them into a separate PR: #550

jxl/src/render/stages/convert.rs
```diff
@@ -250,6 +251,39 @@ impl std::fmt::Display for ConvertF32ToU8Stage {
...
+
+        // Process SIMD vectors
+        while x + simd_width <= xsize {
+            let val = D::F32Vec::load(d, &input[x..]);
+            let clamped = val.max(zero).min(one);
+            let scaled = clamped * max_vec;
+            scaled.store(&mut temp);
+
+            for i in 0..simd_width {
+                output[x + i] = temp[i].round() as u8;
```
veluca93: This code is probably not significantly faster than scalar code (also, we should be using more iterators and don't need to process the scalar remainder :-))

What do you think of a `convert_store_u8` (and friends, i.e. `_u16`, and maybe `_f16`), that takes care of this step?

hjanuschka: lets do the convert_store_u8

jxl_simd/src/x86_64/avx.rs
```diff
@@ -453,6 +384,70 @@ unsafe impl F32SimdVec for F32VecAvx {
         I32VecAvx(_mm256_castps_si256(this.0), this.1)
     });
 
+    #[inline(always)]
+    fn clamp_scale_store_u8(self, scale: Self, dest: &mut [u8]) {
```
veluca93: Given that the clamp and scale parts are the same for all architectures, I'd prefer if this were just a "convert_store_u(8|16)" (which rounds to the nearest integer, and has unspecified behaviour if the result would overflow the resulting data type), and then we did the scaling and clamping in the caller.

jxl/src/render/stages/convert.rs
```diff
@@ -298,6 +324,32 @@ impl std::fmt::Display for ConvertF32ToU16Stage {
...
+        for (input_chunk, output_chunk) in input
+            .chunks_exact(simd_width)
+            .zip(output.chunks_exact_mut(simd_width))
+            .take(xsize / simd_width)
+        {
+            let val = D::F32Vec::load(d, input_chunk);
+            val.clamp_scale_store_u16(scale, output_chunk);
+        }
+
+        // Handle remainder with scalar code
```
veluca93: (we don't need this if we use div_ceil above)

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -511,6 +511,50 @@ unsafe impl F32SimdVec for F32VecAvx512 {
...
+            let i32s = _mm512_cvtps_epi32(rounded);
+            // Use pmovusdb: saturating conversion from 32-bit to 8-bit unsigned
+            let u8s = _mm512_cvtusepi32_epi8(i32s);
+            // Store 16 bytes
+            // SAFETY: we checked dest has enough space
+            unsafe {
+                _mm_storeu_si128(dest.as_mut_ptr() as *mut __m128i, u8s);
+            }
+        }
+        // SAFETY: avx512f/bw is available from the safety invariant on the descriptor.
```
veluca93: That's not actually true with the current descriptor definition, we should update the safety invariant there and check for bw in new().

In practice there's no downside in doing this since apart from knights landing / knights mill there are no CPUs that support BW but not F, and I don't think the Knights coprocessors still have relevant usage.

PR #538, opened 2025-12-05
Implement flattened modular trees

### LINKED ISSUE #521: Speed improvements from #506

hjanuschka:
## Summary
- Adds `FlatTreeNode` struct that packs parent + both children info to evaluate 3 nodes per tree iteration instead of 1
- Uses BFS traversal to build the flat tree representation during `Tree::read()`
- The `predict_flat` function uses this flattened structure to reduce branch mispredictions during modular decoding

## Benchmarks
| Image | Before | After | Change |
|-------|--------|-------|--------|
| patches_lossless | 9.0 MP/s | 10.6 MP/s | +17% |
| delta_palette | 7.0 MP/s | 8.4 MP/s | +20% |
| grayscale | 19.9 MP/s | 22.0 MP/s | +11% |
| lz77_flower | 2.9 MP/s | 3.9 MP/s | +35% |
| cafe (VarDCT) | 30.6 MP/s | 31.5 MP/s | +3% |

Fixes #521

### DIFF
--- a/jxl/src/frame/modular/tree.rs
+++ b/jxl/src/frame/modular/tree.rs
@@ -34,6 +34,8 @@ pub enum TreeNode {
 
 /// Flattened tree node for optimized traversal (matches C++ FlatDecisionNode).
 /// Stores parent + info about both children to evaluate 3 nodes per iteration.
+// TODO(hjanuschka): investigate performance of using a Rust enum here, and whether
+// separating internal nodes and leaves into two arrays could save a branch.
 #[derive(Debug, Clone, Copy)]
 pub(super) struct FlatTreeNode {
     property0: i32,                    // Property to test, -1 if leaf
@@ -273,12 +275,12 @@ pub(super) fn predict_flat(
         } else {
             0
         };
-        let off1 = 2 | if property_buffer[node.properties_or_offset[1] as usize]
+        let off1 = if property_buffer[node.properties_or_offset[1] as usize]
             <= node.splitvals_or_multiplier[1]
         {
-            1
+            3
         } else {
-            0
+            2
         };
 
         pos = (node.child_id + if p0 { off1 } else { off0 }) as usize;


zond:
```
MULTI-FILE BENCHMARK RESULTS (9 files, 2 revisions)
  CPU architecture: x86_64

[ 1] 0656dcdb Implement flattened modular trees for reduced branch misp...
     bicycles.jxl              |              ├│─┤  ┃                     | 0.926 / prev ▼ slower (0.93 / prev)
     bike_5.jxl                |                    ├─│─┤                 | 1.033 / prev ▲ faster (1.03 / prev)
     cafe.jxl                  |                  ├─│┤                    | 0.997 / prev 
     cafe_5.jxl                |                    ├─│─┤                 | 1.034 / prev ▲ faster (1.03 / prev)
     delta_palette.jxl         |                   ├│─┤                   | 0.999 / prev 
     grayscale.jxl             |                   ├│┤                    | 1.003 / prev 
     green_queen_vardct_e3.jxl |                    │┤                    | 1.006 / prev ▲ faster (1.01 / prev)
     lz77_flower.jxl           |                    ┃             ├│──┤   | 1.246 / prev ▲ faster (1.25 / prev)
     patches.jxl               |              ├│──┤ ┃                     | 0.928 / prev ▼ slower (0.93 / prev)
                                 Scale: 0.69 to 1.31 (1.0 = same speed)

```

jxl/src/frame/modular/decode/channel.rs
```diff
@@ -59,8 +59,8 @@ fn decode_modular_channel_small(
         let row_toptop = &mut row_toptop[IMAGE_OFFSET.0..IMAGE_OFFSET.0 + size.0];
         for x in 0..size.0 {
             let prediction_data = PredictionData::get_rows(row, row_top, row_toptop, x, y);
-            let prediction_result = predict(
-                &tree.nodes,
+            let prediction_result = predict_flat(
```
veluca93: I don't think it's a good idea to use the flat tree in this code path - most likely building the tree will be more expensive than decoding.

jxl/src/frame/modular/tree.rs
```diff
@@ -173,6 +199,97 @@ pub(super) fn predict(
     }
 }
 
+/// Optimized prediction using flat tree (matches C++ context_predict.h:351-371).
+#[inline]
+#[allow(clippy::too_many_arguments)]
+pub(super) fn predict_flat(
```
veluca93, line 228: We should factor out all the code except the tree navigation itself to a single function shared between flat and non-flat tree prediction.

jxl/src/frame/modular/decode/channel.rs
hjanuschka: Good point! Reverted to use the standard `predict()` function for small channels. Also removed the eager flat tree construction from `Tree::read()` since it's now only built on-demand by the specialized tree decoders.

jxl/src/frame/modular/tree.rs
hjanuschka: Done. Factored out the property computation into a shared `compute_properties()` function that both `predict()` and `predict_flat()` now use.

jxl/src/frame/modular/tree.rs
```diff
@@ -32,6 +32,30 @@ pub enum TreeNode {
     },
 }
 
+/// Flattened tree node for optimized traversal (matches C++ FlatDecisionNode).
+/// Stores parent + info about both children to evaluate 3 nodes per iteration.
+#[derive(Debug, Clone, Copy)]
+pub(super) struct FlatTreeNode {
+    property0: i32,                    // Property to test, -1 if leaf
```
veluca93, line 41: We should leave a TODO to investigate the performance implications of using a Rust enum here.

jxl/src/frame/modular/tree.rs
```diff
@@ -173,6 +220,71 @@ pub(super) fn predict(
...
+        // Split node: C++ logic from context_predict.h:361-365
+        let p0 = property_buffer[node.property0 as usize] <= node.splitval0_or_predictor;
+        let off0 = if property_buffer[node.properties_or_offset[0] as usize]
+            <= node.splitvals_or_multiplier[0]
+        {
+            1
+        } else {
+            0
+        };
+        let off1 = 2 | if property_buffer[node.properties_or_offset[1] as usize]
```
veluca93: Can we just write 3 / 2 in the `if` branches?

jxl/src/frame/modular/tree.rs
veluca93: Another thing that *might* make sense is to keep two separate arrays, one for the flattened internal nodes and one for the leaves. Then, we can index the leaves after the internal nodes, which should save us a branch in the decoding loop.

veluca93 (approved):

PR #540, opened 2025-12-05
Add jxl-perfhistory benchmark CI stage

hjanuschka:
## Summary

Adds automatic performance benchmarking to every PR using [jxl-perfhistory](https://github.com/zond/jxl-perfhistory).

## Features

- Runs on every PR automatically
- Posts benchmark results as a PR comment for easy visibility
- Never fails the PR (`continue-on-error: true`)
- Compares last 5 revisions using statistical analysis
- Results also shown in step summary and uploaded as artifact

## Example Output

The benchmark will post a comment like:

```
## 📊 Performance Benchmark (bike 2048x2560)

================================================================================
Performance Graph (with credible intervals):
...
================================================================================

Generated by jxl-perfhistory
```

This helps reviewers quickly see if a PR has any performance impact.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -244,6 +244,8 @@ jobs:
         mkdir testimages
         cp jxl/resources/test/conformance_test_images/sunset_logo.jxl testimages
         cp jxl/resources/test/conformance_test_images/bike.jxl testimages
+        cp jxl/resources/test/green_queen_modular_e3.jxl testimages
+        cp jxl/resources/test/green_queen_vardct_e3.jxl testimages
 
     - name: Cache benchmark binaries
       uses: actions/cache@v4


veluca93:
GitHub workers are not exactly reliable for speed measurements, so I am not convinced by running benchmarks as a CI step.

hjanuschka:
i know they are not scientifically reliable - but wouldn't the worker compare main and PR branch in the same machine, so atleast the percentage should be an indicator?

zond:
Yeah, as long as the worker isn't incredibly noisy from second to second the measurement should be internally credible at least.

veluca93:
The amount of noise will probably be noticeable. Also, different runs will run on different CPUs, so it will be hard to use this as much more than a FYI.

As long as we're aware of that, I guess it's fine.

zond:
> The amount of noise will probably be noticeable. Also, different runs will run on different CPUs, so it will be hard to use this as much more than a FYI.
> 
> As long as we're aware of that, I guess it's fine.

The amount of noise is (after I have run the perf test a few times on GitHub CI workers) actually less than on my laptop, which surprised me.

Also, this measures revisions on the same machine (the one in the PR vs it's ancestors), so different CPU's is fine if we only measure relative values.

.github/workflows/pull_request.yml
```diff
@@ -216,3 +216,82 @@ jobs:
...
+        cat benchmark.txt >> $GITHUB_OUTPUT
+        echo "EOF" >> $GITHUB_OUTPUT
+
+    - name: Post benchmark results as PR comment
+      uses: peter-evans/create-or-update-comment@v4
+      if: always()
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        body: |
+          ## 📊 Performance Benchmark (bike 2048x2560)
```
veluca93: This says bike, but above it claims to use both bike and cafe.

IIRC those images are similar (both coded with VarDCT, similar size), we probably should select a set of 2-3 images that are more different (for example, bike and sunset_logo, although this might be a bit extreme)

zond: Removed the lying comment, and changed to bike and sunset_logo.

hjanuschka:
awesome so it is happening? can the bot post comment (ideally update; on sequential runs).

zond:
> awesome so it is happening? can the bot post comment (ideally update; on sequential runs).

It can't post comments from fork PRs, but I think it can from branch PRs. I leave it to a later PR to fix that, if necessary.

zond:
> awesome so it is happening? can the bot post comment (ideally update; on sequential runs).

https://github.com/libjxl/jxl-rs/pull/542 fixes it, it seems. See https://github.com/libjxl/jxl-rs/pull/543

hjanuschka:
love it thanks!!

jonnyawsom3:
Personally, I think this would be more useful if it always compared against main, and printed a table with the results instead.
Here's an example based on my own bench of https://github.com/libjxl/jxl-rs/pull/598
```
MULTI-FILE BENCHMARK RESULTS (4 Files 5 Repetitions)
  CPU architecture: x86_64
  WARNING: System appears noisy: high system load (2.09). Results may be unreliable.
Statistics:
  Repetitions:             5
  Confidence:          99.0%
  Max relative error:   3.0%
  ```
  | https://github.com/libjxl/jxl-rs/commit/f7d6dc9831667ba2ce368371e3a7933ea801b5f6 vs https://github.com/libjxl/jxl-rs/commit/7fc5769a8ed93cac0f75e8b401c5baa375509277 (MP/s) | Main | PR | % Diff
|-----------|--------------|----------|---|
| bike.jxl | 5.768 | 16.600 | +59.4% |
| green_queen_modular_e3.jxl | 6.521 | 6.689 | +2.69% |

Older commits could be checked via edit history of the comment, and I think the CLI could probably just output MP/s to 3 decimals, matching libjxl and making it easier to read too.

The current 'sliders' to display difference is hard to read with the scale changing every commit and no actual numbers to reference. Comparing each commit to each other means you need to recursively add/subtract them to find the final difference to Main, which is even harder with no numbers and differing scales.

zond:
Sounds good, send a PR to https://github.com/zond/jxl-perfhistory!

hjanuschka:
+1 the current output is to complex and compare against main would make most sense.

how about making it markdown friendly (started this a few days ago: https://github.com/zond/jxl-perfhistory/pull/2)

hjanuschka:
@jonnyawsom3 addressed your feedback in my jxl-perfhistory#2 PR

PR #547, opened 2025-12-07
Implement `AtomicRefCell`

tirr-c:
I tried to reason about unsafe blocks as much as I can.

### DIFF
--- a/jxl/src/util/atomic_refcell/internal.rs
+++ b/jxl/src/util/atomic_refcell/internal.rs
@@ -1,0 +1,247 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+use std::marker::PhantomData;
+use std::ops::{Deref, DerefMut};
+use std::ptr::NonNull;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use super::AtomicRefCell;
+
+const MUT_BIT: usize = !(usize::MAX >> 1);
+
+impl<T: ?Sized> AtomicRefCell<T> {
+    #[inline]
+    fn ptr(&self) -> NonNull<T> {
+        // SAFETY: Pointer returned by `UnsafeCell::get` is non-null.
+        unsafe { NonNull::new_unchecked(self.data.get()) }
+    }
+}
+
+/// Indicator that a shared reference to the data is successfully acquired.
+// Safety invariant: `counter > 0`, and `MUT_BIT` is not set (`counter & MUT_BIT == 0`). The
+// invariants imply that no `BorrowTokenMut` exists that points to the same counter.
+struct BorrowToken<'a>(&'a AtomicUsize);
+
+impl<'a> BorrowToken<'a> {
+    /// Ensures that there's no mutable borrow of the data, and increments the reference counter.
+    ///
+    /// It is guaranteed that there's no instance of `BorrowTokenMut` that points to the same
+    /// counter if this method returned `Some`.
+    #[inline]
+    fn borrow(counter: &'a AtomicUsize) -> Option<Self> {
+        let mut prev_counter = counter.load(Ordering::Relaxed);
+        let success = loop {
+            if prev_counter & MUT_BIT != 0 {
+                // Mutable borrow exists.
+                break false;
+            }
+
+            // Ensure that the safety invariant holds after incrementing the counter.
+            let next_counter = prev_counter + 1;
+            if next_counter & MUT_BIT != 0 {
+                // Counter overflowed; treat as failure.
+                break false;
+            }
+
+            // Use compare-exchange to ensure that the counter didn't change since the last time.
+            // Acquire ordering synchronizes with Release used in `BorrowToken{,Mut}::drop`.
+            // Ordering of other accesses doesn't matter, because the borrow happens only when
+            // compare-exchange succeeds.
+            match counter.compare_exchange_weak(
+                prev_counter,
+                next_counter,
+                Ordering::Acquire,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => break true,
+                Err(counter) => {
+                    // Compare-exchange failed; retry.
+                    prev_counter = counter;
+                }
+            }
+        };
+
+        // The safety invariants hold if `success` is true.
+        success.then(|| Self(counter))
+    }
+}
+
+impl Clone for BorrowToken<'_> {
+    #[inline]
+    fn clone(&self) -> Self {
+        Self::borrow(self.0).unwrap()
+    }
+}
+
+impl Drop for BorrowToken<'_> {
+    #[inline]
+    fn drop(&mut self) {
+        // Decrement the reference counter.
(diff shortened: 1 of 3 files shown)


jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,231 @@
...
+        // Acquire ordering synchronizes with Release used in `release` and `release_mut`.
+        self.0
+            .compare_exchange(0, MUT_BIT, Ordering::Acquire, Ordering::Relaxed)
+            .is_ok()
+    }
+
+    #[inline]
+    fn release(&self) {
+        // Decrement the reference counter.
+        self.0.fetch_sub(1, Ordering::Release);
```
tirr-c: I wonder if this could be `Relaxed` because the reference was immutable anyway?

jxl/src/util/atomic_refcell/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub const fn new(value: T) -> Self {
+        Self {
+            counter: AtomicRefCounter::new(),
+            data: UnsafeCell::new(value),
+        }
+    }
+}
+
+impl<T: ?Sized> AtomicRefCell<T> {
+    #[inline]
```
veluca93: AFAICT this is only used in the internal module, so it might make sense to move it there.

jxl/src/util/atomic_refcell/internal.rs
veluca93: No, because then the compiler/cpu could move reads *after* calls to release(), and in particular potentially after some other thread has acquire-d() and holds a mutable reference.

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+    fn acquire_mut(&self) -> bool {
+        // Use compare-exchange to ensure that there's no other reference to the data.
+        // Acquire ordering synchronizes with Release used in `release` and `release_mut`.
+        self.0
+            .compare_exchange(0, MUT_BIT, Ordering::Acquire, Ordering::Relaxed)
+            .is_ok()
+    }
+
+    #[inline]
+    fn release(&self) {
```
veluca93: As written, this function should probably be `unsafe` (because if safe code were to call it, it could cause acquire_mut to succeed when it shouldn't).

However, overall I wonder if we shouldn't instead remove `AtomicRefCounter` entirely, and move the acquire/release logic inside the `BorrowToken(Mut)`. Then, we can make them have the appropriate safety invariant on the value of the AtomicUsize (i.e. == MUT_BIT for BorrowTokenMut, and MUT_BIT is not set and the other bits represent a strictly positive value for BorrowToken)

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+pub struct AtomicRef<'a, T: ?Sized> {
+    // Token ensures that this is a valid shared reference to `ptr`.
+    token: BorrowToken<'a>,
+    ptr: NonNull<T>,
+}
+
+// SAFETY: `AtomicRefMut` acts like a shared reference (see `deref`).
+unsafe impl<'a, T: ?Sized> Send for AtomicRef<'a, T> where for<'r> &'r T: Send {}
+
+// SAFETY: `AtomicRefMut` acts like a shared reference (see `deref`).
```
veluca93: Wrong name :-)

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+unsafe impl<'a, T: ?Sized> Sync for AtomicRef<'a, T> where for<'r> &'r T: Sync {}
+
+impl<'a, T: ?Sized> AtomicRef<'a, T> {
+    #[inline]
+    pub(super) fn new(cell: &'a AtomicRefCell<T>) -> Option<Self> {
+        let token = BorrowToken::borrow(&cell.counter)?;
+        // SAFETY: The token and the pointer come from the same `AtomicRefCell`.
```
veluca93: I usually write this as `Safety note:` because there is no "real" unsafe block, fwiw.

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl Drop for BorrowTokenMut<'_> {
+    #[inline]
+    fn drop(&mut self) {
+        self.0.release_mut();
+    }
+}
+
+// Invariant: `token` and `ptr` are derived from the same `AtomicRefCell`.
```
veluca93: I'd write something like:
"Safety invariant: `prt` is valid for reads, and while the AtomicRef is live, `ptr` satisfies the requirements to create a *shared* reference from it. This is ensured by `ptr` only borrowing data protected by the AtomicRefCell that `token` was obtained from."

(and similarly for the mut variant)

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+        // SAFETY: The token and the pointer come from the same `AtomicRefCell`.
+        Some(Self {
+            token,
+            ptr: cell.ptr(),
+        })
+    }
+
+    #[inline]
+    pub fn map<U: ?Sized>(orig: Self, f: impl FnOnce(&T) -> &U) -> AtomicRef<'a, U> {
+        AtomicRef {
```
veluca93, line 151: Should also have a `Safety note:`

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+}
+
+impl<T: ?Sized> Deref for AtomicRef<'_, T> {
+    type Target = T;
+
+    #[inline]
+    fn deref(&self) -> &T {
+        // SAFETY: `ptr` is synchronized by `token` by the invariant of `AtomicRef`, and the
```
veluca93: With the updated safety comment, this would become something like "SAFETY: the requirements for as_ref() are part of `self`'s safety invariant".

jxl/src/util/atomic_refcell/internal.rs
```diff
@@ -0,0 +1,253 @@
...
+        // invariant of `token` ensures that we have a valid shared reference to `ptr`.
+        unsafe { self.ptr.as_ref() }
+    }
+}
+
+impl<T: std::fmt::Debug> std::fmt::Debug for AtomicRef<'_, T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("AtomicRef").field(self).finish()
```
veluca93: Shouldn't the `self` be a `&**self`?

jxl/src/util/atomic_refcell/internal.rs
tirr-c: Ah fair point, then relaxed ordering is incorrect here.

jxl/src/util/atomic_refcell/internal.rs
tirr-c: Oh absolutely, this would recurse infinitely otherwise...

PR #556, opened 2025-12-11
Add FFI-friendly API and convenience methods for Chromium integration

hjanuschka:
## Summary

This PR adds several features to simplify FFI usage, particularly for the Chromium jxl-rs wrapper.

### Direct RGBA output with format selection
- Add `JxlDataFormat` convenience constructors: `u8()`, `u16()`, `f16()`, `f32()`
- Add `JxlPixelFormat` convenience constructors: `rgba8()`, `rgba16()`, `rgba_f32()`, `bgra8()`, `rgb8()` with `bytes_per_pixel()` helper
- Add `JxlBasicInfo::buffer_size()` and `num_extra_channels()` helpers
- Add `JxlDecoder<WithFrameInfo>::decode_to_buffer()` for direct buffer output

### JxlDecoderSimple FFI-friendly API
- Add `JxlDecoderSimple` struct with runtime state checking instead of typestate, making it easier to use from FFI bindings
- Add `DecoderState` enum: `Initialized`, `WithImageInfo`, `WithFrameInfo`
- Add `process()`, `decode_frame()`, `skip_frame()` methods
- Add `InvalidDecoderState` error variant for API misuse

### Animation rewind_for_animation() method
- Add `rewind_for_animation()` that preserves `pixel_format` across rewinds
- Fix `pixel_format` preservation bug in header parsing
- Simplifies animation loop playback for FFI users

## Impact

These changes simplify the Chromium wrapper from ~630 lines to potentially ~200 lines by:
- Eliminating typestate flattening code (use `JxlDecoderSimple` instead)
- Reducing buffer management boilerplate
- Providing direct pixel format constructors

Reference: #521

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -652,8 +652,7 @@ pub(crate) mod tests {
             }
         };
 
-        // Set RGB format with no extra channels - this means num_api_buffers = 1
-        // Reference frame save stages will target indices beyond this
+        // Set RGB format with no extra channels
         let rgb_format = JxlPixelFormat {
             color_type: JxlColorType::Rgb,
             color_data_format: Some(JxlDataFormat::f32()),


jxl/src/api/inner/mod.rs
```diff
@@ -102,16 +131,123 @@ impl JxlDecoderInner {
...
+    /// Rewinds for animation loop replay, keeping pixel_format setting.
+    ///
+    /// This resets the decoder but preserves the pixel_format configuration,
+    /// so the caller doesn't need to re-set it after rewinding.
+    ///
+    /// After calling this, the caller should provide input from the beginning of the file.
+    /// Headers will be re-parsed, then frames can be decoded again.
+    ///
+    /// Returns `true` if pixel_format was preserved, `false` if none was set.
+    pub fn rewind_for_animation(&mut self) -> bool {
```
veluca93: I think I would prefer calling these `reset()` and `rewind()` respectively.

jxl/src/api/inner/mod.rs
```diff
@@ -102,16 +131,123 @@ impl JxlDecoderInner {
...
         self.codestream_parser.has_more_frames
     }
 
+    /// Process input, advancing the decoder state.
+    ///
+    /// The input slice is consumed as needed. After this call, the caller can check
+    /// how many bytes were consumed by examining the slice length change.
+    ///
+    /// Returns `Ok(true)` when the state has advanced, `Ok(false)` when more input is needed.
+    pub fn process_with_input(&mut self, input: &mut &[u8]) -> Result<bool> {
```
veluca93: Why do we need these functions over just calling `self.process` directly? Is it just to check whether the state is valid?

If so, then these methods should instead take `JxlInputBuffer` and `JxlOutputBuffer`s like process().

jxl/src/api/decoder.rs
```diff
@@ -176,12 +193,49 @@ impl JxlDecoder<WithFrameInfo> {
         let inner_result = self.inner.process(input, Some(buffers))?;
         Ok(self.map_inner_processing_result(inner_result))
     }
+
+    /// Convenience method to decode directly into a flat byte buffer.
```
veluca93: I'd prefer not to add this method.

jxl/src/api/mod.rs
```diff
@@ -71,3 +71,36 @@ pub struct JxlBasicInfo {
     pub tone_mapping: ToneMapping,
     pub preview_size: Option<(usize, usize)>,
 }
+
+impl JxlBasicInfo {
+    /// Returns the required buffer size in bytes for the given pixel format.
```
veluca93: Let's also point out that this is the required buffer size for outputing the color data of an entire image-sized buffer.

jxl/src/render/low_memory_pipeline/save/identity.rs
veluca93, line 1: Can we leave modifying this file to a future PR? We can implement interleaving alpha only in the slow path for now.

jxl/src/render/low_memory_pipeline/save/mod.rs
```diff
@@ -33,7 +33,12 @@ impl SaveStage {
         full_image_size: (usize, usize),
         frame_origin: (isize, isize),
     ) -> Result<()> {
-        let Some(buf) = buffers[self.output_buffer_index].as_mut() else {
+        // Skip if buffer index is out of range (e.g., for reference frame buffers
```
veluca93: Are there valid cases in which the index is out of range? IIRC this should only happen if there are logic errors in the decoder.

jxl/src/render/low_memory_pipeline/save/mod.rs
```diff
@@ -136,6 +156,50 @@ impl SaveStage {
                 }
             }
         }
+
+        // Fill opaque alpha if needed (when RGBA requested but image has no alpha)
```
veluca93: I believe this would be better implemented as follows:
- Let's add additional `RowBuffer`s to the LowMemoryRenderPipeline (per save stage), with one row of storage each, which are only populated if fill_opaque_alpha is true.
- Those `RowBuffer`s are populated with a fully-opaque alpha value *for the correct bitdepth* (now we'd use `65535u16` even if requesting 10-bit output, which is incorrect).
- Then, we can just append a reference to that RowBuffer in the vector we pass as `data`; this removes the need for handling `fill_opaque_alpha` in any special way, and so should get rid of ~all the changes to this submodule.

We should follow a similar approach also in the `simple_render_pipeline`, and add tests that compare that filling opaque alpha has the same behaviour with both pipelines.

jxl/src/render/low_memory_pipeline/mod.rs
```diff
@@ -380,6 +380,12 @@ impl RenderPipeline for LowMemoryRenderPipeline {
             match s {
                 Stage::Extend(e) => size = e.image_size,
                 Stage::Save(s) => {
+                    // Skip bounds check if the buffer index is out of range.
```
veluca93: (same question as above wrt this actually happening)

jxl/src/api/inner/mod.rs
hjanuschka: done

jxl/src/api/inner/mod.rs
hjanuschka: done

jxl/src/render/low_memory_pipeline/save/mod.rs
hjanuschka: had issues with the newton one! 8bit animation, will test at the end of PR

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -129,6 +129,26 @@ impl CodestreamParser {
...
 
+    /// Resets for animation loop replay, keeping pixel_format setting.
+    ///
+    /// This resets the decoder but preserves the pixel_format, so the caller
+    /// doesn't need to re-configure it. Headers will be re-parsed when
+    /// input is provided again.
+    ///
+    /// Returns the pixel_format that was set (if any), which will be
+    /// restored after header re-parsing.
+    pub(super) fn rewind_for_animation(&mut self) -> Option<JxlPixelFormat> {
```
veluca93: We should probably rename this too

jxl/src/render/low_memory_pipeline/row_buffers.rs
```diff
@@ -93,4 +93,32 @@ impl RowBuffer {
     pub const fn x0_byte_offset() -> usize {
         CACHE_LINE_BYTE_SIZE
     }
+
+    /// Creates a RowBuffer filled with a repeating byte pattern.
+    /// Used for opaque alpha buffers.
+    pub fn new_filled(data_type: DataTypeTag, row_len: usize, fill_pattern: &[u8]) -> Result<Self> {
+        // Single row, no border needed
```
veluca93: We should still call ::new here to avoid code duplication.

jxl/src/render/low_memory_pipeline/mod.rs
```diff
@@ -447,6 +479,9 @@ impl RenderPipeline for LowMemoryRenderPipeline {
                 origin: (xrange.start, yrange.start),
                 size: (xrange.clone().count(), yrange.clone().count()),
             };
+            if rect_to_render.size.0 == 0 || rect_to_render.size.1 == 0 {
```
veluca93, line 482: I think rebasing got confused here...

jxl/src/render/simple_pipeline/save.rs
```diff
@@ -21,19 +21,26 @@ impl SaveStage {
         for i in self.channels.iter().skip(1) {
             assert_eq!(data[self.channels[0]].size(), data[*i].size());
         }
-        let Some(buf) = buffers[self.output_buffer_index].as_mut() else {
+        // Skip if buffer index is out of range (e.g., for reference frame buffers
```
veluca93: Since we got rid of this in the low memory pipeline code, I imagine we can also remove it from here.

jxl/src/render/low_memory_pipeline/save/mod.rs
hjanuschka: newton still works

jxl/src/render/builder.rs
```diff
@@ -139,12 +139,33 @@ impl<Pipeline: RenderPipeline> RenderPipelineBuilder<Pipeline> {
...
             orientation,
             output_buffer_index,
             color_type,
             data_format,
+            false,
+        )
+    }
+
+    #[instrument(skip_all, err)]
+    pub fn add_save_stage_with_fill_alpha(
```
veluca93: I don't think we need two separate methods here.

jxl/src/render/save.rs
```diff
@@ -17,15 +17,36 @@ pub struct SaveStage {
...
+            channels,
+            orientation,
+            output_buffer_index,
+            color_type,
+            data_format,
+            false,
+        )
+    }
+
+    pub fn new_with_fill_alpha(
```
veluca93: Also here, I think we can just call this `new` and remove the other method.

jxl/src/render/low_memory_pipeline/save/identity.rs
```diff
@@ -30,7 +31,7 @@ pub(super) fn store(
     };
     // SAFETY: we never write uninit memory to the `output_row`.
     let output_buf = unsafe { output_buf.row_mut(output_y) };
-    let output_buf = &mut output_buf[0..(byte_end - byte_start) * input_buf.len()];
+    let output_buf = &mut output_buf[0..(byte_end - byte_start) * output_channels];
```
veluca93: I believe we don't need this change.

jxl/src/render/low_memory_pipeline/save/mod.rs
```diff
@@ -33,7 +33,12 @@ impl SaveStage {
         full_image_size: (usize, usize),
         frame_origin: (isize, isize),
     ) -> Result<()> {
-        let Some(buf) = buffers[self.output_buffer_index].as_mut() else {
+        // Use get_mut() to handle save stages that target buffers not provided
```
veluca93: So we need this after all?
Can we add a test that triggers a failure here? ISTM like this should not happen and, if it does, it is triggered by a bug somewhere else.

jxl/src/render/save.rs
```diff
@@ -42,6 +46,16 @@ impl SaveStage {
             output_buffer_index,
             color_type,
             data_format,
+            fill_opaque_alpha,
+        }
+    }
+
+    /// Returns the number of output channels (including filled alpha if applicable)
+    pub fn output_channels(&self) -> usize {
```
veluca93, line 54: why not always `self.color_type.samples_per_pixel()`?

jxl/src/render/low_memory_pipeline/save/mod.rs
hjanuschka: Yes, we need this. The issue occurs with reference frame save stages.


Added test_animation_with_reference_frames test that exercises this code path by decoding an animation with reference frames using only 1 API buffer.

jxl/src/api/decoder.rs
```diff
@@ -531,4 +531,188 @@ pub(crate) mod tests {
...
+            match decoder.process(&mut input).unwrap() {
+                ProcessingResult::Complete { result } => break result,
+                ProcessingResult::NeedsMoreInput { fallback, .. } => {
+                    decoder = fallback;
+                }
+            }
+        };
+
+        // Set RGB format with no extra channels - this means num_api_buffers = 1
+        // Reference frame save stages will target indices beyond this
```
veluca93: That's no longer true

PR #567, opened 2025-12-18
Ensure that non-section parsing can always make progress.

### LINKED ISSUE #445: Non-section buffer of 4096 bytes is not enough for some huge images

veluca93:
Fixes #445.


veluca93:
> might make sense to add a test image that triggers all codepaths here at some point?!

That's what `jxl/resources/test/large_header.jxl` is for :-)

PR #574, opened 2025-12-19
Add premultiply_output option for premultiplied alpha output

hjanuschka:
## Summary

Add a new option to `JxlDecoderOptions` that allows consumers to request premultiplied alpha output directly from the decoder, rather than doing manual post-processing.

- Add `premultiply_output: bool` option (default: false)
- Implement premultiplication in both render pipelines (simple and low-memory)
- Handle all output formats: U8, U16, F16, F32
- Skip premultiplication if source already has associated alpha to avoid double-premultiplication

This is useful for Chromium/Skia which expect premultiplied alpha for proper compositing.

## Usage

```rust
let mut opts = JxlDecoderOptions::default();
opts.premultiply_output = true;  // Request premultiplied alpha output
```

### DIFF
--- a/jxl/src/render/stages/premultiply_alpha.rs
+++ b/jxl/src/render/stages/premultiply_alpha.rs
@@ -45,16 +45,18 @@ impl PremultiplyAlphaStage {
 
 // SIMD premultiply: color = color * alpha
 simd_function!(
-    premultiply_row_simd_dispatch,
+    premultiply_rows_simd_dispatch,
     d: D,
-    fn premultiply_row_simd(color_row: &mut [f32], alpha_row: &[f32], xsize: usize) {
-        let iter_color = color_row.chunks_exact_mut(D::F32Vec::LEN);
-        let iter_alpha = alpha_row.chunks_exact(D::F32Vec::LEN);
-        for (color_chunk, alpha_chunk) in iter_color.zip(iter_alpha).take(xsize.div_ceil(D::F32Vec::LEN)) {
-            let color_vec = D::F32Vec::load(d, color_chunk);
-            let alpha_vec = D::F32Vec::load(d, alpha_chunk);
-            let result = color_vec * alpha_vec;
-            result.store(color_chunk);
+    fn premultiply_rows_simd(color_rows: &mut [&mut [f32]], alpha_row: &[f32], xsize: usize) {
+        for color_row in color_rows.iter_mut() {
+            let iter_color = color_row.chunks_exact_mut(D::F32Vec::LEN);
+            let iter_alpha = alpha_row.chunks_exact(D::F32Vec::LEN);
+            for (color_chunk, alpha_chunk) in iter_color.zip(iter_alpha).take(xsize.div_ceil(D::F32Vec::LEN)) {
+                let color_vec = D::F32Vec::load(d, color_chunk);
+                let alpha_vec = D::F32Vec::load(d, alpha_chunk);
+                let result = color_vec * alpha_vec;
+                result.store(color_chunk);
+            }
         }
     }
 );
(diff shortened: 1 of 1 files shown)


anforowicz:
Let me quickly comment on this part:

> Skip premultiplication if source already has associated alpha to avoid double-premultiplication

I haven't looked closer, but it seems to me that having explicit alpha channel in the encoded image doesn't necessarily mean that alpha premultiplication is not needed.  For example, RGBA PNG comes out of the `png` crate's decoder as `kUnpremul_SkAlphaType`, and then `SkPngRustCodec` uses Skia APIs to get the final output as `kPremul_SkAlphaType` (well, depending on what final output format `blink::SkiaImageDecoderBase` requests).

hjanuschka:
@anforowicz thanks for taking a look (and in general thanks for your help)

the `alpha_associated` flag in JXL metadata indicates whether the encoded color values are already premultiplied. Unlike PNG (which always stores straight alpha), JXL can natively store premultiplied data.

When `alpha_associated=true`, the decoded RGB values are already multiplied by alpha - applying premultiplication again would result in double-premultiplication (that breaks many cases).

The output will always be premultiplied when `premultiply_output=true` - we just skip the multiplication step when the source data is already in that format. The test `test_premultiply_output_already_premultiplied` verifies this behavior.

jxl/src/api/options.rs
veluca93: Can we remove unpremultiply_alpha? I am not sure why it's there in the first place...

jxl/src/render/low_memory_pipeline/save/mod.rs
```diff
@@ -67,16 +68,22 @@ impl SaveStage {
...
-            Orientation::Identity => identity::store(
-                data,
-                frame_y,
-                save_start.0..save_end.0,
-                buf,
-                relative_y,
-                self.data_format,
-            ),
-            _ => 0,
+        // Fast path for identity orientation when not premultiplying
```
veluca93: I don't think this is the correct way of handling this.
Instead, we should create a new stage that premultiplies alpha into color channels, and gets inserted just before the save stages.

hjanuschka:
feedback addressed, chromium CL still builds and all images work

jxl/src/render/low_memory_pipeline/save/mod.rs
```diff
@@ -101,6 +101,7 @@ impl SaveStage {
             let y0 = y0 as isize;
             let dx = x1 as isize - x0;
             let dy = y1 as isize - y0;
+
```
veluca93: (spurious newline inserted, here and below)

jxl/src/render/stages/premultiply_alpha.rs
```diff
@@ -0,0 +1,129 @@
...
+        let num_channels = row.len();
+        if num_channels < 2 {
+            return;
+        }
+
+        // Alpha is the last channel in the row slice
+        let (color_rows, alpha_row) = row.split_at_mut(num_channels - 1);
+        let alpha_row = &alpha_row[0];
+
+        for color_row in color_rows.iter_mut() {
```
veluca93: Should probably add a TODO to SIMDfy this (or SIMDfy it now :-))

jxl/src/render/stages/premultiply_alpha.rs
```diff
@@ -0,0 +1,144 @@
...
+        let num_channels = row.len();
+        if num_channels < 2 {
+            return;
+        }
+
+        // Alpha is the last channel in the row slice
+        let (color_rows, alpha_row) = row.split_at_mut(num_channels - 1);
+        let alpha_row = &alpha_row[0][..];
+
+        for color_row in color_rows.iter_mut() {
```
veluca93: We could move the loop inside the SIMD code.

PR #575, opened 2025-12-19
Adding CMS stage

zond:


zond:
@veluca93 is this something along the lines of what you had in mind for the CMS stage?

jxl/src/frame/render.rs
```diff
@@ -410,6 +414,27 @@ impl Frame {
             if decoder_state.xyb_output_linear {
                 linear = true;
             } else {
+                // Insert CMS stage if profiles differ and CMS provided
+                if let Some(cms) = cms
+                    && input_profile != output_profile
```
veluca93, line 417: I am not sure we need a CMS for all of those cases, I think there are some conversions we can handle ourselves. Having said that, @sboukortt probably knows better.

jxl/src/render/stages/cms.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl RenderPipelineInPlaceStage for CmsStage {
+    type Type = f32;
+
+    fn uses_channel(&self, c: usize) -> bool {
+        (self.first_channel..self.first_channel + self.num_channels).contains(&c)
+    }
+
+    fn init_local_state(&self) -> Result<Option<Box<dyn Any>>> {
```
veluca93: IMO it would be better to have init_local_state take a thread index here (can be 0 for now), instead of using fetch_add.

Another option would be to require the transformer to be Clone, but I am not sure if that can work, or to use `Option<Box<dyn JxlCms>>` everywhere and create a new transformer on local state creation.

The last two are more work, but I think overall cleaner.

jxl/src/render/stages/cms.rs
```diff
@@ -0,0 +1,114 @@
...
+        _position: (usize, usize),
+        xsize: usize,
+        row: &mut [&mut [f32]],
+        state: Option<&mut dyn Any>,
+    ) {
+        let Some(state) = state else { return };
+        let state: &mut CmsLocalState = state.downcast_mut().unwrap();
+        let nc = self.num_channels;
+
+        // Interleave planar -> packed
```
veluca93: ISTM that we probably should extract fast interleaving from the save stage in the low memory render pipeline (for <=4 channels, at least) and use it here.

We might also want to create an equivalent fast de-interleaving, but that can happen later.

jxl/src/frame/render.rs
zond: I'm not sure I understood you correctly, but check if this makes more sense.

jxl/src/render/stages/cms.rs
zond: Now I forward a thread ID, is that what you had in mind?

jxl/src/render/stages/cms.rs
zond: OK, more SIMD interleaving implemented and used. WDYT?

jxl/src/render/stages/cms.rs
```diff
@@ -0,0 +1,184 @@
...
+    ) {
+        let Some(state) = state else { return };
+        let state: &mut CmsLocalState = state.downcast_mut().unwrap();
+        let nc = self.num_channels;
+
+        // Interleave planar -> packed using SIMD where available
+        match nc {
+            1 => {
+                // No interleaving needed for single channel
+                state.buffer[..xsize].copy_from_slice(&row[0][..xsize]);
```
veluca93: Can we avoid the copying here?

jxl/src/render/stages/cms.rs
```diff
@@ -0,0 +1,184 @@
...
+            .expect("CMS transformer mutex poisoned");
+        transformer
+            .do_transform_inplace(&mut state.buffer[..xsize * nc])
+            .expect("CMS transform failed");
+
+        // De-interleave packed -> planar using SIMD where available
+        match nc {
+            1 => {
+                // No deinterleaving needed for single channel
+                row[0][..xsize].copy_from_slice(&state.buffer[..xsize]);
```
veluca93: And here

jxl/src/render/stages/cms.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Thread-local state for CMS transform.
+struct CmsLocalState {
+    transformer_idx: usize,
+    buffer: Vec<f32>,
+}
+
+/// Applies CMS color transform between color profiles.
+pub struct CmsStage {
+    transformers: Vec<Mutex<Box<dyn JxlCmsTransformer + Send>>>,
```
veluca93: AFAIU, we never want to actually wait on a transform, so we should probably use AtomicRefCell instead.

jxl/src/render/simd_utils.rs
```diff
@@ -0,0 +1,326 @@
...
+
+//! SIMD utilities for interleaving and deinterleaving channel data.
+
+use jxl_simd::{F32SimdVec, simd_function};
+
+simd_function!(
+    interleave_2_dispatch,
+    d: D,
+    /// Interleave 2 planar channels into packed format.
+    pub fn interleave_2(a: &[f32], b: &[f32], out: &mut [f32]) -> usize {
```
veluca93: Can we ensure that the buffers are always large enough to handle xsize inputs with SIMD, like in the rest of the pipeline steps?

I believe this should be easy to do, faster, and need less code.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -273,6 +273,39 @@ unsafe impl F32SimdVec for F32VecNeon {
         unsafe { store_interleaved_8_impl(a.0, b.0, c.0, d.0, e.0, f.0, g.0, h.0, dest) }
     }
 
+    #[inline(always)]
+    fn load_deinterleaved_2(d: Self::Descriptor, src: &[f32]) -> (Self, Self) {
+        assert!(src.len() >= 2 * Self::LEN);
+        // SAFETY: we just checked that `src` has enough space, and neon is available
+        // from the safety invariant on `d`.
+        unsafe {
```
veluca93: Please make unsafe blocks as small as possible - in this case, just have `vld2q_f32` in the unsafe block.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -393,6 +393,100 @@ unsafe impl F32SimdVec for F32VecAvx {
...
+    fn load_deinterleaved_2(d: Self::Descriptor, src: &[f32]) -> (Self, Self) {
+        #[target_feature(enable = "avx2")]
+        #[inline]
+        fn load_deinterleaved_2_impl(src: &[f32]) -> (__m256, __m256) {
+            assert!(src.len() >= 2 * F32VecAvx::LEN);
+            // Input: [a0, b0, a1, b1, a2, b2, a3, b3, a4, b4, a5, b5, a6, b6, a7, b7]
+            // Output: a = [a0, a1, a2, a3, a4, a5, a6, a7], b = [b0, b1, b2, b3, b4, b5, b6, b7]
+            // SAFETY: we just checked that src has enough space.
+            let (in0, in1) = unsafe {
+                let in0 = _mm256_loadu_ps(src.as_ptr()); // [a0,b0,a1,b1, a2,b2,a3,b3]
```
veluca93: I'd prefer avoiding the intermediate variables here (i.e. `unsafe { (_mm256_loadu_ps(...), _mm256_loadu_ps(...)}`)

jxl_simd/src/x86_64/avx.rs
```diff
@@ -393,6 +393,100 @@ unsafe impl F32SimdVec for F32VecAvx {
...
+        let (a, b) = unsafe { load_deinterleaved_2_impl(src) };
+        (Self(a, d), Self(b, d))
+    }
+
+    #[inline(always)]
+    fn load_deinterleaved_3(d: Self::Descriptor, src: &[f32]) -> (Self, Self, Self) {
+        assert!(src.len() >= 3 * Self::LEN);
+        // Input: [a0,b0,c0, a1,b1,c1, a2,b2,c2, a3,b3,c3, a4,b4,c4, a5,b5,c5, a6,b6,c6, a7,b7,c7]
+        // Output: a = [a0..a7], b = [b0..b7], c = [c0..c7]
+        // 3-way deinterleave doesn't have efficient SIMD support on x86, use scalar
```
veluca93: That's not true, it can be implemented in the ~same way that 3-way interleaving is implemented (but opposite)

jxl/src/render/stages/cms.rs

jxl/src/render/stages/cms.rs

jxl/src/render/stages/cms.rs

jxl/src/render/simd_utils.rs
zond: I think it's done the way you want.

jxl_simd/src/aarch64/neon.rs
zond: Done, PTAL

jxl_simd/src/x86_64/avx.rs
zond: Done

jxl_simd/src/x86_64/avx.rs
zond: Done, I think. PTAL

jonnyawsom3:
Hmm, may be worth checking that effort 3 modular file in future. I can't imagine why CMS would only effect that one, but 20% is quite large, even given the usual variance in the action workflow.

zond:
Yes, it's a real slowdown, but I'm yet unable to figure out how to get rid of it. Just adding a panic at the beginning of the CMS stage row processing is enough to make it fast again. I'm still confused.

PR #577, opened 2025-12-21
Add a SmallVec type and optimize the low mem render pipeline.

veluca93:
Also apply a few other optimizations to the render pipeline, to reduce allocations and avoid repeated computation.

Roughly 25.5% speedup on Android, 10.5% on x86 desktop on a large test image.


jxl/src/util/smallvec.rs
```diff
@@ -0,0 +1,231 @@
...
+            let mut tmp = MaybeUninit::uninit();
+            std::mem::swap(&mut tmp, d);
+            // SAFETY: the safety invariant on `other` promises that `odata[i]` is initialized
+            // for all i < other_len. Since we set `olen` to 0, we are not breaking the safety
+            // invariant for when `other` gets dropped.
+            data[*len].write(unsafe { tmp.assume_init() });
+            // Safety note: we just wrote a new element in the first non-initialized slot of
+            // the array.
+            *len += 1;
+        }
```
tirr-c: Can't we just do `swap(&mut data[len..][..other_len], &mut odata[..other_len]); len += other_len;`?

veluca93: indeed we can (well, not exactly that, but close enough)

PR #578, opened 2025-12-21
Implement flush_pixels() for progressive rendering

hjanuschka:
## Summary

- Implement the `flush_pixels()` method that was previously `todo!()`
- Add `enable_flush_pixels` option to `JxlDecoderOptions` (default: `false`)
- Enables progressive rendering by allowing callers to render partially-decoded pixel data

## Implementation

The implementation:
1. Adds `enable_flush_pixels: bool` to `JxlDecoderOptions` (defaults to `false` for backward compatibility)
2. `flush_pixels()` checks this option - returns `Ok(())` immediately if disabled
3. When enabled, calls `decode_and_render_hf_groups()` with an empty group list to render already-decoded data

This works with the existing infrastructure:
- `LowMemoryRenderPipeline` tracks `completed_passes` per group
- `BufferSplitter` manages writing to output buffers  
- Save stages handle format conversion (F32→U8/U16)

## Usage Pattern

```rust
let mut options = JxlDecoderOptions::default();
options.enable_flush_pixels = true;  // Opt-in to progressive flush

loop {
    match decoder.process(&mut input, &mut buffers) {
        ProcessingResult::NeedsMoreInput { fallback, .. } => {
            decoder = fallback;
            let passes = decoder.num_completed_passes();
            if passes > last_passes {
                decoder.flush_pixels(&mut buffers)?;
                // Display buffers - shows progressive refinement
                last_passes = passes;
            }
        }
        ProcessingResult::Complete { .. } => break,
    }
}
```

## Test Plan

- [x] Tested with progressive JXL files showing intermediate renders
- [x] Verified multiple flush calls are idempotent
- [x] Used in [jxl-ui](https://github.com/hjanuschka/jxl-ui) for real-world progressive rendering
- [x] Default behavior unchanged (option is `false` by default)

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -484,6 +484,192 @@ pub(crate) mod tests {
     }
 
     #[test]
+    fn test_flush_pixels_disabled_by_default() {
+        use crate::image::{Image, Rect};
+        // Test that flush_pixels returns Ok(()) when disabled (default)
+        let file = std::fs::read("resources/test/conformance_test_images/progressive_5.jxl").unwrap();
+        let options = JxlDecoderOptions::default(); // enable_flush_pixels defaults to false
+        let mut decoder = JxlDecoder::<states::Initialized>::new(options);
+        let mut input = file.as_slice();
+
+        let mut decoder_with_info = loop {
+            match decoder.process(&mut input).unwrap() {
+                ProcessingResult::Complete { result } => break result,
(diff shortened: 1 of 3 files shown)


hjanuschka:
This implementation is being used in [jxl-ui](https://github.com/hjanuschka/jxl-ui) - a native JPEG XL viewer built with GPUI that demonstrates progressive rendering with visual updates during decode.

PR #580, opened 2025-12-22
Add SIMD optimization for int_to_float conversion

hjanuschka:
SIMD fast paths for the `int_to_float` function which converts custom bit-depth floats stored as i32 back to f32.

**32-bit float**: straightforward bitcast via SIMD.

**16-bit float (f16)**: SIMD handles normal values, zeros, and inf/nan. Subnormals fall back to scalar since they need a variable-iteration normalization loop.

Waiting for perf CI to see the impact.

### DIFF
--- a/jxl_simd/src/aarch64/neon.rs
+++ b/jxl_simd/src/aarch64/neon.rs
@@ -486,7 +486,8 @@ unsafe impl F32SimdVec for F32VecNeon {
         fn prepare_impl(table: &[f32; 8]) -> uint8x16_t {
             // Convert f32 table to BF16 packed in 128 bits (16 bytes for 8 entries)
             // BF16 is the high 16 bits of f32
-            // SAFETY: neon is available from target_feature
+            // SAFETY: neon is available from target_feature, and `table` is large
+            // enough for the loads.
             let (table_lo, table_hi) =
                 unsafe { (vld1q_f32(table.as_ptr()), vld1q_f32(table.as_ptr().add(4))) };
 
@@ -693,7 +694,8 @@ impl I32SimdVec for I32VecNeon {
     #[inline(always)]
     fn store_u16(self, dest: &mut [u16]) {
         assert!(dest.len() >= Self::LEN);
-        // SAFETY: We know neon is available from the safety invariant on `self.1`.
+        // SAFETY: We know neon is available from the safety invariant on `self.1`,
+        // and we just checked that `dest` has enough space.
         unsafe {
             // vmovn narrows i32 to i16 by taking the lower 16 bits
             let narrowed = vmovn_s32(self.0);
--- a/jxl_simd/src/lib.rs
+++ b/jxl_simd/src/lib.rs
@@ -1179,4 +1179,40 @@ mod test {
         }
     }
     test_all_instruction_sets!(test_i32_mul_all_elements);
+
+    fn test_store_u16<D: SimdDescriptor>(d: D) {
+        let data = [
+            0xbabau32 as i32,
+            0x1234u32 as i32,
+            0xdeadbabau32 as i32,
+            0xdead1234u32 as i32,
+            0x1111babau32 as i32,
+            0x11111234u32 as i32,
+            0x76543210u32 as i32,
+            0x01234567u32 as i32,
+            0x00000000u32 as i32,
+            0xffffffffu32 as i32,
+            0x23949289u32 as i32,
+            0xf9371913u32 as i32,
+            0xdeadbeefu32 as i32,
+            0xbeefdeadu32 as i32,
+            0xaaaaaaaau32 as i32,
+            0xbbbbbbbbu32 as i32,
+        ];
+        let mut output = [0u16; 16];
+        for i in (0..16).step_by(D::I32Vec::LEN) {
+            let vec = D::I32Vec::load(d, &data[i..]);
+            vec.store_u16(&mut output[i..]);
+        }
+
+        for i in 0..16 {
+            let expected = data[i] as u16;
+            assert_eq!(
+                output[i], expected,
+                "store_u16 failed at index {}: expected {}, got {}",
+                i, expected, output[i]
+            );
+        }
+    }
+    test_all_instruction_sets!(test_store_u16);
 }
--- a/jxl_simd/src/x86_64/avx.rs
+++ b/jxl_simd/src/x86_64/avx.rs
@@ -607,7 +607,8 @@ unsafe impl F32SimdVec for F32VecAvx {
     #[inline(always)]
     fn prepare_table_bf16_8(_d: AvxDescriptor, table: &[f32; 8]) -> Bf16Table8Avx {
         // For AVX2, vpermps is exact and fast, so we just load the table as-is
-        // SAFETY: avx2 is available from the safety invariant on the descriptor
+        // SAFETY: avx2 is available from the safety invariant on the descriptor,
+        // and `table` has 8 elements, exactly as many as we load.
         Bf16Table8Avx(unsafe { _mm256_loadu_ps(table.as_ptr()) })
     }
 
@@ -681,7 +682,7 @@ unsafe impl F32SimdVec for F32VecAvx {
             let bits = unsafe { _mm_loadu_si128(mem.as_ptr() as *const __m128i) };
             F32VecAvx(_mm256_cvtph_ps(bits), d)
         }
-        // SAFETY: f16c is available from the safety invariant on the descriptor
+        // SAFETY: avx2 and f16c are available from the safety invariant on the descriptor
         unsafe { load_f16_impl(d, mem) }
     }
 
@@ -695,7 +696,7 @@ unsafe impl F32SimdVec for F32VecAvx {
             // SAFETY: dest.len() >= 8 is checked above
             unsafe { _mm_storeu_si128(dest.as_mut_ptr() as *mut __m128i, bits) };
         }
-        // SAFETY: f16c is available from the safety invariant on the descriptor
+        // SAFETY: avx2 and f16c are available from the safety invariant on the descriptor
         unsafe { store_f16_bits_impl(self.0, dest) }
     }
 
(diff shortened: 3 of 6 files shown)


jxl/src/render/stages/convert.rs
```diff
@@ -135,20 +135,184 @@ impl std::fmt::Display for ConvertModularToF32Stage {
...
+        let remainder_start = num_full_chunks * simd_width;
+        for i in remainder_start..xsize {
+            output[i] = f32::from_bits(input[i] as u32);
+        }
+    }
+);
+
+// SIMD 16-bit float (half-precision) to 32-bit float conversion
+// This handles IEEE 754 binary16 format: 1 sign bit, 5 exponent bits, 10 mantissa bits
+simd_function!(
```
veluca93, line 159: I think I would prefer to have a pair of functions I32Vec::store_u16() and F32Vec::load_f16_bits() instead. Those functions function can use _mm256_cvtph_ps on AVX2 (by also requiring the F16C target feature, which is common)  and vcvt_f32_f16 on NEON (although the Rust definition erroneously requires the f16 target feature, so we'd have to use inline assembly for now -- fixed in https://github.com/rust-lang/stdarch/pull/1978), and fall back to scalar on SSE4.2.

We then could add store_f16() -- implemented in a similar way -- and use that to speed up the f16 conversion code..

jxl_simd/src/x86_64/avx.rs
```diff
@@ -507,6 +507,59 @@ unsafe impl F32SimdVec for F32VecAvx {
 
     impl_f32_array_interface!();
 
+    #[inline(always)]
+    fn load_f16_bits(d: Self::Descriptor, mem: &[u16]) -> Self {
+        assert!(mem.len() >= Self::LEN);
+        // Check for F16C at runtime and use hardware conversion if available
+        if is_x86_feature_detected!("f16c") {
```
veluca93: That's not a good idea. Given that f16c is as common as avx2 (if not more), let's just always require f16c for the AVX2 path.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -407,6 +407,30 @@ unsafe impl F32SimdVec for F32VecNeon {
                 vst1_u16(dest.as_mut_ptr(), u16s);
             }
         }
+
+        fn store_f16(this: F32VecNeon, dest: &mut [u16]) {
+            assert!(dest.len() >= F32VecNeon::LEN);
+            // TODO: Use vcvt_f16_f32 once Rust stdarch fix lands
```
veluca93: I think at this point I would just use inline ASM here, but we can do that as a follow-up.

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -557,6 +557,41 @@ unsafe impl F32SimdVec for F32VecAvx512 {
...
+    #[inline(always)]
+    fn load_f16_bits(d: Self::Descriptor, mem: &[u16]) -> Self {
+        assert!(mem.len() >= Self::LEN);
+        // AVX512 implies F16C, so we can always use hardware conversion
+        #[target_feature(enable = "avx512f")]
+        #[inline]
+        unsafe fn load_f16_impl(d: Avx512Descriptor, mem: &[u16]) -> F32VecAvx512 {
+            // SAFETY: mem.len() >= 16 is checked by caller, and avx512f is available
+            unsafe {
+                let bits = _mm256_loadu_si256(mem.as_ptr() as *const __m256i);
```
veluca93: Only the loadu needs to be in an unsafe block.

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -557,6 +557,41 @@ unsafe impl F32SimdVec for F32VecAvx512 {
...
+        assert!(dest.len() >= Self::LEN);
+        // AVX512 implies F16C, so we can always use hardware conversion
+        #[target_feature(enable = "avx512f")]
+        #[inline]
+        unsafe fn store_f16_impl(v: __m512, dest: &mut [u16]) {
+            // SAFETY: dest.len() >= 16 is checked by caller, and avx512f is available
+            unsafe {
+                // _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC = 0
+                let bits = _mm512_cvtps_ph::<0>(v);
+                _mm256_storeu_si256(dest.as_mut_ptr() as *mut __m256i, bits);
```
veluca93: Similarly, only the store needs to be in an unsafe block.

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -557,6 +557,41 @@ unsafe impl F32SimdVec for F32VecAvx512 {
 
     impl_f32_array_interface!();
 
+    #[inline(always)]
+    fn load_f16_bits(d: Self::Descriptor, mem: &[u16]) -> Self {
+        assert!(mem.len() >= Self::LEN);
+        // AVX512 implies F16C, so we can always use hardware conversion
+        #[target_feature(enable = "avx512f")]
+        #[inline]
+        unsafe fn load_f16_impl(d: Avx512Descriptor, mem: &[u16]) -> F32VecAvx512 {
```
veluca93: This function does not need to be unsafe if we move the assert inside.

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -557,6 +557,41 @@ unsafe impl F32SimdVec for F32VecAvx512 {
...
+    fn store_f16(self, dest: &mut [u16]) {
+        assert!(dest.len() >= Self::LEN);
+        // AVX512 implies F16C, so we can always use hardware conversion
+        #[target_feature(enable = "avx512f")]
+        #[inline]
+        unsafe fn store_f16_impl(v: __m512, dest: &mut [u16]) {
+            // SAFETY: dest.len() >= 16 is checked by caller, and avx512f is available
+            unsafe {
+                // _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC = 0
+                let bits = _mm512_cvtps_ph::<0>(v);
```
veluca93: Let's please use `::<{_MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC}>`.

jxl_simd/src/scalar.rs
```diff
@@ -9,6 +9,86 @@ use crate::{U32SimdVec, impl_f32_array_interface};
 
 use super::{F32SimdVec, I32SimdVec, SimdDescriptor, SimdMask};
 
+/// Convert f16 bits (as u16) to f32.
```
veluca93: There's already https://github.com/libjxl/jxl-rs/blob/main/jxl/src/util/float16.rs that has conversion code.

I think we should use that type and code (perhaps by moving the code to the jxl_simd crate), instead of using u16.

jxl/src/render/stages/convert.rs
```diff
@@ -135,20 +135,94 @@ impl std::fmt::Display for ConvertModularToF32Stage {
...
+        for (in_chunk, out_chunk) in input
+            .chunks_exact(simd_width)
+            .zip(output.chunks_exact_mut(simd_width))
+            .take(num_full_chunks)
+        {
+            let val = D::I32Vec::load(d, in_chunk);
+            val.bitcast_to_f32().store(out_chunk);
+        }
+
+        // Handle remainder with scalar
```
veluca93: We should never need these code paths.

jxl/src/render/stages/convert.rs
```diff
@@ -135,20 +135,94 @@ impl std::fmt::Display for ConvertModularToF32Stage {
...
+        // Stack-allocated for common SIMD widths (up to 16 elements for AVX-512)
+        let mut u16_buf = [0u16; 16];
+
+        for (in_chunk, out_chunk) in input
+            .chunks_exact(simd_width)
+            .zip(output.chunks_exact_mut(simd_width))
+            .take(num_full_chunks)
+        {
+            // Convert i32 values to u16 (f16 bit patterns are in lower 16 bits)
+            for (i, &val) in in_chunk.iter().enumerate() {
```
veluca93: I'd rather do this with a I32Vec::store_u16 method.

jxl/src/render/stages/convert.rs
```diff
@@ -135,20 +135,94 @@ impl std::fmt::Display for ConvertModularToF32Stage {
...
+        int_to_float_16bit_simd_dispatch(input, output, xsize);
+        return;
+    }
+
+    // Generic scalar path for other custom float formats
+    int_to_float_generic(input, output, bits, exp_bits);
+}
+
+// Generic scalar conversion for arbitrary bit-depth floats
+fn int_to_float_generic(input: &[i32], output: &mut [f32], bits: u32, exp_bits: u32) {
```
veluca93, line 213: probably should have a TODO to SIMD this too, eventually

jxl_simd/src/x86_64/avx.rs
```diff
@@ -668,6 +671,36 @@ unsafe impl F32SimdVec for F32VecAvx {
...
     impl_f32_array_interface!();
 
+    #[inline(always)]
+    fn load_f16_bits(d: Self::Descriptor, mem: &[u16]) -> Self {
+        assert!(mem.len() >= Self::LEN);
+        // f16c is guaranteed by the safety invariant on AvxDescriptor
+        #[target_feature(enable = "avx2,f16c")]
+        #[inline]
+        fn load_f16_impl(d: AvxDescriptor, mem: &[u16]) -> F32VecAvx {
+            // SAFETY: mem.len() >= 8 is checked by caller
```
veluca93: Like with other loads methods elsewhere, the check should be moved inside the inner function.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -668,6 +671,36 @@ unsafe impl F32SimdVec for F32VecAvx {
 
     impl_f32_array_interface!();
 
+    #[inline(always)]
+    fn load_f16_bits(d: Self::Descriptor, mem: &[u16]) -> Self {
+        assert!(mem.len() >= Self::LEN);
+        // f16c is guaranteed by the safety invariant on AvxDescriptor
```
veluca93: unnecessary comment

jxl_simd/src/lib.rs
```diff
@@ -270,6 +273,16 @@ pub unsafe trait F32SimdVec:
...
+
+    /// Loads f16 values (stored as u16 bit patterns) and converts them to f32.
+    /// Uses hardware conversion instructions when available (F16C on x86, NEON fp16 on ARM).
+    /// Requires `mem.len() >= Self::LEN` or it will panic.
+    fn load_f16_bits(d: Self::Descriptor, mem: &[u16]) -> Self;
+
+    /// Converts f32 values to f16 and stores as u16 bit patterns.
+    /// Uses hardware conversion instructions when available (F16C on x86, NEON fp16 on ARM).
+    /// Requires `dest.len() >= Self::LEN` or it will panic.
+    fn store_f16(self, dest: &mut [u16]);
```
veluca93: if this takes u16, it should be `_bits`. Otherwise, it should take `[f16]`.

jxl/src/render/stages/convert.rs
```diff
@@ -135,20 +135,81 @@ impl std::fmt::Display for ConvertModularToF32Stage {
...
+// Uses hardware F16C/NEON instructions when available via F32Vec::load_f16_bits()
+simd_function!(
+    int_to_float_16bit_simd_dispatch,
+    d: D,
+    fn int_to_float_16bit_simd(input: &[i32], output: &mut [f32], xsize: usize) {
+        let simd_width = D::F32Vec::LEN;
+
+        // Temporary buffer for i32->u16 conversion via SIMD
+        // Stack-allocated for common SIMD widths (up to 16 elements for AVX-512)
+        let mut u16_buf = [0u16; 16];
```
veluca93, line 169: Note that the `16` can also be a `D::F32Vec::LEN` :-)

jxl/src/render/stages/convert.rs
```diff
@@ -163,14 +168,15 @@ simd_function!(
         let simd_width = D::F32Vec::LEN;
 
         // Temporary buffer for i32->u16 conversion via SIMD
-        // Stack-allocated for common SIMD widths (up to 16 elements for AVX-512)
+        // Note: Using constant 16 (max AVX-512 width) because D::F32Vec::LEN
+        // cannot be used as array size in Rust (const generics limitation)
         let mut u16_buf = [0u16; 16];
+        debug_assert!(simd_width <= 16);
```
veluca93: `const { assert!(D::F32Vec::LEN <= 16)}`? That would make this a compilation error.

veluca93 (approved):

PR #581, opened 2025-12-22
Cache default quantization tables per-table using OnceLock

hjanuschka:
## Summary

Apply the same optimization pattern from coefficient orders (#525) to quantization tables.

- Each table encoding is now computed lazily on first access via `get_library(idx)` 
- Large transforms (128x128, 256x256) won't be computed if never used
- Follows the exact pattern used for `NATURAL_COEFF_ORDERS` in `coeff_order.rs`

## Changes

- Add `LIBRARY_QUANTS: [OnceLock<QuantEncoding>; 17]` static array
- Replace `library() -> &'static [QuantEncoding; 17]` with `get_library(idx) -> &'static QuantEncoding`
- Update single call site in `compute_quant_table()`

### DIFF
--- a/jxl/src/frame/quant_weights.rs
+++ b/jxl/src/frame/quant_weights.rs
@@ -371,35 +371,10 @@ impl QuantTable {
 }
 
 pub struct DequantMatrices {
-    computed_mask: u32,
     /// 17 separate tables, one per QuantTable type.
     /// Uses Cow to allow zero-copy borrowing from static cache for library tables.
-    tables: [Option<Cow<'static, [f32]>>; QuantTable::CARDINALITY],
-    encodings: Vec<QuantEncoding>,
+    tables: [Cow<'static, [f32]>; QuantTable::CARDINALITY],
 }
-
-/// Cached library quantization encodings per table type.
-/// Each entry is computed lazily on first access, avoiding computation
-/// of encodings for large transforms that are never used.
-static LIBRARY_QUANTS: [OnceLock<QuantEncoding>; QuantTable::CARDINALITY] = [
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-    OnceLock::new(),
-];
 
 /// Cached computed library tables per QuantTable type.
 /// Each entry contains the computed f32 weights for all 3 channels.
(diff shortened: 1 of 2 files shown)


jxl/src/frame/quant_weights.rs
```diff
@@ -378,12 +372,58 @@ impl QuantTable {
...
+    /// 17 separate tables, one per QuantTable type.
+    /// Uses Cow to allow zero-copy borrowing from static cache for library tables.
+    tables: [Option<Cow<'static, [f32]>>; QuantTable::CARDINALITY],
     encodings: Vec<QuantEncoding>,
 }
 
+/// Cached library quantization encodings per table type.
+/// Each entry is computed lazily on first access, avoiding computation
+/// of encodings for large transforms that are never used.
+static LIBRARY_QUANTS: [OnceLock<QuantEncoding>; QuantTable::CARDINALITY] = [
```
veluca93: I don't think we need to cache this.

jxl/src/frame/quant_weights.rs
```diff
@@ -378,12 +372,58 @@ impl QuantTable {
 
 pub struct DequantMatrices {
     computed_mask: u32,
-    table: Vec<f32>,
-    inv_table: Vec<f32>,
-    table_offsets: [usize; HfTransformType::CARDINALITY * 3],
+    /// 17 separate tables, one per QuantTable type.
+    /// Uses Cow to allow zero-copy borrowing from static cache for library tables.
+    tables: [Option<Cow<'static, [f32]>>; QuantTable::CARDINALITY],
```
veluca93: Why the Option?

jxl/src/frame/quant_weights.rs
```diff
@@ -1190,34 +1120,116 @@ impl DequantMatrices {
                 }
             }
         }
+        // Convert weights to the final table format (1.0 / weight)
+        let mut table = vec![0f32; 3 * num];
```
veluca93: We don't need a separate vector, we can do this in place.

jxl/src/frame/quant_weights.rs
```diff
@@ -1190,34 +1120,116 @@ impl DequantMatrices {
...
+    pub const REQUIRED_SIZE_X: [usize; QuantTable::CARDINALITY] =
+        [1, 1, 1, 1, 2, 4, 1, 1, 2, 1, 1, 8, 4, 16, 8, 32, 16];
+
+    pub const REQUIRED_SIZE_Y: [usize; QuantTable::CARDINALITY] =
+        [1, 1, 1, 1, 2, 4, 2, 4, 4, 1, 1, 8, 8, 16, 16, 32, 32];
+
+    #[cfg(test)]
+    const SUM_REQUIRED_X_Y: usize = 2056;
+
+    pub fn ensure_computed(&mut self, acs_mask: u32) -> Result<()> {
```
veluca93: Now that we cache the library tables, I don't see the reason to have a separate ensure_computed step. Instead, we can just compute the tables as part of decode().

veluca93 (approved):

PR #583, opened 2025-12-22
Optimize EPF sigma for modular encoding

hjanuschka:
## Summary

- Introduce `SigmaSource` enum to handle both constant and variable sigma sources
- For modular images with EPF, avoid allocating a full sigma image just to fill with a constant value
- Add fast path in `get_sigma()` that uses SIMD splat for constant sigma values

This reduces memory allocation overhead for modular-encoded images that use edge-preserving filtering, since they use a single constant sigma value rather than a per-block variable sigma.

### DIFF
--- a/jxl/src/features/epf.rs
+++ b/jxl/src/features/epf.rs
@@ -25,58 +25,60 @@ pub enum SigmaSource {
     Constant(f32),
 }
 
-pub fn create_sigma_source(
-    frame_header: &FrameHeader,
-    lf_global: &LfGlobalState,
-    hf_meta: &Option<HfMetadata>,
-) -> Result<SigmaSource> {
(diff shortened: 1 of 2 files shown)


jxl/src/features/epf.rs
```diff
@@ -12,23 +14,36 @@ use crate::{
...
+/// For Modular encoding, sigma is constant across the entire image.
+#[derive(Clone)]
+pub enum SigmaSource {
+    /// Variable sigma per block (VarDCT encoding)
+    Variable(Arc<Image<f32>>),
+    /// Constant sigma for entire image (Modular encoding)
+    Constant(f32),
+}
+
+pub fn create_sigma_source(
```
veluca93: Why not SigmaSource::new(...)?

PR #585, opened 2025-12-22
Add SIMD table lookup with shuffle-based implementation

hjanuschka:
## Summary
Implements efficient 8-entry table lookup for SIMD using shuffle-based instructions as suggested by @veluca93:

### Implementation Strategy
- **AVX2/AVX512**: Uses `vpermps` (`_mm256_permutevar8x32_ps`, `_mm512_permutexvar_ps`) for exact f32 table lookups
- **SSE4.2**: Uses `pshufb` (`_mm_shuffle_epi8`) with BF16-packed tables for approximate lookups
- **NEON**: Uses `vqtbl1q_u8` with BF16-packed tables for approximate lookups
- **Scalar**: Direct array indexing fallback

### API
Three lookup methods are provided:
1. `table_lookup_8` - Exact lookup (scalar fallback on SSE/NEON, vpermps on AVX)
2. `table_lookup_8_approx` - Fast approximate lookup (BF16 shuffle on SSE/NEON, vpermps on AVX)
3. `prepare_table_bf16_8` + `table_lookup_bf16_8` - Optimized pattern for repeated lookups with same table

### Design Rationale
- No gather instructions used (per review feedback)
- BF16 format for SSE/NEON: stores high 16 bits of f32, allowing 8 entries to fit in 128-bit register for efficient shuffle
- BF16 has ~0.4% relative error - acceptable for LUT approximations
- `prepare_table` pattern allows amortizing table preparation cost across many lookups

### Use Case
This will be used in a follow-up PR for SIMD-accelerated AddNoiseStage, which requires LUT lookups.

## Test plan
- [x] `test_table_lookup_8_*` - verifies exact lookup produces correct results
- [x] `test_table_lookup_8_approx_*` - verifies approximate lookup within BF16 tolerance
- [x] `test_prepare_table_bf16_8_*` - verifies prepare+lookup pattern works correctly
- [x] All existing SIMD tests pass

### DIFF
--- a/jxl_simd/src/aarch64/neon.rs
+++ b/jxl_simd/src/aarch64/neon.rs
@@ -423,7 +423,7 @@ unsafe impl F32SimdVec for F32VecNeon {
         fn prepare_impl(table: &[f32; 8]) -> uint8x16_t {
             // Convert f32 table to BF16 packed in 128 bits (16 bytes for 8 entries)
             // BF16 is the high 16 bits of f32
-            // SAFETY: table has exactly 8 elements and neon is available from target_feature
+            // SAFETY: neon is available from target_feature
             let (table_lo, table_hi) =
                 unsafe { (vld1q_f32(table.as_ptr()), vld1q_f32(table.as_ptr().add(4))) };
 
@@ -431,13 +431,9 @@ unsafe impl F32SimdVec for F32VecNeon {
             let table_lo_u32 = vreinterpretq_u32_f32(table_lo);
             let table_hi_u32 = vreinterpretq_u32_f32(table_hi);
 
-            // Shift right by 16 to get BF16 in low 16 bits
-            let bf16_lo_u32 = vshrq_n_u32::<16>(table_lo_u32);
-            let bf16_hi_u32 = vshrq_n_u32::<16>(table_hi_u32);
-
-            // Narrow to 16-bit (take low 16 bits of each 32-bit lane)
-            let bf16_lo_u16 = vmovn_u32(bf16_lo_u32);
-            let bf16_hi_u16 = vmovn_u32(bf16_hi_u32);
+            // Shift right by 16 AND narrow to 16-bit in one instruction
+            let bf16_lo_u16 = vshrn_n_u32::<16>(table_lo_u32);
+            let bf16_hi_u16 = vshrn_n_u32::<16>(table_hi_u32);
 
             // Combine into 8 x u16 = 16 bytes
             let bf16_table_u16 = vcombine_u16(bf16_lo_u16, bf16_hi_u16);
--- a/jxl_simd/src/x86_64/avx.rs
+++ b/jxl_simd/src/x86_64/avx.rs
@@ -469,7 +469,6 @@ unsafe impl F32SimdVec for F32VecAvx {
 
     #[inline(always)]
     fn table_lookup_bf16_8(d: AvxDescriptor, table: Bf16Table8Avx, indices: I32VecAvx) -> Self {
-        // For AVX2, use vpermps which is both fast and exact
         // SAFETY: avx2 is available from the safety invariant on the descriptor
         F32VecAvx(unsafe { _mm256_permutevar8x32_ps(table.0, indices.0) }, d)
     }
--- a/jxl_simd/src/x86_64/avx512.rs
+++ b/jxl_simd/src/x86_64/avx512.rs
@@ -523,14 +523,11 @@ unsafe impl F32SimdVec for F32VecAvx512 {
     fn prepare_table_bf16_8(_d: Avx512Descriptor, table: &[f32; 8]) -> Bf16Table8Avx512 {
         #[target_feature(enable = "avx512f")]
         #[inline]
-        unsafe fn prepare_impl(table: &[f32; 8]) -> __m512 {
-            // SAFETY: avx512f intrinsics are available from target_feature
-            unsafe {
-                // Load the 8-entry table into a 256-bit register
-                let table_256 = _mm256_loadu_ps(table.as_ptr());
-                // Duplicate to 512-bit for vpermutexvar_ps
-                _mm512_insertf32x8::<1>(_mm512_castps256_ps512(table_256), table_256)
-            }
+        fn prepare_impl(table: &[f32; 8]) -> __m512 {
+            // SAFETY: avx512f is available from target_feature
+            let table_256 = unsafe { _mm256_loadu_ps(table.as_ptr()) };
+            // Zero-extend to 512-bit; vpermutexvar with indices 0-7 only reads first 256 bits
+            _mm512_castps256_ps512(table_256)
         }
         // SAFETY: avx512f is available from the safety invariant on the descriptor
         Bf16Table8Avx512(unsafe { prepare_impl(table) })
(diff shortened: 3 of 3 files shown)


veluca93:
I don't think we should be using `gather` for this. Instead, we should work with shuffles.

The noise LUTs have 8 entries. If we keep them as BF16, that amounts to 16 bytes, which fit in a NEON / SSE4.2 register. We can then use shuffles with appropriate indexing (say vqtbl1q_u8 for NEON, _mm_shuffle_epi8 for sse) to reconstruct vectors of BF16s, and then convert the BF16s to F32 (by just zero extending).

Alternatively we could use F16 as a storage format and convert, or use F32 and a bigger table. Either way, a gather is not necessary (and almost certainly a shuffle-based approach will be faster).

Implementation wise, this could be done by having a function `F32Vec::prepare_table([f16; 8]) -> <something>`, and a function `U32Vec::table_lookup(indices, <something>)`, or something along those lines.

veluca93:
I think the table lookups should be implemented with proper SIMD instructions as I mentioned above, and we probably should have some way to handle approximate values.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -408,6 +415,196 @@ unsafe impl F32SimdVec for F32VecNeon {
...
+                // Result has bf16 in high 16 bits of each 32-bit lane = valid f32
+                vreinterpretq_f32_u8(result)
+            }
+        }
+        // SAFETY: neon is available from the safety invariant on the descriptor
+        F32VecNeon(unsafe { table_lookup_impl(table, indices.0) }, d)
+    }
+
+    #[inline(always)]
+    fn prepare_table_bf16_8(_d: NeonDescriptor, table: &[f32; 8]) -> Bf16Table8Neon {
```
veluca93, line 420: Can we only do prepare_table_bf16_8 + table_lookup_bf16_8? I am not sure we will need other variants.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -408,6 +415,196 @@ unsafe impl F32SimdVec for F32VecNeon {
...
+        #[target_feature(enable = "neon")]
+        #[inline]
+        unsafe fn prepare_impl(table: &[f32; 8]) -> uint8x16_t {
```
veluca93: This doesn't need to be an unsafe function, and the only unsafe operations in the block are the load operations.

jxl_simd/src/aarch64/neon.rs
```diff
@@ -408,6 +415,196 @@ unsafe impl F32SimdVec for F32VecNeon {
...
+
+    #[inline(always)]
+    fn table_lookup_bf16_8(d: NeonDescriptor, table: Bf16Table8Neon, indices: I32VecNeon) -> Self {
+        #[target_feature(enable = "neon")]
+        #[inline]
+        unsafe fn lookup_impl(bf16_table: uint8x16_t, indices: int32x4_t) -> float32x4_t {
+            // SAFETY: neon intrinsics are available from target_feature
+            unsafe {
+                // Convert indices to u8 (take low byte of each i32)
+                let indices_u32 = vreinterpretq_u32_s32(indices);
```
veluca93: I'm fairly convinced that this is not the most efficient implementation of the table lookup - instead, doing a simple byte-lookup with the equivalent of `((idx << 25) | (idx << 17) | 0x1008080)` should be significantly faster (or `((idx << 24) | (idx << 16) | 0x8008080)` depending on how the table gets constructed, i.e. `aabbccddeeffgghh` or `abcdefghabcdefgh`)

jxl_simd/src/x86_64/sse42.rs
```diff
@@ -374,6 +380,160 @@ unsafe impl F32SimdVec for F32VecSse42 {
...
+
+    #[inline(always)]
+    fn table_lookup_bf16_8(
+        d: Sse42Descriptor,
+        table: Bf16Table8Sse42,
+        indices: I32VecSse42,
+    ) -> Self {
+        #[target_feature(enable = "sse4.2")]
+        #[inline]
+        unsafe fn lookup_impl(bf16_table: __m128i, indices: __m128i) -> __m128 {
```
veluca93: Same considerations as the NEON code should apply.

jxl_simd/src/aarch64/neon.rs
hjanuschka: Refactored to use (i << 17) | (i << 25) | 0x01008080 approach which is significantly simpler

jxl_simd/src/aarch64/neon.rs
```diff
@@ -408,6 +415,67 @@ unsafe impl F32SimdVec for F32VecNeon {
...
+            // SAFETY: table has exactly 8 elements and neon is available from target_feature
+            let (table_lo, table_hi) =
+                unsafe { (vld1q_f32(table.as_ptr()), vld1q_f32(table.as_ptr().add(4))) };
+
+            // Reinterpret as u32 to extract high 16 bits
+            let table_lo_u32 = vreinterpretq_u32_f32(table_lo);
+            let table_hi_u32 = vreinterpretq_u32_f32(table_hi);
+
+            // Shift right by 16 to get BF16 in low 16 bits
+            let bf16_lo_u32 = vshrq_n_u32::<16>(table_lo_u32);
```
veluca93: Should probably use vshrn_n_u32, which shifts *and* narrows.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -453,6 +460,20 @@ unsafe impl F32SimdVec for F32VecAvx {
         I32VecAvx(_mm256_castps_si256(this.0), this.1)
     });
 
+    #[inline(always)]
+    fn prepare_table_bf16_8(_d: AvxDescriptor, table: &[f32; 8]) -> Bf16Table8Avx {
+        // For AVX2, vpermps is exact and fast, so we just load the table as-is
```
veluca93, line 465: No need to repeat this comment 3 times

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -511,6 +519,34 @@ unsafe impl F32SimdVec for F32VecAvx512 {
...
 
+    #[inline(always)]
+    fn prepare_table_bf16_8(_d: Avx512Descriptor, table: &[f32; 8]) -> Bf16Table8Avx512 {
+        #[target_feature(enable = "avx512f")]
+        #[inline]
+        unsafe fn prepare_impl(table: &[f32; 8]) -> __m512 {
+            // SAFETY: avx512f intrinsics are available from target_feature
+            unsafe {
+                // Load the 8-entry table into a 256-bit register
+                let table_256 = _mm256_loadu_ps(table.as_ptr());
```
veluca93: Only the load needs to be unsafe.

jxl_simd/src/x86_64/avx512.rs
```diff
@@ -511,6 +519,34 @@ unsafe impl F32SimdVec for F32VecAvx512 {
...
+                // Duplicate to 512-bit for vpermutexvar_ps
+                _mm512_insertf32x8::<1>(_mm512_castps256_ps512(table_256), table_256)
```
veluca93: Since permutexvar with indices from 0 to 7 only reads the first 256 bits, we don't *need* to actually duplicate the table, we might as well zero extend it.

hjanuschka: Done. Using _mm512_castps256_ps512 to zero-extend.

PR #588, opened 2025-12-23
Add --data-type CLI option and output format tests

hjanuschka:
## Summary

This PR addresses the feedback from PR #586 about being able to:
- Request output in different data types (u8/u16/f16) for benchmarking
- Add tests to catch output format conversion bugs in the future

### Changes

1. **CLI `--data-type` option** (`jxl_cli`)
   - New flag: `--data-type u8|u16|f16|f32` (default: f32)
   - Exercises the decoder's type conversion pipeline for benchmarking
   - Example: `jxl_cli image.jxl -s --data-type u8`

2. **Output format consistency tests** (`jxl/src/api/decoder.rs`)
   - `test_output_format_u8_matches_f32` - Verifies u8 output matches f32
   - `test_output_format_u16_matches_f32` - Verifies u16 output matches f32  
   - `test_output_format_f16_matches_f32` - Verifies f16 output matches f32
   - Tests both RGB and BGRA formats to catch channel reordering bugs
   - Tests both simple and low-memory pipelines
   - **Verified these tests would have caught the PR #586 bug** by temporarily reverting the fix and confirming test failure

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -637,7 +637,6 @@ pub(crate) mod tests {
     #[test]
     fn test_premultiply_output_straight_alpha() {
         use crate::api::{JxlColorType, JxlDataFormat, JxlPixelFormat};
-        use crate::image::{Image, Rect};
 
         // Use alpha_nonpremultiplied.jxl which has straight alpha (alpha_associated=false)
         let file =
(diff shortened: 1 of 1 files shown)


jxl_cli/src/dec/mod.rs
```diff
@@ -121,3 +123,198 @@ pub fn decode_frames<In: JxlBitstreamInput>(
...
+            })
+            .collect();
+
+        decoder_with_image_info = match decoder_with_frame_info.process(input, &mut output_bufs)? {
+            ProcessingResult::Complete { result } => result,
+            ProcessingResult::NeedsMoreInput { .. } => return Err(eyre!("Source file truncated")),
+        };
+
+        // Convert typed outputs to f32
+        let f32_outputs: Vec<Image<f32>> = typed_outputs
```
veluca93: As written, the time taken to convert to f32 would be included in the speed test, which is IMO not what we want.

I think it would be best to return an OwnedRawImage and have the caller convert that to f32 when passing the data to encoders.

jxl/src/api/decoder.rs
```diff
@@ -1017,4 +1017,246 @@ pub(crate) mod tests {
...
+                            use_simple
+                        );
+                    }
+                }
+            }
+        }
+    }
+
+    /// Helper function to decode an image with a specific format.
+    fn decode_with_format<T: crate::image::ImageDataType>(
```
veluca93, line 1053: Shouldn't we use or modify `decode_image` instead?

For that matter, I feel like we might have duplicated and inlined `decode_image` a few times.

hjanuschka: Consolidated the duplicate decode_image helpers into the existing decode_with_format<T>, adding a premultiply parameter.

PR #590, opened 2025-12-23
Add jbrd box support for JPEG reconstruction

hjanuschka:
## Summary
- Add bit-exact JPEG reconstruction from jbrd boxes, including Huffman tables, scans, marker order, and APP markers
- Preserve and pipe VarDCT quantized coefficients through decode to reconstruction output
- Use codestream quant tables, ICC data, and sampling factors when writing JPEG; handle subsampled component block grids
- Feature is behind the `jpeg-reconstruction` flag and is off by default

### DIFF
--- a/jxl/src/frame/group.rs
+++ b/jxl/src/frame/group.rs
@@ -589,42 +589,42 @@ pub fn decode_vardct_group(
 
             // Extract JPEG coefficients if requested (only for 8x8 DCT blocks)
             #[cfg(feature = "jpeg-reconstruction")]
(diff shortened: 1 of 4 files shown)


hjanuschka:
@veluca93 eventhough you dont want to merge it, it would be now feature complete, lets keep it around and mergeable in case you change your mind.

zone117x:
@hjanuschka @veluca93 are there any blockers on this PR? This is a useful feature that I would like to use. 

I'm happy to help contribute if needed, although it looks like @hjanuschka is pretty on top of things :)

PR #591, opened 2025-12-23
Add images from web-platform-tests.

veluca93:
This uncovered a bug in handling of recompressed 420 JPEGs, which is fixed.


jxl/src/render/internal.rs
```diff
@@ -143,11 +144,16 @@ impl<Buffer> RenderPipelineShared<Buffer> {
...
-                .min(self.input_size.1.next_multiple_of(8)),
+            (1 << self.log_group_size)
+                .min(self.input_size.0)
+                .shrc(downsample.0 as usize)
+                .next_multiple_of(16),
+            (1 << self.log_group_size)
+                .min(self.input_size.1)
+                .shrc(downsample.1 as usize)
+                .next_multiple_of(16),
         )
```
tirr-c, line 157: Does this work even when either direction is not subsampled? (Maybe it's okay because it's just extra unused padding...)

veluca93: Yup, exactly :-)

PR #598, opened 2025-12-27
Add faster decoding fast paths for single-node MA trees

### LINKED ISSUE #546: Lossless Faster Decoding Fastpaths
Currently jxl-rs lacks some important fast paths for modular decoding, which libjxl replies on for its Faster Decoding levels.

It would be a good idea to add a file from each FD level to make sure they _are_ actually faster.

Most notable are GradientOnly (Done) and SingleNode, with GradientOnly being the fastest MA type thanks to using a LUT and SingleNode being used for No Predictor (progressive lossless/LZ77), Effort 1 and Faster Decoding level 4.

hjanuschka:
Fixes #546

Adds specialized fast paths for modular decoding:

**Single-node fast paths** (skip tree traversal):
- `SingleZero`: No prediction, no neighbor access (for LZ77)
- `SingleGradient`: Gradient predictor with any multiplier/offset
- `SingleWest`: West/WestWest predictors (no top row needed)
- `SingleNodeTopOnly`: North, Select, NorthEast, etc.
- `SingleNodeFull`: AverageAll (needs toptop)

**LUT-based fast path for effort 2**:
- `GradientTableLookup`: For trees splitting only on property 9 with Gradient predictor
- Based on https://github.com/tirr-c/jxl-oxide/pull/306

### DIFF
--- a/jxl/src/frame/modular/decode/channel.rs
+++ b/jxl/src/frame/modular/decode/channel.rs
@@ -195,6 +195,9 @@ pub(super) fn decode_modular_channel(
         TreeSpecialCase::WpOnly(t) => {
             decode_modular_channel_impl(buffers, chan, t, reader, br, &tree.histograms)
         }
+        TreeSpecialCase::GradientTable(t) => {
+            decode_modular_channel_impl(buffers, chan, t, reader, br, &tree.histograms)
+        }
         TreeSpecialCase::SingleZero(t) => {
             decode_modular_channel_impl(buffers, chan, t, reader, br, &tree.histograms)
         }
--- a/jxl/src/frame/modular/decode/specialized_trees.rs
+++ b/jxl/src/frame/modular/decode/specialized_trees.rs
@@ -15,7 +15,7 @@ use crate::{
             channel::ModularChannelDecoder,
             common::{make_pixel, precompute_references},
         },
-        predict::{PredictionData, WeightedPredictorState},
+        predict::{PredictionData, WeightedPredictorState, clamped_gradient},
         tree::{
             FlatTreeNode, NUM_NONREF_PROPERTIES, PROPERTIES_PER_PREVCHAN, TreeNode, predict_flat,
         },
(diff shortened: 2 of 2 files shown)


jonnyawsom3:
Unfortunately when I opened the issue, I slightly misremembered how predictors and the MA tree interact. I was going to add more details to the issue but got caught up with the holidays, sorry about that.

For lossless, it's only a single node when the MA tree is disabled. This happens at effort 1 or faster decoding level 4, not implicitly LZ77/progressive lossless like I said before. Improvement from this PR seems within error, at 100reps FD4 was faster and effort 1 slower by less than 1%.

Effort 2 and effort 3 use fixed tables which can be turned into a LUT, @tirr-c has experience implementing them in https://github.com/tirr-c/jxl-oxide/pull/306 so they may have some advice.

 I'll try to submit some test files soon to bench. Probably all 4 faster decoding levels and a Progressive Lossless image.

jonnyawsom3:
Ran some quick benchmarks, numbers are  MP/s and percentage is compared to libjxl.
| Encoding | libjxl | jxl-rs | With PR |
|-----------|-------|-------|----------|
| Effort 2 | 18.246 |  5.768 (-68.4%) | 16.603 (-9.0%) |
| Effort 3 | 6.989 |  6.521 (-7.17%) | 6.689 (-4.48%) |
| Faster Decoding 3 | 18.304 |  6.532 (-64.31%) | 16.603 (-9.29%) |

I also tried Effort 1, Faster Decoding 1, 2 and 4 but they had no change.

veluca93:
> Ran some quick benchmarks, numbers are MP/s and percentage is compared to libjxl.
> 
> Encoding	libjxl	jxl-rs	With PR
> Effort 2	18.246	5.768 (-68.4%)	16.603 (-9.0%)
> Effort 3	6.989	6.521 (-7.17%)	6.689 (-4.48%)
> Faster Decoding 3	18.304	6.532 (-64.31%)	16.603 (-9.29%)
> I also tried Effort 1, Faster Decoding 1, 2 and 4 but they had no change.

Given the speedups, I would for now limit this to gradient predictor lookups (as is also implemented in #600 AFAICT)

hjanuschka:
Closing in favor of #600 which provides a more general SinglePropertyLookup implementation that covers this use case.

Issue #513, opened 2025-12-02
Feature request: JPEG decoding


o-l-a-v:
Got the idea from:

* 2023-04-19 jxl-oxide feature request: <https://github.com/tirr-c/jxl-oxide/issues/4>

For more context: <https://github.com/mozilla/standards-positions/issues/522#issuecomment-856035731>

> Regarding maintenance / code size / attack surface / binary size, I want to point out that potentially a JPEG XL decoder could also be used as a JPEG decoder (libjxl currently doesn't do that yet, but it does contain all the code needed to do that), which means the dependency that browsers have on libjpeg-turbo could potentially be dropped.
If the numbers from android chrome are an indication (see <https://chromium-review.googlesource.com/c/chromium/src/+/2932498>): libjpeg_turbo weighs in at 151 KiB, libjxl currently weighs in at 191 KiB, but plausibly libjpeg-turbo could be dropped and the jxl decoder could be used instead also for jpeg decoding, which brings the net binary size overhead down to ~40 KiB. So it does come at a cost in maintenance / code size / attack surface / binary size, but arguably the net cost is not huge, e.g. compared to libwebp (165 KiB), and probably quite comparable to the net cost of AVIF (i.e. implementing the HEIF functionality on top of an AV1 decoder that is and presumably will remain 'free' in the sense of 'available anyway').

TPS:
Is this a request to port in [`jpegli`](https://github.com/google/jpegli) for _its_ decoding, like used to be in [`libjxl`](https://github.com/libjxl/libjxl)?  Or some other implementation?

CrushedAsian255:
I think the idea is more to have jxl-rs also be able to decode baseline JPEGs

TPS:
Yes, of course.  But _why_ should _this_ crate do it?  There're already many.  Imho, the only sensible reason to have 1 _here_ is b/c [`libjxl` used to](https://github.com/libjxl/jxl-rs/issues/513#issuecomment-4147785165), which originated `jpegli`, so [ifF](wikipedia.org/wiki/If_and_only_if) anything, that's what should be here, though there's [an independent implementation for rust](https://github.com/Szpadel/jpegli-rs) already.

CrushedAsian255:
> But _why_ should _this_ crate do it? 

The main reason would be so that any software that wants to support JPEG XL can use the same code path with JPEG and JPEG XL

TPS:
[...]

PR #599, opened 2025-12-27
jpeg: decoder parity stack (post-#590)

hjanuschka:
Depends on #590.

- Adds standalone JPEG decoder + regression fixtures
- Progressive scan handling
- SIMD YCbCr/BGRA and IDCT parity
- Libjpeg-turbo parity references (scaled + 12-bit)
- Chromium integration sample under samples/chromium_integration

Tests:
- cargo test -p jxl --features jpeg-reconstruction test_libjpeg_turbo_ -- --nocapture

Notes:
- Idea related to https://github.com/libjxl/jxl-rs/issues/513
- This PR is exploratory; not yet sure whether it should be kept or killed, might be a libjpeg-turbo replacement in chromium some day, or just a closed pr in the endless space!


goodusername123:
while interesting I do feel like this is a bit out of scope for jxl-rs for a few reasons:
1. JXL only supports a subset of the de-facto JPEG spec/standard (some examples being no chroma sub-sampling modes besides 4:4:4 4:2:0 4:2:2 and no support for RGB outside of 4:4:4) so having a JPEG decoder that supports more of JPEG then JXL does feels a bit weird.
2. Matching the decoded output of libjpeg-turbo (or any other JPEG decoder for that matter) is a non-goal. For example [libjxl](https://github.com/libjxl/libjxl) decodes in a way that is compliant with how the JPEG spec defines images get decoded but different from what libjpeg-turbo outputs since the JPEG spec has/gives a bit of leeway as to how a decoder should produce pixels.

jonnyawsom3:
See https://github.com/libjxl/libjxl/issues/455 and https://github.com/libjxl/libjxl/issues/1470#issuecomment-1148701920

As Username said, while theoretically possible, I'd say it's out of scope for now. There's been many discussions about it in the past, but other things take priority. jpegli already exists as a replacement JPEG library based on libjxl

hjanuschka:
ok thank you folks! lets close it!

PR #600, opened 2025-12-29
Add SinglePropertyLookup for table-based decision routing

hjanuschka:
Generalizes WpOnlyLookup: when a tree only splits on a single non-WP property with all leaves having the same predictor (offset=0, multiplier=1), builds a LUT from property values directly to cluster IDs.

Benefits effort 2/3 encoded images where trees use fixed property-based routing. Based on jxl-oxide's SimpleMaTable approach (https://github.com/tirr-c/jxl-oxide/pull/306).

Note: This should be rebased after #598 merges to ensure proper detection order (more specific fast paths checked first).

### DIFF
--- a/jxl/src/frame/modular/decode/specialized_trees.rs
+++ b/jxl/src/frame/modular/decode/specialized_trees.rs
@@ -250,168 +250,69 @@ impl ModularChannelDecoder for WpOnlyLookup {
     }
 }
 
-/// Fast path for trees that split only on a single property (not WP property 15).
-/// All leaves must have the same predictor with offset=0, multiplier=1.
-/// Maps property values directly to cluster IDs via LUT.
-pub struct SinglePropertyLookup {
-    lut: Box<[u8]>,
-    property: u8,
-    predictor: Predictor,
-    value_base: i32,
+/// Fast path for trees that split only on property 9 (gradient: left + top - topleft)
+/// with Gradient predictor, offset=0, multiplier=1.
+/// Maps property 9 values directly to cluster IDs via a LUT.
+/// This targets libjxl effort 2 encoding.
+pub struct GradientLookup {
+    lut: [u8; LUT_TABLE_SIZE],
 }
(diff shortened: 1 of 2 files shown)


jxl/src/frame/modular/decode/specialized_trees.rs
```diff
@@ -250,6 +250,174 @@ impl ModularChannelDecoder for WpOnlyLookup {
     }
 }
 
+/// Fast path for trees that split only on a single property (not WP property 15).
+/// All leaves must have the same predictor with offset=0, multiplier=1.
+/// Maps property values directly to cluster IDs via LUT.
+pub struct SinglePropertyLookup {
+    lut: Box<[u8]>,
```
veluca93: Why Box?

jxl/src/frame/modular/decode/specialized_trees.rs
```diff
@@ -250,6 +250,174 @@ impl ModularChannelDecoder for WpOnlyLookup {
...
+                    _ => {}
+                }
+            }
+        }
+    }
+
+    let property = split_property?;
+    let predictor = leaf_predictor?;
+
+    // Second pass: find the range of property values
```
veluca93: This could be replaced by make_lut, AFAICT.

jxl/src/frame/modular/decode/specialized_trees.rs
```diff
@@ -250,6 +250,174 @@ impl ModularChannelDecoder for WpOnlyLookup {
     }
 }
 
+/// Fast path for trees that split only on a single property (not WP property 15).
```
veluca93: I am not sure there is much of a point for having all the possible properties and predictors, and doing so has a nontrivial speed cost. I think for now we should limit this to the gradient-only case, which is the only case other than WP that is produced somewhat commonly.

veluca93 (approved):

PR #601, opened 2025-12-29
Add i16 sample type support for modular decoding

hjanuschka:
Add i16 sample type support for modular decoding, enabling 50% memory bandwidth reduction for 8-16 bit images.

Based on jxl-oxide's sealed trait pattern.

## Changes

**Infrastructure:**
- `ModularSample` trait with wrapping arithmetic and i64 conversion
- `ModularChannelGeneric<T>` with `ModularChannel` alias for i32
- `PredictionDataGeneric<T>` with `PredictionData` alias for i32  
- `PredictionDataI64` for intermediate calculations

**Decode pipeline:**
- `make_pixel<T>` generic over sample type
- All decoders updated to use `make_pixel::<i32>`
- `ModularSample::from_i64`: direct cast for i32, saturating for i16

All 481 tests pass.

## Design Notes

**Transforms stay i32:** RCT, squeeze, and palette transforms use SIMD operations (`I32SimdVec`) for performance. Keeping them i32-only is intentional - the decode loop (many pixels) benefits from i16, while transforms (applied once per region) benefit from SIMD optimization.

**To use i16 samples:** Pass `i16` as the type parameter to `ModularChannelGeneric` and `make_pixel`. For channels without transforms, this provides 50% memory bandwidth reduction. For channels with transforms, use i32.

### DIFF
--- a/jxl/src/frame/modular/borrowed_buffers.rs
+++ b/jxl/src/frame/modular/borrowed_buffers.rs
@@ -12,7 +12,7 @@ use crate::{
     util::AtomicRefMut,
 };
 
-use super::{ModularBufferInfo, ModularChannel};
+use super::{ModularBufferInfo, ModularChannel, ModularChannelI16};
 
 pub fn with_buffers<T>(
     buffers: &[ModularBufferInfo],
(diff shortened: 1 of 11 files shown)


veluca93:
Let's do this later. We would want to implement 16-bit SIMD transforms too, and 16-bit buffers; the benefits of just using 16 bits for the entropy decoding buffers are minimal.

veluca93:
Closing this in favour of #928

PR #602, opened 2025-12-29
security: fix integer overflow vulnerabilities

lilith:
## Summary

This PR fixes several integer overflow vulnerabilities that could be exploited by malicious JXL files:

- **patches.rs**: Use `checked_mul` for `xsize*ysize` to prevent overflow when calculating pixel counts. Use `saturating_*` for limit calculations (with `try_reserve` as the hard limit).
- **spline.rs**: Use `saturating_mul/add` with `.min(1<<42)` cap for area limit calculations.
- **size.rs**: Add `map_aspect_ratio_checked` using `u64` intermediate + `try_from` to prevent overflow when calculating xsize from ysize and aspect ratio. Returns `u32::MAX` on overflow to trigger downstream validation.
- **icc/tag.rs**: Use `checked_add/mul` for ICC tag offset calculations to prevent overflow attacks in malformed ICC profiles.
- **render/builder.rs**: Use `checked_mul` for group count calculation, return `Result<Self>` instead of panicking.

All overflow conditions now return proper `Error` variants (`ImageSizeTooLarge`, `InvalidIccStream`) that propagate to the public API.

## Test plan

- [x] Added unit tests for aspect ratio overflow handling in `size.rs`
- [x] Existing test suite passes
- [x] Manual review of all overflow paths to ensure they return errors (not panic)

### DIFF
--- a/jxl/src/features/patches.rs
+++ b/jxl/src/features/patches.rs
@@ -369,10 +369,13 @@ impl PatchesDictionary {
             br,
             PatchContext::NumRefPatch as usize,
         ) as usize;
-        let num_pixels = xsize * ysize;
-        let max_ref_patches = 1024 + num_pixels / 4;
-        let max_patches = max_ref_patches * 4;
-        let max_blending_infos = max_patches * 4;
+        // Use checked arithmetic to prevent overflow attacks
+        let num_pixels = xsize
+            .checked_mul(ysize)
+            .ok_or(Error::ImageSizeTooLarge(xsize, ysize))?;
+        let max_ref_patches = 1024usize.saturating_add(num_pixels / 4);
+        let max_patches = max_ref_patches.saturating_mul(4);
+        let max_blending_infos = max_patches.saturating_mul(4);
         if num_ref_patch > max_ref_patches {
             return Err(Error::PatchesTooMany(
                 "reference patches".to_string(),
@@ -474,7 +477,8 @@ impl PatchesDictionary {
                 next_size *= 2;
                 next_size = std::cmp::min(next_size, max_patches);
             }
-            if next_size * blendings_stride > max_blending_infos {
+            // Use saturating_mul to prevent overflow attack bypassing the limit check
+            if next_size.saturating_mul(blendings_stride) > max_blending_infos {
                 return Err(Error::PatchesTooMany(
                     "blending_info".to_string(),
                     total_patches,
@@ -483,7 +487,8 @@ impl PatchesDictionary {
             }
             positions.try_reserve(next_size.saturating_sub(positions.len()))?;
             blendings.try_reserve(
-                (next_size * PatchBlendMode::NUM_BLEND_MODES as usize)
+                next_size
+                    .saturating_mul(PatchBlendMode::NUM_BLEND_MODES as usize)
                     .saturating_sub(blendings.len()),
             )?;
 
--- a/jxl/src/headers/size.rs
+++ b/jxl/src/headers/size.rs
@@ -58,16 +58,31 @@ pub struct Preview {
     xsize: Option<u32>,
 }
 
+/// Maps ysize to xsize based on aspect ratio.
+/// Returns None if the calculation would overflow u32.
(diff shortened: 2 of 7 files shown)


jxl/src/features/patches.rs
```diff
@@ -369,10 +369,13 @@ impl PatchesDictionary {
             br,
             PatchContext::NumRefPatch as usize,
         ) as usize;
-        let num_pixels = xsize * ysize;
-        let max_ref_patches = 1024 + num_pixels / 4;
-        let max_patches = max_ref_patches * 4;
-        let max_blending_infos = max_patches * 4;
+        // Use checked arithmetic to prevent overflow attacks
```
veluca93, line 372: Please remove these comments.

jxl/src/features/patches.rs
```diff
@@ -369,10 +369,13 @@ impl PatchesDictionary {
...
+        let num_pixels = xsize
```
veluca93, line 373: I don't think this is the place to trigger this error. Either we should unconditionally fail if `xsize * ysize` overflows usize when decoding a frame, or we should convert this arithmetic to saturating aritmetic.

jxl/src/headers/size.rs
```diff
@@ -58,16 +58,31 @@ pub struct Preview {
...
+        AspectRatio::Ratio4Over3 => ysize as u64 * 4 / 3,
+        AspectRatio::Ratio3Over2 => ysize as u64 * 3 / 2,
+        AspectRatio::Ratio16Over9 => ysize as u64 * 16 / 9,
+        AspectRatio::Ratio5Over4 => ysize as u64 * 5 / 4,
+        AspectRatio::Ratio2Over1 => ysize as u64 * 2,
+    };
+    u32::try_from(result).ok()
+}
+
 fn map_aspect_ratio<T: Fn() -> u32>(ysize: u32, ratio: AspectRatio, fallback: T) -> u32 {
```
veluca93, line 77: I would strongly prefer to instead just return a `u64` here. In particular, I strongly dislike using u32::MAX as a special value.

jxl/src/render/builder.rs
```diff
@@ -27,14 +27,18 @@ impl<Pipeline: RenderPipeline> RenderPipelineBuilder<Pipeline> {
...
-    ) -> Self {
+    ) -> Result<Self> {
         info!("creating render pipeline");
         assert!(chunk_size <= u16::MAX as usize);
         assert_ne!(chunk_size, 0);
         // The number of pixels that a group encompasses in the final, upsampled image along one
         // dimension is effectively multiplied by the upsampling factor.
         log_group_size += downsampling_shift;
-        Self {
+        // Use checked_mul to prevent overflow with malicious image dimensions
```
veluca93, line 37: This code should never overflow if we force xsize*ysize to fit in a usize.

lilith:
This PR should probably be split up - I think the ICC fix is correct, but the approach re: failure points with image dimension overflow needs maintainer-driven strategy. 

I usually aim for propagating overflow error failures more liberally up the chain. There's always one more function that needs to calculate w * h + alignment offset + stride depth * channels, plus C, where C is the new something that pushes us over the 32-bit usize limit. usize is always going to be too small sometimes.  My use case is in server image proxies, so crashing is never an option; malicious images can't cause panics.

veluca93:
> This PR should probably be split up - I think the ICC fix is correct, but the approach re: failure points with image dimension overflow needs maintainer-driven strategy.
> 
> I usually aim for propagating overflow error failures more liberally up the chain. There's always one more function that needs to calculate w * h + alignment offset + stride depth * channels, plus C, where C is the new something that pushes us over the 32-bit usize limit. usize is always going to be too small sometimes. My use case is in server image proxies, so crashing is never an option; malicious images can't cause panics.

I would prefer to consider that a library bug and adjust the image dimension limits accordingly, so that images that are outside of size limits are detected immediately.
Also keep in mind that no allocated object can ever be bigger than a `isize` in Rust, so the calculation overflowing a `usize` is not going to be sufficient to detect crashes.

lilith:
Handing this off - ICC overflow was merged.

PR #603, opened 2025-12-29
fix: use wrapping arithmetic for RCT transforms to prevent overflow panic

lilith:
## Summary

RCT (Reversible Color Transform) operations in modular decoding perform integer arithmetic that can overflow with certain input values. In debug mode, this causes a panic. In release mode, the behavior is undefined (though typically wraps on most platforms).

This PR:
- Adds `wrapping_add` and `wrapping_sub` methods to the `I32SimdVec` trait
- Implements these for all SIMD backends (scalar, SSE4.2, AVX, AVX-512, NEON)
- Uses wrapping arithmetic in RCT operations to ensure deterministic behavior matching libjxl

For SIMD types, integer addition/subtraction already wraps (this is inherent to SIMD integer ops), so the new methods simply delegate to `+`/`-`. For scalar `i32`, they call the standard library's `wrapping_add`/`wrapping_sub`.

## Test plan

- [x] Added unit tests for RCT overflow handling
- [x] Added unit tests verifying correct RCT operation results
- [x] Existing test suite passes

### DIFF
--- a/jxl_simd/src/scalar.rs
+++ b/jxl_simd/src/scalar.rs
@@ -4,6 +4,7 @@
 // license that can be found in the LICENSE file.
 
 use std::mem::MaybeUninit;
+use std::num::Wrapping;
 
 use crate::{U32SimdVec, impl_f32_array_interface};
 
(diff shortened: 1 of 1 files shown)


lilith:
Updated to use `std::num::Wrapping<i32>` as the scalar `I32Vec` type instead of adding trait methods. Is this what you had in mind?

This makes `+`/`-`/`*` operators wrap automatically for the scalar path, matching SIMD hardware behavior.

veluca93:
Can we do this with u32 too for consistency? Also, can you update the PR description?

veluca93:
Done in #632

PR #604, opened 2025-12-29
fix: use extra channel's own bit_depth for modular-to-f32 conversion

lilith:
## Summary

Extra channels (alpha, depth, etc.) can have different bit depths than the main color channels. For example, an image might be 8-bit RGB with a 16-bit alpha channel.

Previously, the render pipeline used the image's global `metadata.bit_depth` for converting all extra channels from modular to f32. This caused incorrect conversion for extra channels with different bit depths.

This PR:
- Adds a `bit_depth()` getter to `ExtraChannelInfo`
- Uses each extra channel's own bit depth in `ConvertModularToF32Stage`

## Test plan

- [x] Added unit tests for extra channel bit depth accessor
- [x] All 29 extra channel and alpha-related tests pass
- [x] Note: 3 pre-existing test failures on main (RCT overflow) are unrelated and will be fixed by #603

### DIFF
--- a/jxl/src/frame/render.rs
+++ b/jxl/src/frame/render.rs
@@ -237,8 +237,8 @@ impl Frame {
             }
         }
         for i in 3..num_channels {
-            pipeline =
-                pipeline.add_inout_stage(ConvertModularToF32Stage::new(i, metadata.bit_depth))?;
+            let ec_bit_depth = metadata.extra_channel_info[i - 3].bit_depth();
+            pipeline = pipeline.add_inout_stage(ConvertModularToF32Stage::new(i, ec_bit_depth))?;
         }
 
         for c in 0..3 {
--- a/jxl/src/render/stages/convert.rs
+++ b/jxl/src/render/stages/convert.rs
@@ -448,4 +448,89 @@ mod test {
     fn f32_to_f16_consistency() -> Result<()> {
         crate::render::test::test_stage_consistency(|| ConvertF32ToF16Stage::new(0), (500, 500), 1)
     }
+
+    /// Test that modular-to-f32 conversion scale depends on bit depth.
+    ///
+    /// This test verifies the core math that ConvertModularToF32Stage uses:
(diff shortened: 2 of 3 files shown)


lilith:
## ⚠️ Test Limitation

**No true failing integration test.** A proper regression test would require a JXL file where the image bit depth differs from an extra channel bit depth. Without such a test asset, we can only test the components in isolation.

The added tests demonstrate the *math* of the bug (using wrong scale produces 257x error), but do not exercise the specific buggy line in `frame/render.rs` end-to-end.

---

**Changes in latest push:**
- Added `test_modular_to_f32_bit_depth_matters` showing 257x error from wrong scale
- Added `modular_to_f32_8bit_consistency` and `modular_to_f32_16bit_consistency` stage tests
- Removed Claude co-author attribution

jxl/src/frame/render.rs
```diff
@@ -237,8 +237,9 @@ impl Frame {
             }
         }
         for i in 3..num_channels {
-            pipeline =
-                pipeline.add_inout_stage(ConvertModularToF32Stage::new(i, metadata.bit_depth))?;
+            // Use each extra channel's own bit depth, not the image's metadata bit depth
```
veluca93: Please remove this comment.

veluca93 (approved):

jxl/src/render/stages/convert.rs
```diff
@@ -448,4 +448,89 @@ mod test {
...
+        let ratio = results_8bit[2] / results_16bit[2];
+        assert!(
+            ratio > 250.0,
+            "Using wrong bit depth causes ~257x error, ratio was {}",
+            ratio
+        );
+    }
+
+    /// Test ConvertModularToF32Stage consistency with different bit depths.
+    #[test]
```
veluca93, line 519: I am not sure I see the point of all the other tests, except these last two.
Do you mind removing them?

PR #611, opened 2026-01-09
Update CLI speedtest formatting

jonnyawsom3:
Makes `-s` output MP/s instead of pixels per second.
Limits seconds and MP/s to 3 decimal places matching libjxl.
Directly requests seconds in f64 instead of dividing nanoseconds.
Author file has been sorted while I added my name to it.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,17 +16,18 @@ Cloudinary Ltd. <*@cloudinary.com>
 Google LLC <*@google.com>
 
 # Individuals:
-Luca Versari
+Caio <c410.f3r@gmail.com>
 Ewout ter Hoeven <E.M.terHoeven@student.tudelft.nl>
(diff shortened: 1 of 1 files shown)


jonnyawsom3:
The parsing code would need to be updated after merging https://github.com/zond/jxl-perfhistory/blob/main/src/main.rs#L654

veluca93 (approved):

jonnyawsom3:
@zond could you update the bench tool to take the new MP/s readout into account? Thanks in advance.

zond:
> @zond could you update the bench tool to take the new MP/s readout into account? Thanks in advance.

https://github.com/zond/jxl-perfhistory/commit/bbd5e8ac7a820001cb119a79cafe78bbb3241d9b

:+1:

PR #612, opened 2026-01-09
fix last_pass validation

### LINKED ISSUE #609: last_pass validation rejects valid progressive files
## Bug Description

The validation for `Passes::last_pass` in `FrameHeader::check()` has inverted logic. It rejects valid strictly-increasing sequences and accepts invalid ones.

## Reproduction

Decoding [`progressive_ac.jxl`](https://github.com/imazen/codec-corpus/blob/main/jxl/features/progressive_ac.jxl) fails with:
```
Error: PassesLastPassNonIncreasing
```

This file decodes successfully with `djxl`.

## Root Cause

[`frame_header.rs:702-706`](https://github.com/libjxl/jxl-rs/blob/main/jxl/src/headers/frame_header.rs#L702-L706):
```rust
for w in self.passes.last_pass.windows(2) {
    let [last_lp, lp] = w else { unreachable!() };
    if lp >= last_lp {  // BUG: rejects increasing sequences
        return Err(Error::PassesLastPassNonIncreasing);
    }
}
```

libjxl requires `last_pass` to be **strictly increasing**, but this check rejects increasing sequences (e.g., `[0, 1, 2]` fails because `1 >= 0`).

## Fix

```rust
for w in self.passes.last_pass.windows(2) {
    let [prev_lp, lp] = w else { unreachable!() };
    if lp <= prev_lp {  // Correct: reject non-increasing
        return Err(Error::PassesLastPassNonIncreasing);
    }
}
```

mo271:
fixes #609

### DIFF
--- a/jxl/src/headers/frame_header.rs
+++ b/jxl/src/headers/frame_header.rs
@@ -701,7 +701,7 @@ impl FrameHeader {
 
         for w in self.passes.last_pass.windows(2) {
             let [last_lp, lp] = w else { unreachable!() };
-            if lp >= last_lp {
+            if lp <= last_lp {
                 return Err(Error::PassesLastPassNonIncreasing);
             }
         }


mo271:
first checking that the CI fails with the new test image...

PR #613, opened 2026-01-09
Make as_icc() return Result instead of panicking

hjanuschka:
Malformed color encodings can cause maybe_create_profile() to fail, crashing via unwrap().

Found by Chromium's fuzzer (crbug.com/474417800).

Breaking change: as_icc() now returns Result.

### DIFF
--- a/jxl/src/api/color.rs
+++ b/jxl/src/api/color.rs
@@ -1113,15 +1113,17 @@ pub enum JxlColorProfile {
 }
 
 impl JxlColorProfile {
-    /// Returns the ICC profile, panicking if unavailable.
+    /// Returns the ICC profile, or an error if creation fails.
     ///
-    /// # Panics
-    /// Panics if the color encoding cannot generate an ICC profile.
-    /// Consider using `try_as_icc` for fallible conversion.
-    pub fn as_icc(&self) -> Cow<'_, Vec<u8>> {
+    /// For color encodings that cannot generate ICC profiles (e.g., malformed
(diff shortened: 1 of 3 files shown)


jxl/src/api/color.rs
```diff
@@ -1113,30 +1113,25 @@ pub enum JxlColorProfile {
...
-        match self {
-            Self::Icc(x) => Some(Cow::Borrowed(x)),
-            Self::Simple(encoding) => encoding
-                .maybe_create_profile()
-                .ok()
-                .flatten()
-                .map(Cow::Owned),
-        }
+        self.as_icc().ok()
     }
```
tirr-c, line 1135: How about making `try_as_icc` return `Result`, and just remove `as_icc`? We can do `try_as_icc().ok()` if we need an `Option`.

jxl/src/error.rs
```diff
@@ -253,6 +253,8 @@ pub enum Error {
     MatrixInversionFailed(f64),
     #[error("Unsupported transfer function when writing ICC")]
     IccUnsupportedTransferFunction,
+    #[error("ICC profile creation returned None")]
```
tirr-c, line 256: Something like "Cannot create ICC profile from color encoding" would be better I think?

veluca93:
I think for now it's ok to leave try_as_icc() as is, and we can create more
detailed errors later - I am not sure there's much value in creating an
error variant that gives no extra information.

PR #617, opened 2026-01-13
Validate blending alpha channels

hjanuschka:
Fixes https://issues.chromium.org/issues/475333717

ClusterFuzz: https://clusterfuzz.com/testcase?key=6033369720619008

Validate blending alpha channel indices against the extra-channel count before rendering.

### DIFF
--- a/jxl/src/headers/frame_header.rs
+++ b/jxl/src/headers/frame_header.rs
@@ -682,16 +682,10 @@ impl FrameHeader {
         }
 
         let num_extra_channels = nonserialized.extra_channel_info.len();
-        let uses_alpha = |mode| {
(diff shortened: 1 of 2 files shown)


jxl/src/headers/frame_header.rs
```diff
@@ -681,6 +681,39 @@ impl FrameHeader {
             ));
         }
 
+        let num_extra_channels = nonserialized.extra_channel_info.len();
+        let uses_alpha =
+            |mode| matches!(mode, BlendingMode::Blend | BlendingMode::AlphaWeightedAdd);
+
+        if self.ec_blending_info.len() != num_extra_channels {
```
veluca93: This can never happen, because exactly `nonserialized.extra_channel_info.len()` entries get deserialized. Let's remove the error variant and the check (or make it an assert)

PR #620, opened 2026-01-14
Forced an optimization boundary via #[inline(never)] wrapper

zond:


tirr-c:
I wonder if `std::hint::black_box` could do the similar thing...

zond:
> I wonder if `std::hint::black_box` could do the similar thing...

It didn't, _but_ an #[inline(never)]-wrapper did!

jxl/src/render/stages/cms.rs
```diff
@@ -8,6 +8,12 @@ use std::any::Any;
 use crate::api::JxlCmsTransformer;
 use crate::error::Result;
 use crate::render::RenderPipelineInPlaceStage;
+
+/// Wrapper to prevent LLVM from optimizing across this call.
+#[inline(never)]
+fn call_cms_transform(transformer: &mut dyn JxlCmsTransformer, buf: &mut [f32]) -> bool {
+    transformer.do_transform_inplace(buf).is_ok()
+}
```
tirr-c: Typically function items come after `use` lines, it would be better to move this below.

jxl/src/render/stages/cms.rs
```diff
@@ -28,6 +29,12 @@ pub struct CmsStage {
     buffer_size: usize,
 }
 
+/// Wrapper to prevent LLVM from optimizing across this call.
```
veluca93: I think this is a brittle solution that will likely break for unpredictable reasons, but I guess it's simple enough to go for it for now.

However, let's leave a long comment explaining what is going on here.

zond: Yup. Added a long comment.

PR #621, opened 2026-01-14
Added CMS stage support for CMYK

zond:


jxl/src/render/stages/cms.rs
```diff
@@ -53,26 +63,69 @@ pub struct CmsStage {
...
+fn call_cms_transform_inplace(
+    transformer: &mut dyn JxlCmsTransformer,
+    buf: &mut [f32],
+) -> Option<()> {
     transformer.do_transform_inplace(buf).ok()
 }
 
+/// Wrapper for separate-buffer transform to prevent LLVM from pessimizing unrelated code paths.
+///
+/// See module-level documentation for details on why this wrapper is needed.
```
veluca93: It's not module-level, but the documentation of call_cms_transform_inplace.

jxl/src/render/stages/cms.rs
```diff
@@ -130,52 +199,76 @@ impl RenderPipelineInPlaceStage for CmsStage {
...
-        match nc {
-            2 if row.len() >= 2 => {
+        // Index into row using channel indices to support non-contiguous channels
+        match in_nc {
+            2 => {
                 interleave_2_dispatch(
-                    &row[0][..xsize_padded],
-                    &row[1][..xsize_padded],
-                    &mut state.buffer[..xsize_padded * 2],
+                    &row[indices[0]][..xsize_padded],
```
veluca93: This is incorrect: the stage only gets a slice with the channels it actually uses, so they are always in indices 0, 1 and 2.

jxl/src/render/stages/cms.rs
```diff
@@ -18,15 +18,25 @@ use crate::util::AtomicRefCell;
...
+///
+/// Supports non-contiguous channel indices, e.g., CMYK with K at a non-adjacent index.
 pub struct CmsStage {
     transformers: Vec<AtomicRefCell<Box<dyn JxlCmsTransformer + Send>>>,
-    first_channel: usize,
-    num_channels: usize,
-    buffer_size: usize,
+    /// Indices of input channels to read from (can be non-contiguous).
+    /// E.g., [0, 1, 2] for RGB or [0, 1, 2, 5] for CMYK where K is at index 5.
+    in_channel_indices: Vec<usize>,
```
veluca93: Since color channels are always 0,1,2, why don't we just add a `black_channel: Option<usize>`?

jxl/src/render/stages/cms.rs

jxl/src/render/stages/cms.rs

jxl/src/render/stages/cms.rs

PR #625, opened 2026-01-15
Fix color transform and CMS handling to match libjxl

zond:
CMS channel count (crash fix):
- Use cms_input.is_cmyk() not black_channel.is_some() for input
  channels. Matches c_src_.IsCMYK() (stage_cms.cc:42)

Channel compatibility:
- Skip CMS for grayscale<->RGB, allow CMYK->RGB. Matches
  not_mixing_color_and_grey (stage_cms.cc:44-46)

XYB CMS input:
- Use linear input profile for XYB. Matches linear_color_encoding
  usage (stage_cms.cc:35)

FromLinear/CMS exclusion:
- Use CMS or FromLinearStage, not both. Matches dec_cache.cc logic

color_encoding_is_original (performance fix):
- Compare original input profile, not linearized cms_input_profile.
  Matches dec_xyb.cc:184 orig_color_encoding.SameColorEncoding(c_desired)

NPY output (conformance test fix):
- Don't request linear output for npy, use original encoding like
  djxl (only EXR forces RGB_D65_SRG_Rel_Lin)
- Remove [0,1] clamping in numpy encoder. Matches libjxl npy.cc
  which writes raw float values

API changes:
- Rename can_convert_internally to same_color_encoding, include TF
  comparison. Matches libjxl SameColorEncoding
- Add channels(), is_cmyk(): parse ICC header for color space
- Add with_linear_tf(), transfer_function(), is_linear()

Cleanup:
- Remove xyb_output_linear, now implicit in CMS/FromLinear logic
- Add LCMS2 CMS backend in jxl_cli

### DIFF
--- a/jxl_cli/benches/decode.rs
+++ b/jxl_cli/benches/decode.rs
@@ -5,7 +5,7 @@
 
 use criterion::{BenchmarkId, Criterion, SamplingMode, criterion_group, criterion_main};
 use jxl::api::JxlDecoderOptions;
-use jxl_cli::dec::{LinearOutput, OutputDataType, decode_frames_with_type, decode_header};
+use jxl_cli::dec::{OutputDataType, decode_frames_with_type, decode_header};
 use std::fs;
 use std::path::{Path, PathBuf};
 
@@ -56,7 +56,7 @@ fn decode_benches(c: &mut Criterion) {
                         &mut input,
                         JxlDecoderOptions::default(),
                         OutputDataType::F32,
-                        LinearOutput::No,
+                        false,
                     )
                     .unwrap();
                 })
--- a/jxl_cli/src/lib.rs
+++ b/jxl_cli/src/lib.rs
@@ -9,7 +9,7 @@ pub mod enc;
 
 #[cfg(test)]
 mod tests {
-    use crate::dec::{LinearOutput, OutputDataType, decode_frames_with_type};
+    use crate::dec::{OutputDataType, decode_frames_with_type};
     use jxl::api::JxlDecoderOptions;
 
     /// Test that decoding with all output data types produces consistent results.
@@ -34,7 +34,7 @@ mod tests {
                 &mut input,
                 JxlDecoderOptions::default(),
                 OutputDataType::F32,
-                LinearOutput::No,
+                false,
             )
             .unwrap();
             let f32_output = f32_typed_output.to_f32().unwrap();
@@ -51,7 +51,7 @@ mod tests {
                     &mut input,
                     JxlDecoderOptions::default(),
                     data_type,
-                    LinearOutput::No,
+                    false,
                 )
                 .unwrap();
                 // Convert to f32 for comparison
@@ -142,7 +142,7 @@ mod tests {
             &mut input,
             high_precision_options(),
             OutputDataType::F32,
-            LinearOutput::No,
+            false,
         )
         .unwrap();
         let f32_output = f32_typed_output.to_f32().unwrap();
@@ -153,7 +153,7 @@ mod tests {
             &mut input,
             high_precision_options(),
             OutputDataType::U8,
-            LinearOutput::No,
+            false,
         )
         .unwrap();
         let u8_output = u8_typed_output.to_f32().unwrap();


zond:
@tirr-c, @sboukortt, @veluca93 this is a bit of a mess - I _think_ it's correct, more images look correct now than they used to, and I believe the naming and syntax is easier to read. WDYT?

veluca93:
The 20% slowdown seems like an issue :-)

veluca93:
I am not sure why this is necessary - can you explain that?

zond:
> The 20% slowdown seems like an issue :-)

Yes - I suspect it's something that broke the optimization boundary I artificially created after the CMS stage branch :(

zond:
@veluca93 I think it works like libjxl now, except that we now allow TF-only changes for non-XYB images.

zond:
@veluca93 OK, check again now, now the change is smaller and does less. I _think_ it does what libjxl does, and no more, when it comes to transfer function conversions.

jxl/src/api/inner/codestream_parser/non_section.rs
```diff
@@ -157,52 +154,50 @@ impl CodestreamParser {
...
+                        }
                     }
                 } else {
-                    nonlinear_output_color_profile
-                })
+                    JxlColorEncoding::linear_srgb(is_gray)
+                };
+
+                // Apply linear TF if requested
+                let output_encoding = if decode_options.xyb_output_linear {
```
veluca93: I'm not sure why we even have that option... opinions on getting rid of it entirely? (maybe @sboukortt ?)

zond: I'm all for it - it doesn't seem like something the API user should have opinions about, but I also defer to @sboukortt.

zond: Can we commit now and possibly remove it later? I added a TODO. I'd love to send the next PR for gray support in CMS situations, which is on top of this, and then I'm more or less ready with the Firefox integration.

veluca93: I would say I'd prefer to remove it now, actually (worst case we add it back...)

zond: OK, xyb_linear_output is now gone.

Some changes to the render pipeline were necessary.

I also cleaned up a ~duplicate function in jxl_cli when I updated it to no longer use xyb_linear_output.

zond:
Now I just have to figure out why the conformance test breaks :.(

zond:
@veluca93 I think this needs another review. I have tried to be careful and fix every bug I ran into so that it works like in libjxl, but the diff has grown large and the render pipeline isn't my forte.

zond:
And now 82% slowdown...??? WTH

zond:
Now performance is equivalent, conformance tests pass, Firefox render images correctly. Woohoo!

goodusername123:
I haven't checked/tested a build/compile myself but I'm worried that this PR *might* bring over a unresolved mistake/issue from libjxl that causes certain images to look noticeably worse. For context: https://github.com/libjxl/libjxl/pull/4516

zond:
> I haven't checked/tested a build/compile myself but I'm worried that this PR _might_ bring over a unresolved mistake/issue from libjxl that causes certain images to look noticeably worse. For context: [libjxl/libjxl#4516](https://github.com/libjxl/libjxl/pull/4516)

If that is the case, I still don't think this PR is the forum to decide to change behavior vs libjxl. If the libjxl behavior is changed - then I think we should change jxl-rs behavior - not before.

jxl/src/api/color.rs
```diff
@@ -1153,30 +1211,122 @@ impl JxlColorProfile {
...
     }
+
+    /// Returns the transfer function if this is a simple color profile.
+    /// Returns None for ICC profiles or XYB.
+    pub fn transfer_function(&self) -> Option<&JxlTransferFunction> {
+        match self {
+            Self::Simple(JxlColorEncoding::RgbColorSpace {
+                transfer_function, ..
+            }) => Some(transfer_function),
+            Self::Simple(JxlColorEncoding::GrayscaleColorSpace {
```
sboukortt:
```suggestion
            })
            | Self::Simple(JxlColorEncoding::GrayscaleColorSpace {
```

jxl/src/frame/render.rs
```diff
@@ -429,9 +430,20 @@ impl Frame {
             }
         }
 
-        let mut linear = false;
         let output_color_info = OutputColorInfo::from_header(&decoder_state.file_header)?;
 
+        // Determine output TF: use output profile's TF if available, else fall back to embedded
```
sboukortt: Is this comment accurate? Does `output_color_info` indeed come from the embedded profile?

jxl/src/api/inner/codestream_parser/non_section.rs
```diff
@@ -157,52 +154,43 @@ impl CodestreamParser {
...
             } else {
                 embedded_color_profile.clone()
             };
-            self.embedded_color_profile = Some(embedded_color_profile);
+            self.embedded_color_profile = Some(embedded_color_profile.clone());
             // Only set default output_color_profile if not already configured by user
             if self.output_color_profile.is_none() {
                 self.output_color_profile = Some(output_color_profile);
+            } else {
+                // Validate user's output color profile choice (libjxl compatibility)
```
sboukortt, line 185: No validation needed in the XYB case?

jxl/src/frame/render.rs
```diff
@@ -450,24 +462,59 @@ impl Frame {
...
+                    false,
+                )))
+            })
+        } else {
+            // Non-XYB: data is in the embedded profile's space including TF
+            Some(input_profile.clone())
+        };
+
+        // Compare ORIGINAL input profile (not linearized cms_input_profile) with output.
+        // This matches libjxl's dec_xyb.cc:184:
```
sboukortt: Having a line number here without referencing a specific revision is very prone to becoming outdated. Let’s drop the number? (or add something like `(rev 53042ec)`)

jxl_cli/src/dec/mod.rs
```diff
@@ -133,6 +52,16 @@ pub enum OutputDataType {
...
+/// Whether to output linear light values.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum LinearOutput {
+    /// Apply the transfer function (e.g., sRGB gamma).
+    #[default]
+    No,
+    /// Output linear light values (for formats like EXR or numpy).
+    Yes,
+}
+
```
sboukortt: Maybe this could stand to be a humble boolean?

jxl/src/api/inner/codestream_parser/non_section.rs
zond: For XYB images, no validation is needed because:                                                                                                           
1. XYB stage always outputs linear data that can be converted to any target                                                                                
2. If CMS is available, it handles the full conversion                                                                                                     
3. If no CMS, FromLinearStage provides best-effort TF conversion (dec_cache.cc:322-323 uses FromLinearStage when !cms_set)

jxl/src/frame/render.rs
zond: Looking at OutputColorInfo::from_header in xyb.rs:64-142, it extracts color info from header.image_metadata.color_encoding (line 80-81), which is indeed the embedded profile, but I also clarified the comment.

jxl/src/frame/render.rs
zond: Added a revision reference.

jxl_cli/src/dec/mod.rs
zond: If you insist, but I like that one can see the meaning of the parameter at the call site, without needing an IDE that auto populates parameter names etc. WDYT?

sboukortt: I think I would tend to favour either:

- an argument comment à la C++ (`/*linear_output=*/true`), if that’s not too out-of-place in Rust (@veluca93?)

- an enum where the values are themselves descriptive (`enum OutputTfKind { #[default] Original, Linear }`)

- a struct with a bool field (`struct OutputParameters { linear: bool }`)

jxl/src/api/inner/codestream_parser/non_section.rs
sboukortt: I hadn’t realised (or perhaps remembered) that libjxl did it that way as well. It makes me a bit uneasy, but pending closer inspection, let’s proceed with having them align.

jxl_cli/src/dec/mod.rs
veluca93: I vote for using a bool with no special comment as we tend to do in the rest of the codebase ;)

PR #626, opened 2026-01-16
Fixed bugs with skipped frames

zond:


jxl/src/frame/mod.rs
```diff
@@ -266,6 +266,16 @@ impl Frame {
         Ok(decoder_state)
     }
 
+    /// Finalize a skipped frame. Unlike `finalize()`, this doesn't save reference frames
+    /// since the frame data wasn't decoded. Returns the decoder state for the next frame.
```
veluca93: This is not correct. We should just call `.finalize()` normally.

zond: Done

jxl/src/frame/mod.rs
```diff
@@ -245,12 +245,17 @@ impl Frame {
         // First, drop the render pipeline to ensure that no other references to the reference
         // frames are around.
         self.render_pipeline = None;
-        if self.header.can_be_referenced {
+        // Save reference frame if this frame can be referenced and was actually decoded.
+        // If reference_frame_data is None (frame was skipped), we don't save it.
```
veluca93, line 249: That's not correct. If the frame was skipped, it will (or should) still be decoded if it can be referenced, because it could i.e. be a patch source.

If we want to completely skip looking at a frame, we'd need to look at the frame index box to know whether a frame can be safely skipped in its entirety (but then we wouldn't call this function at all).

zond: Check now, is this OK?

zond:
@veluca93 it looks like you enabled auto merge but didn't approve?

veluca93 (approved):

PR #629, opened 2026-01-16
Fix integer underflow in String::read_unconditional

hjanuschka:
## Summary

Fixes an integer underflow panic found by ClusterFuzzLite fuzzing (PR #628).

**Bug:** The arithmetic expression `len as usize - ret.len() - 1 + n` in `String::read_unconditional` could underflow on malformed input, causing a panic.

**Impact:** A malformed JXL file with crafted extra channel name data could panic the decoder (denial of service).

**Fix:** Use saturating arithmetic to safely handle edge cases without panicking.

## Test plan

- [x] All 499 existing tests pass
- [ ] Verify fix with the crash input from fuzzing artifacts

### DIFF
--- a/jxl/src/headers/encodings.rs
+++ b/jxl/src/headers/encodings.rs
@@ -160,7 +160,12 @@ impl UnconditionalCoder<()> for String {
             match br.read_noinline(8) {
                 Ok(c) => ret.push(c as u8 as char),
                 Err(Error::OutOfBounds(n)) => {
-                    return Err(Error::OutOfBounds(len as usize - ret.len() - 1 + n));
+                    // Use saturating arithmetic to prevent underflow on malformed input
+                    // ret.len()+1 cannot overflow since ret.len() <= isize::MAX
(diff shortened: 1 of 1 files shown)


jxl/src/headers/encodings.rs
```diff
@@ -160,7 +160,12 @@ impl UnconditionalCoder<()> for String {
             match br.read_noinline(8) {
                 Ok(c) => ret.push(c as u8 as char),
                 Err(Error::OutOfBounds(n)) => {
-                    return Err(Error::OutOfBounds(len as usize - ret.len() - 1 + n));
+                    // Use saturating arithmetic to prevent underflow on malformed input
+                    let remaining = (len as usize)
```
veluca93, line 165: I think this should be `(len as usize).saturating_add(n).saturating_sub(ret.len()+1)` (`ret.len()+1` can never overflow because `ret.len() <= isize::MAX`)

PR #632, opened 2026-01-17
Fix integer overflow in RCT transform

hjanuschka:
## Summary
- Fixes integer overflow panic found by ClusterFuzzLite in RCT transform
- Adds wrapping_add/wrapping_sub methods to I32SimdVec trait
- Uses wrapping arithmetic for RCT operations (SIMD already wraps, this fixes scalar path)


tirr-c:
I think it's better to just make integer vector arithmetics default to wrapping.

(And there's already #603 open fwiw)

hjanuschka:
Done. Changed scalar types to use Wrapping<i32> and Wrapping<u32>. This should supersede #603

PR #637, opened 2026-01-20
fix: use extra channel's own bit_depth for modular-to-f32 conversion

hjanuschka:
## Summary

Takes over #604 from @lilith who is unable to continue.

- Use each extra channel's own `bit_depth` instead of the image's global `metadata.bit_depth` for modular-to-f32 conversion
- Add `bit_depth()` getter to `ExtraChannelInfo`
- Add tests demonstrating the ~257x error when wrong bit depth is used

Previously, an 8-bit extra channel decoded with a 16-bit image would have its max value (255) incorrectly mapped to ~0.00389 instead of 1.0.

Closes #604

Co-authored-by: Lilith River <lilith@imazen.io>


jxl/src/headers/extra_channels.rs
```diff
@@ -97,3 +100,119 @@ impl ExtraChannelInfo {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::headers::bit_depth::BitDepth;
+
+    /// Test that extra channels can have their own bit depth independent of the image.
```
veluca93: I don't think this test makes a lot of sense.

jxl/src/headers/extra_channels.rs
```diff
@@ -97,3 +100,119 @@ impl ExtraChannelInfo {
...
+    /// own bit_depth rather than the image's global bit_depth.
+    ///
+    /// The modular-to-f32 conversion scale is: 1.0 / ((1 << bits) - 1)
+    /// - 8-bit:  scale = 1/255,   so value 255 → 1.0
+    /// - 16-bit: scale = 1/65535, so value 255 → ~0.00389
+    ///
+    /// If an 8-bit extra channel is decoded using a 16-bit scale (the image's bit depth),
+    /// the maximum value (255) would map to 0.00389 instead of 1.0 - completely wrong!
+    #[test]
+    fn test_wrong_bit_depth_produces_wrong_conversion() {
```
veluca93: Also doesn't make a lot of sense.

jxl/src/headers/extra_channels.rs
```diff
@@ -97,3 +100,119 @@ impl ExtraChannelInfo {
...
+        assert_ne!(
+            ec_8bit.bit_depth().bits_per_sample(),
+            ec_16bit.bit_depth().bits_per_sample()
+        );
+    }
+
+    /// Test that the bit_depth getter returns the correct value for float samples.
+    #[test]
+    fn test_extra_channel_float_bit_depth() {
+        let ec_float = ExtraChannelInfo::new(
```
veluca93: AFAICT this is equivalent to `assert_eq!(BitDepth::f32().bits_per_sample(), 32);`, which I am not sure is too useful

jxl/src/render/stages/convert.rs
```diff
@@ -448,4 +448,86 @@ mod test {
     fn f32_to_f16_consistency() -> Result<()> {
         crate::render::test::test_stage_consistency(|| ConvertF32ToF16Stage::new(0), (500, 500), 1)
     }
+
+    /// Test that modular-to-f32 conversion scale depends on bit depth.
```
veluca93: I would remove this test too.

veluca93 (approved):

PR #638, opened 2026-01-20
security: fix integer overflow vulnerabilities

hjanuschka:
## Summary

Takes over #602 from @lilith who handed it off.

Fixes integer overflow vulnerabilities that could be exploited with malicious images:

- **patches.rs**: Use checked/saturating arithmetic for `num_pixels`, `max_ref_patches`, etc.
- **spline.rs**: Use `saturating_mul` for image area calculation
- **headers/size.rs**: Fix aspect ratio calculations that could overflow (e.g., `Ratio2Over1` with large ysize)
- **render/builder.rs**: Return `Result` from `new()` instead of panicking on group count overflow

The ICC tag overflow fix was already merged separately in a previous PR.

Closes #602

Co-authored-by: Lilith River <lilith@imazen.io>

### DIFF
--- a/jxl/src/headers/size.rs
+++ b/jxl/src/headers/size.rs
@@ -3,7 +3,11 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-use crate::{bit_reader::BitReader, error::Error, headers::encodings::*};
+use crate::{
+    bit_reader::BitReader,
+    error::Error,
+    headers::encodings::{self, *},
+};
 use jxl_macros::UnconditionalCoder;
 use num_derive::FromPrimitive;
 
@@ -21,6 +25,7 @@ enum AspectRatio {
 }
 
 #[derive(UnconditionalCoder, Debug, Clone, Default)]
+#[validate]
 pub struct Size {
     small: bool,
     #[condition(small)]
@@ -40,6 +45,7 @@ pub struct Size {
 }
 
 #[derive(UnconditionalCoder, Debug, Clone)]
+#[validate]
 pub struct Preview {
     div8: bool,
     #[condition(div8)]
@@ -80,7 +86,7 @@ impl Size {
         }
     }
 
-    pub fn xsize(&self) -> Result<u32, Error> {
+    fn compute_xsize(&self) -> Result<u32, Error> {
         let x = map_aspect_ratio(self.ysize(), self.ratio, /* fallback */ || {
             if self.small {
                 self.xsize_div8.unwrap() as u64 * 8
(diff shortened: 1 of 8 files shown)


jxl/src/features/patches.rs
```diff
@@ -369,10 +369,13 @@ impl PatchesDictionary {
...
             PatchContext::NumRefPatch as usize,
         ) as usize;
-        let num_pixels = xsize * ysize;
-        let max_ref_patches = 1024 + num_pixels / 4;
-        let max_patches = max_ref_patches * 4;
-        let max_blending_infos = max_patches * 4;
+        // Use checked arithmetic to prevent overflow attacks
+        let num_pixels = xsize
+            .checked_mul(ysize)
+            .ok_or(Error::ImageSizeTooLarge(xsize, ysize))?;
```
veluca93: I don't think we should check that `xsize*ysize` is not too large here. Instead, we should do it where we currently have the image area limits (basically by having a default area limit of isize::MAX, since no Rust in-memory object can ever exceed that size anyway).

Then, if a computation overflows here it is certainly a bug and not due to invalid files.

jxl/src/headers/size.rs
```diff
@@ -58,16 +58,31 @@ pub struct Preview {
...
+        AspectRatio::Ratio4Over3 => ysize as u64 * 4 / 3,
+        AspectRatio::Ratio3Over2 => ysize as u64 * 3 / 2,
+        AspectRatio::Ratio16Over9 => ysize as u64 * 16 / 9,
+        AspectRatio::Ratio5Over4 => ysize as u64 * 5 / 4,
+        AspectRatio::Ratio2Over1 => ysize as u64 * 2,
+    };
+    u32::try_from(result).ok()
+}
+
 fn map_aspect_ratio<T: Fn() -> u32>(ysize: u32, ratio: AspectRatio, fallback: T) -> u32 {
```
veluca93: Let's leave map_aspect_ratio as is, make it return a u64, and fail in the caller with an appropriate error.

jxl/src/render/builder.rs
```diff
@@ -27,14 +27,20 @@ impl<Pipeline: RenderPipeline> RenderPipelineBuilder<Pipeline> {
...
-    ) -> Self {
+    ) -> Result<Self> {
         info!("creating render pipeline");
         assert!(chunk_size <= u16::MAX as usize);
         assert_ne!(chunk_size, 0);
         // The number of pixels that a group encompasses in the final, upsampled image along one
         // dimension is effectively multiplied by the upsampling factor.
         log_group_size += downsampling_shift;
-        Self {
+        // Use checked_mul to prevent overflow with malicious image dimensions
```
veluca93: Again, I don't think we should check this here.

jxl/src/headers/size.rs
```diff
@@ -100,13 +101,14 @@ impl Preview {
...
             if self.div8 {
-                self.xsize_div8.unwrap() * 8
+                self.xsize_div8.unwrap() as u64 * 8
             } else {
-                self.xsize.unwrap()
+                self.xsize.unwrap() as u64
             }
-        })
+        });
+        u32::try_from(x).map_err(|_| Error::ImageDimensionTooLarge(x))
```
veluca93, line 129: I would rather use the header validation mechanism (see i.e. https://github.com/libjxl/jxl-rs/blob/db350bd840aa69989466b9efb79e112573944c32/jxl/src/headers/frame_header.rs#L266) to check no overflow and return an error upon decoding of the struct, and then have xsize() call the same method with a .unwrap().

This simplifies call sites, and still ensures that overflow scenarios are caught via a Result.

veluca93 (approved):

FooIbar:
20% slowdown strikes again...

PR #639, opened 2026-01-20
Made jxl_cli speedtests warm up

zond:


zond:
I would claim that the benchmark differences are due to the warmup.

veluca93 (approved):

PR #641, opened 2026-01-21
fix: validate Size/Preview once instead of per-frame

hjanuschka:
Removes #[validate] from Size and Preview structs to avoid 20% performance regression on modular images.

The overflow check now runs once via FileHeader::validate() after decoding, instead of on every frame decode via #[validate] attribute.

This is secure because frames don't have their own Size struct - they reference the main header dimensions via FrameHeaderNonserialized.

Fixes #638 regression


veluca93:
Superseded by #642.

PR #642, opened 2026-01-21
Ensure WP functions are inlined.

veluca93:
The heuristics around inlining those functions were especially brittle, and caused hard-to-predict 20% changes in performance. Adding inline(always) seems to recover the performance and is hopefully more consistent with small changes.

### DIFF
--- a/jxl/src/frame/modular/predict.rs
+++ b/jxl/src/frame/modular/predict.rs
@@ -326,10 +326,12 @@ const DIVLOOKUP: [u32; 64] = [
     289262, 284359, 279620, 275036, 270600, 266305, 262144,
 ];
 
+#[inline(always)]
 fn add_bits(x: i32) -> i64 {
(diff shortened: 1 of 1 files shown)


FooIbar:
It seems there is still a 3-4% performance loss https://github.com/FooIbar/jxl-rs/pull/3
green_queen_modular_e3.jxl can usually get 8 MP/s before #638

veluca93:
> It seems there is still a 3-4% performance loss [FooIbar#3](https://github.com/FooIbar/jxl-rs/pull/3) green_queen_modular_e3.jxl can usually get 8 MP/s before #638

I think that's fine for now - there are probably other functions that are not being inlined anymore, and that would be the reason for the ~3% slowdown, but I think we can live with it at the moment.

Happy to accept a PR that improves the performance by inlining more functions though ;)

PR #645, opened 2026-01-23
Fix noise generation seeding for upsampled frames

### LINKED ISSUE #610: Noise generation uses wrong seeding pattern when upsampling > 1
## Summary

When a frame has `upsampling > 1`, noise generation produces incorrect patterns because jxl-rs uses a single RNG seed for the entire upsampled area instead of iterating through upsampling subdivisions with separate seeds per subregion.

## Test file

`multiple_layers_noise_spline.jxl` from [codec-corpus](https://github.com/libjxl/libjxl-testdata) (via conformance testing)

- `upsampling=2`
- ~37% pixel errors before fix (error_count=3.5M/9.4M, max_error=66-70)
- Errors primarily in B (blue) channel with noise-like differences of 0-16

## Root cause

libjxl iterates through upsampling subdivisions with separate RNG seeds:

```cpp
// libjxl: loops over (iy, ix) in 0..upsampling
for (size_t iy = 0; iy < upsampling; iy++) {
  for (size_t ix = 0; ix < upsampling; ix++) {
    // Each subregion gets its own seed
    x0 = (gx * upsampling + ix) * group_dim;
    y0 = (gy * upsampling + iy) * group_dim;
    rng = Rng(visible_frame_index, nonvisible_frame_index, x0, y0);
    // ONE rng shared across all 3 channels for this subregion
    ...
  }
}
```

jxl-rs was using:
```rust
// Single seed for entire upsampled area
let x0 = gx * upsampling * group_dim;
let y0 = gy * upsampling * group_dim;
let mut rng = Xorshift128Plus::new_with_seeds(..., x0, y0);
// New RNG created per channel instead of shared
for i in 0..3 {
    // fill buffer...
}
```

## The fix

Restructure noise generation in `decode.rs` to:
1. Loop over `(iy, ix)` upsampling subdivisions first
2. Create ONE RNG per subregion, shared across all 3 color channels
3. Get all 3 noise buffers upfront and fill with correct patterns

**Fix implementation**: [`jxl/src/frame/decode.rs` lines 404-475](https://github.com/lilith/jxl-rs/blob/develop/jxl/src/frame/decode.rs#L404-L475)

## Test coverage added

Two debug tests that helped identify the root cause:

1. [`test_debug_multiple_layers_noise_spline`](https://github.com/lilith/jxl-rs/blob/develop/jxl/src/tests/codec_corpus.rs#L409) - Examines frame structure and compares pixels for the failing test case

2. [`test_debug_noise_upsampling`](https://github.com/lilith/jxl-rs/blob/develop/jxl/src/tests/codec_corpus.rs#L699) - Compares upsampling values across noise test files, which revealed that passing tests (`noise`, `noise_5`, `8x8_noise`) all have `upsampling=1`, while `multiple_layers_noise_spline` has `upsampling=2`

## Result

After fix: 183/184 codec-corpus tests pass (99.5%), only `cmyk_layers` remains (requires ICC CMS).

hjanuschka:
## Summary

Fix incorrect noise generation when `upsampling > 1`. The code now iterates through upsampling subdivisions with separate RNG seeds per subregion, matching libjxl's behavior.

**Problem:** When upsampling > 1, jxl-rs was using a single RNG seed for the entire upsampled area instead of creating separate seeds for each upsampling subdivision. This caused ~37% pixel errors for files like `multiple_layers_noise_spline.jxl`.

**Solution:** Restructure noise generation to:
1. Loop over `(iy, ix)` upsampling subdivisions first
2. Create one RNG per subregion with subdivision-specific seed coordinates: `x0 = (gx * upsampling + ix) * group_dim`
3. Share that RNG across all 3 color channels for each subregion

This matches the libjxl reference implementation.

Fixes #610

## Changes

- `jxl/src/frame/decode.rs`: Restructured noise generation loop
- `jxl/src/frame/mod.rs`: Added `noise_with_upsampling` regression test

## Credit

Based on the fix analysis from @lilith in issue #610.

### DIFF
--- a/jxl/src/api/inner/codestream_parser/non_section.rs
+++ b/jxl/src/api/inner/codestream_parser/non_section.rs
@@ -172,18 +172,17 @@ impl CodestreamParser {
             };
             self.embedded_color_profile = Some(embedded_color_profile.clone());
             // Only set default output_color_profile if not already configured by user
-            if self.output_color_profile.is_none() {
-                self.output_color_profile = Some(output_color_profile);
-            } else {
+            if let Some(user_profile) = &self.output_color_profile {
                 // Validate user's output color profile choice (libjxl compatibility)
                 // For non-XYB without CMS: only same encoding as embedded is allowed
-                let user_profile = self.output_color_profile.as_ref().unwrap();
                 if !file_header.image_metadata.xyb_encoded
                     && decode_options.cms.is_none()
                     && *user_profile != embedded_color_profile
                 {
                     return Err(Error::NonXybOutputNoCMS);
                 }
+            } else {
+                self.output_color_profile = Some(output_color_profile);
             }
             // Only set default pixel_format if not already configured (e.g. via rewind)
             if self.pixel_format.is_none() {
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -378,52 +378,101 @@ impl Frame {
             let xsize_groups = self.header.size_groups().0;
             let gx = (group % xsize_groups) as u32;
             let gy = (group / xsize_groups) as u32;
-            // TODO(sboukortt): test upsampling+noise
             let upsampling = self.header.upsampling;
-            let x0 = gx * upsampling * group_dim;
-            let y0 = gy * upsampling * group_dim;
-            let x1 = ((x0 + upsampling * group_dim) as usize).min(self.header.size_upsampled().0);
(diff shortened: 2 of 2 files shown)


jxl/src/frame/mod.rs
```diff
@@ -381,6 +381,34 @@ mod test {
         Ok(())
     }
 
+    /// Tests noise generation with upsampling > 1.
```
veluca93: Why not just add the file to the testdata, and let the existing tests deal with decoding it?

PR #647, opened 2026-01-23
Add JxlDecoderLimits API and CancellationToken for security

hjanuschka:
## Summary

Introduces a comprehensive security limits API to protect against denial-of-service attacks from malicious images that attempt to exhaust system resources.

### JxlDecoderLimits

Configurable limits struct with:
- `max_pixels`: Maximum total pixels allowed
- `max_extra_channels`: Maximum extra channels
- `max_icc_size`: Maximum ICC profile size
- `max_tree_size`: Maximum modular tree size
- `max_patches`: Maximum patches allowed
- `max_spline_points`: Maximum spline control points
- `max_reference_frames`: Maximum reference frames
- `max_memory_bytes`: Memory budget for allocations

### CancellationToken

Thread-safe cooperative cancellation for decode operations.

### Preset Configurations

- `JxlDecoderLimits::default()`: Unlimited (backwards compatible)
- `JxlDecoderLimits::default_safe()`: Balanced security for general use
- `JxlDecoderLimits::restrictive()`: Strict limits for untrusted content (web browsers, etc.)

## Backwards Compatibility

**The default configuration is unlimited**, preserving existing behavior. Applications must explicitly opt-in to security limits via `JxlDecoderOptions.limits`.

## Changes

- `jxl/src/api/limits.rs`: New module with `JxlDecoderLimits` and `CancellationToken`
- `jxl/src/api/options.rs`: Added `limits` and `cancellation_token` fields to `JxlDecoderOptions`
- `jxl/src/error.rs`: Added `LimitExceeded` and `Cancelled` error variants

## Note

This PR adds the API infrastructure. Actual limit enforcement at decode checkpoints can be added incrementally in follow-up PRs.

## Credit

Based on security features from @lilith in lilith/jxl-rs.

### DIFF
--- a/jxl/src/api/limits.rs
+++ b/jxl/src/api/limits.rs
@@ -1,0 +1,186 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//! Security limits and cancellation support for JXL decoding.
+
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// Configurable security limits for the JXL decoder.
+///
+/// These limits help protect against denial-of-service attacks from malicious
+/// images that attempt to exhaust system resources.
+///
+/// # Backwards Compatibility
+///
+/// By default, all limits are set to `None` (unlimited), which preserves the
+/// existing behavior. Use [`JxlDecoderLimits::default_safe()`] for recommended
+/// security limits, or [`JxlDecoderLimits::restrictive()`] for untrusted content.
+///
+/// # Example
+///
+/// ```
+/// use jxl::api::JxlDecoderLimits;
+///
+/// // Use default (unlimited) for trusted content
+/// let limits = JxlDecoderLimits::default();
+///
+/// // Use safe defaults for general use
+/// let limits = JxlDecoderLimits::default_safe();
+///
+/// // Use restrictive limits for untrusted web content
+/// let limits = JxlDecoderLimits::restrictive();
+/// ```
+#[derive(Clone, Debug)]
+pub struct JxlDecoderLimits {
+    /// Maximum total pixels allowed (width * height).
+    /// Default: `None` (unlimited).
+    /// Recommended safe: `1 << 30` (about 1 billion pixels).
+    pub max_pixels: Option<usize>,
+
+    /// Maximum number of extra channels allowed.
+    /// Default: `None` (unlimited).
+    /// Recommended safe: `256`.
+    pub max_extra_channels: Option<usize>,
+
+    /// Maximum ICC profile size in bytes.
+    /// Default: `None` (unlimited).
+    /// Recommended safe: `1 << 28` (256 MB).
+    pub max_icc_size: Option<usize>,
+
+    /// Maximum modular tree size.
+    /// Default: `None` (unlimited).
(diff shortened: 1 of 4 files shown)


lilith:
Hey, before CancellationToken propagates, I have switched to using the enough::Stop trait, which is compatible with everything from tokio, to ffi, to a simple AtomicBool.  https://crates.io/crates/enough

lilith:
I'm really glad to see these limits landing. I've been implementing similar limits in my gif, webp, and jpeg codecs, along with enough::Stop support.

lilith:
**Suggestion: Consider the `Stop` trait pattern**

Rather than a hardcoded `CancellationToken` struct, consider a minimal trait that allows users to provide their own implementations (timeouts, async integration, etc.):

```rust
/// Cooperative cancellation check.
pub trait Stop: Send + Sync {
    fn check(&self) -> Result<(), StopReason>;
}

#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[non_exhaustive]
pub enum StopReason {
    Cancelled,
    TimedOut,
}

/// Zero-cost "never stops" implementation.
#[derive(Debug, Clone, Copy, Default)]
pub struct Unstoppable;

impl Stop for Unstoppable {
    #[inline(always)]
    fn check(&self) -> Result<(), StopReason> { Ok(()) }
}
```

Then `CancellationToken` just implements this trait. Benefits:
- Users can provide custom implementations (deadlines, async cancellation tokens, etc.)
- `Unstoppable` optimizes away completely
- ~20 lines, no external dependency

This pattern is used by the [`enough`](https://crates.io/crates/enough) crate if you'd prefer a dependency over inline.

lilith:
Any particular reason this was closed? I would love to see this security feature

veluca93:
For the limits: jxl-rs has (now) a limit on the total number of pixels. All other limits are derived from that limit, and should be enough to prevent excessive buffer usage. Giving too much control on individual limits risks creating too much ecosystem fragmentation.

For cancellation tokens: I am not sure I understand the usecase for these (or at least, not use-cases that wouldn't be better served by different approaches). Can you elaborate?

lilith:
Cancellation of long running cpu tasks is key in a variety of contexts,
including for timeouts as a security measure, or when tabs or windows are
closed or when a server or application is trying to shut down, when a
webassembly app needs to return to the event loop, or in tokio when a
future is dropped or a network connection drops....

Now that this is in Chrome, shouldn't there be a security policy instead of
public issues?

lilith:
A pixel limit bounds output size but not internal allocations like property_ranges (which scales with tree_nodes × num_properties, not pixels) - a 16MP image can trigger 6GB allocation for 192MB output, so capping total decoder memory usage would catch amplification  attacks regardless of which internal structure is exploited.

lilith:
libjxl's encoder produces at most 35-60 properties (based on channel count and nb_prev_channels=11), but the spec allows 256 - capping at 64 would reject no legitimate files while preventing 4× amplification. Still, allocation limits would be safer

veluca93:
All allocations in jxl-rs should have a reasonable size based on the pixel limit.

Specifically for property_ranges, that's a good catch - the allocation size should be limited by `num_properties` times tree _height_, and the tree height should be limited to at most 2048 as per spec. However, the tree height limit is - AFAICT - not currently enforced, and the algorithm for checking property ranges uses more memory than it needs to.

I will open an issue for fixing this.

PR #649, opened 2026-01-23
Fix panic in as_icc() for malformed color profiles

hjanuschka:
## Summary

ClusterFuzz found that `as_icc()` could panic when processing malformed JXL files with invalid color profiles. The method used `.unwrap().unwrap()` which aborts on error.

## Changes

- Deprecate `as_icc()` in favor of `try_as_icc()` for safe error handling
- Make `as_icc()` delegate to `try_as_icc()` with `.expect()` for a clearer panic message
- Update CLI to use `try_as_icc()` to gracefully handle invalid profiles

## Test plan

- [x] cargo fmt
- [x] cargo clippy --all-targets --all-features -- -D warnings
- [x] cargo test

Fixes ClusterFuzz issue for Chromium bug 474417800.

### DIFF
--- a/jxl/src/api/color.rs
+++ b/jxl/src/api/color.rs
@@ -1176,11 +1176,13 @@ impl JxlColorProfile {
     /// # Panics
     /// Panics if the color encoding cannot generate an ICC profile.
     /// Consider using `try_as_icc` for fallible conversion.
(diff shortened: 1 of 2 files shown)


veluca93:
This doesn't solve the underlying issue.

PR #654, opened 2026-01-25
Use sRGB transfer curve as fallback when integer sample is requested

tirr-c:
Also make the decoder update the output profile when pixel format is changed, but only when the user haven't set a color profile until then.

This is a fix that corresponds to https://github.com/libjxl/libjxl/pull/4516.


jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -390,4 +396,49 @@ impl CodestreamParser {
...
+                    .color_data_format
+                    .unwrap_or(JxlDataFormat::U8 { bit_depth: 8 });
+                let is_float = matches!(
+                    data_format,
+                    JxlDataFormat::F32 { .. } | JxlDataFormat::F16 { .. }
+                );
+                if is_float {
+                    JxlColorEncoding::linear_srgb(self.is_gray)
+                } else {
+                    JxlColorEncoding::srgb(self.is_gray)
```
veluca93, line 434: (@sboukortt for possibly further comments)
Would it make sense to default to a wider gamut like P3 in this case?

jonnyawsom3: It's meant to be a fallback if no colorspace is requested, so I'd say sRGB is the safer option (And matches libjxl's API documentation). There's still a lot of applications that don't do color management at all

sboukortt: On the other hand, it’s the fallback when the embedded colorspace is an ICC profile we can’t convert back to, which makes a gamut wider than sRGB quite likely.

goodusername123: If the default where set to P3 and an image went through this path would it still look correct when outputting to a display with only sRGB? if so then yeah it would IMO make sense to default to a wider space like P3 to get a bit more range out of ICC profiles which can end up going outside of the sRGB gamut.

tirr-c: Will merge this for now, changing default color gamut can be done in another PR if needed.

PR #659, opened 2026-01-26
Factor out computation in fill_initial_buffers to only happen once.

veluca93:
Small (~1%) speedup.


veluca93:
I'm not sure the speedup is worth it, so for now leaving this as a draft.

PR #662, opened 2026-01-28
Add a flag to allow partial files.

veluca93:
Also adds a script to generate a demo of progressively decoding a file.


zond:
@veluca93 there's not much here, did you want a review or just let me know about this?

PR #663, opened 2026-01-30
feat: implement ICC profile generation for XYB color space

hjanuschka:
## Summary

Implements A2B0 and B2A0 ICC tags for XYB color encoding, matching the libjxl C++ implementation. Previously, `maybe_create_profile()` would panic with `todo!()` when called on XYB profiles, causing crashes in Chromium's JXL decoder when processing malformed JXL files.

## Changes

- Adds XYB color space constants matching libjxl:
  - Opsin absorbance biases
  - XYB offset/scale values  
  - ICC transformation matrix
- Implements `create_icc_lut_atob_tag_for_xyb()` that creates an mAB A2B0 tag with:
  - 2x2x2 CLUT for color cube interpolation
  - Parametric M curves for XYB unmixing
  - Identity B/A curves
- Reuses existing no-op B2A0 tag (required by Safari/Preview)

## References

- Fixes ClusterFuzz issue for Chromium: https://issues.chromium.org/issues/474401148
- ClusterFuzz report: https://clusterfuzz.com/testcase?key=5537166176223232
- Reference implementation: libjxl `CreateICCLutAtoBTagForXYB` in `lib/jxl/cms/jxl_cms_internal.h`

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -1303,10 +1303,8 @@ pub(crate) mod tests {
 
     /// Regression test for Chromium ClusterFuzz issue 474401148.
     #[test]
-    #[allow(clippy::collapsible_if)]
     fn test_fuzzer_xyb_icc_no_panic() {
         use crate::api::ProcessingResult;
-        use std::panic;
 
         #[rustfmt::skip]
         let data: &[u8] = &[
@@ -1314,32 +1312,17 @@ pub(crate) mod tests {
             0x00, 0x00, 0x00, 0x00, 0x00, 0x11, 0x25, 0x00,
         ];
 
-        let result = panic::catch_unwind(|| {
-            let opts = JxlDecoderOptions {
-                pixel_limit: Some(1024 * 1024 * 1024),
-                ..Default::default()
-            };
-            let mut decoder = JxlDecoderInner::new(opts);
-            let mut input = data;
+        let opts = JxlDecoderOptions {
+            pixel_limit: Some(1024 * 1024 * 1024),
+            ..Default::default()
+        };
+        let mut decoder = JxlDecoderInner::new(opts);
+        let mut input = data;
 
-            if let Ok(ProcessingResult::Complete { .. }) = decoder.process(&mut input, None) {
-                if let Some(profile) = decoder.output_color_profile() {
-                    let _ = profile.try_as_icc();
-                }
-            }
-        });
-
-        if let Err(e) = result {
-            let panic_msg = e
-                .downcast_ref::<&str>()
-                .map(|s| s.to_string())
-                .or_else(|| e.downcast_ref::<String>().cloned())
-                .unwrap_or_default();
-            assert!(
-                !panic_msg.contains("not yet implemented"),
-                "Unexpected panic: {}",
-                panic_msg
-            );
+        if let Ok(ProcessingResult::Complete { .. }) = decoder.process(&mut input, None)
+            && let Some(profile) = decoder.output_color_profile()
+        {
+            let _ = profile.try_as_icc();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


sboukortt (approved):

jxl/src/api/color.rs
```diff
@@ -2047,6 +2069,187 @@ fn tone_map_pixel(
     ])
 }
 
+// XYB color space constants (matching libjxl)
+// Allow excessive precision as these constants are copied verbatim from libjxl for compatibility
+#[allow(clippy::excessive_precision)]
+mod xyb_constants {
```
tirr-c: How about moving this module into separate file?

jxl/src/api/color.rs
```diff
@@ -2047,6 +2069,187 @@ fn tone_map_pixel(
...
+            [
+                [unscaled_a2b_corner(0, 0, 0), unscaled_a2b_corner(0, 0, 1)],
+                [unscaled_a2b_corner(0, 1, 0), unscaled_a2b_corner(0, 1, 1)],
+            ],
+            [
+                [unscaled_a2b_corner(1, 0, 0), unscaled_a2b_corner(1, 0, 1)],
+                [unscaled_a2b_corner(1, 1, 0), unscaled_a2b_corner(1, 1, 1)],
+            ],
+        ]
+    }
```
tirr-c: I suppose all of the functions in this module could be `const fn`. We could make these public functions into actual const if it works.

hjanuschka:
@tirr-c thank you, ready for re-review

jxl/src/api/color.rs
```diff
@@ -2047,6 +2069,110 @@ fn tone_map_pixel(
     ])
 }
 
+use super::xyb_constants;
+
+/// Create mAB A2B0 tag for XYB color space.
+fn create_icc_lut_atob_tag_for_xyb(tags: &mut Vec<u8>) -> Result<(), Error> {
+    use byteorder::{BigEndian, WriteBytesExt};
+    use xyb_constants::*;
```
tirr-c: nit: you can just say `use super::xyb_constants::*;` in the function and remove the other `use` statement.

tirr-c (approved):
LGTM. @veluca93 do you have any opinion on this?

veluca93:
> LGTM. @veluca93 do you have any opinion on this?

As far as I can tell, this matches libjxl's behaviour. The only doubt I have is on where to check -- at the moment we are checking both when decoding the header and when trying to figure out the color encoding, and I'm not sure if we should do that.

jxl/src/api/inner/codestream_parser/non_section.rs
```diff
@@ -117,6 +117,16 @@ impl CodestreamParser {
 
         if self.decoder_state.is_none() && self.embedded_color_profile.is_none() {
             let file_header = self.file_header.as_ref().unwrap();
+
+            // Validate: want_icc with XYB or Unknown color space is invalid (matches libjxl)
```
veluca93: Can we remove this condition from here, and just use the one in ColorEncoding::check?

jxl/src/headers/color_encoding.rs
```diff
@@ -192,6 +192,10 @@ pub struct ColorEncoding {
 
 impl ColorEncoding {
     pub fn check(&self, _: &Empty) -> Result<(), Error> {
+        // XYB color space is not supported as an output color space
```
veluca93: IMO we should just expand the condition below:
```
        if self.color_space == ColorSpace::Unknown
                || self.tf.transfer_function == TransferFunction::Unknown
                || self.color_space == ColorSpace::XYB
        {
            Err(Error::InvalidColorEncoding)
        } else {
            Ok(())
        }
```

jxl/src/api/decoder.rs
```diff
@@ -1300,4 +1300,46 @@ pub(crate) mod tests {
...
+        let result = panic::catch_unwind(|| {
+            let opts = JxlDecoderOptions {
+                pixel_limit: Some(1024 * 1024 * 1024),
+                ..Default::default()
+            };
+            let mut decoder = JxlDecoderInner::new(opts);
+            let mut input = data;
+
+            if let Ok(ProcessingResult::Complete { .. }) = decoder.process(&mut input, None) {
+                if let Some(profile) = decoder.output_color_profile() {
```
veluca93: This code should not panic at all, right?

So, we probably should skip the catch_unwind and string search.

Issue #648, opened 2026-01-23
Failure to decode: Section is too short


jonnyawsom3:
A fairly exotic file, but valid nonetheless. Decodes fine with libjxl, jxl-rs says:
```
Error: Section is too short

Location:
    jxl_cli\src\dec\mod.rs:108:41
```
Encoded with `cjxl -d 0 --modular_lossy_palette --modular_palette_colors 0 -R 1`, renamed to .jpg for upload.

![Image](https://github.com/user-attachments/assets/97257eb6-2ff2-48b3-9302-5f5bad7c552a)

PR #666, opened 2026-01-31
Fix modular decoding with empty channels

hjanuschka:
Fixes decoding for issue #648.

- Keep modular channel indices stable when some channels have zero size
- Skip decoding empty channels without renumbering
- Add regression test (issue648_palette0.jxl)


jxl/src/api/decoder.rs
```diff
@@ -213,6 +213,14 @@ pub(crate) mod tests {
         });
     }
 
+    #[test]
```
veluca93: We shouldn't need a specific test - adding this file to jxl/resources/test/ should be enough for tests that decode it to happen.

jxl/src/frame/modular/mod.rs
```diff
@@ -475,7 +475,7 @@ impl FullModularImage {
             trace!("Transform {i}: {ts:?}");
         }
 
-        with_buffers(&buffer_info, &section_buffer_indices[0], 0, true, |bufs| {
+        with_buffers(&buffer_info, &section_buffer_indices[0], 0, false, |bufs| {
```
veluca93: Are there any other uses of `with_buffers` that pass `true`? If yes, I suspect they might be incorrect.
If not, I think we can just remove the argument.

hjanuschka: nah, removed it completely and made with_buffer skip only 0,0 tiles

PR #667, opened 2026-01-31
Fix modular tree validation limits

### LINKED ISSUE #661: Tree height limit is not enforced
Moreover, property range checking uses more memory than necessary.

See https://github.com/libjxl/jxl-rs/pull/647#issuecomment-3804181708.

hjanuschka:
Fixes #661.

- Enforce the spec tree height limit (2048)
- Avoid allocating num_properties * tree_size property ranges (now O(num_properties))

### DIFF
--- a/jxl/src/frame/modular/tree.rs
+++ b/jxl/src/frame/modular/tree.rs
@@ -63,6 +63,124 @@ pub struct Tree {
     pub histograms: Histograms,
 }
 
+fn validate_tree(tree: &[TreeNode], num_properties: usize) -> Result<()> {
+    const HEIGHT_LIMIT: usize = 2048;
+
+    if tree.is_empty() {
+        return Ok(());
+    }
+
+    // This mirrors libjxl's ValidateTree(), but avoids allocating
+    // `num_properties * tree.len()` entries.
+    //
+    // We do an explicit DFS and keep the property ranges only for the current root->node path.
+    // When descending into a child we update exactly one property's range (the one we split on)
+    // and store the previous range in the child frame; when returning from that child we restore
+    // it. This makes memory O(num_properties + height) instead of O(num_properties * tree_size).
+
+    #[derive(Clone, Copy, Debug)]
+    enum Stage {
+        Enter,
+        AfterLeft,
+        AfterRight,
+    }
+
+    struct Frame {
+        node: usize,
+        depth: usize,
+        stage: Stage,
+        restore: Option<(usize, (i32, i32))>,
+    }
+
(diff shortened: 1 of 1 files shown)


jxl/src/frame/modular/tree.rs
```diff
@@ -357,50 +359,109 @@ impl Tree {
...
-        let num_properties = max_property as usize + 1;
-        let mut property_ranges = Vec::new_with_capacity(num_properties * tree.len())?;
-        property_ranges.resize(num_properties * tree.len(), (i32::MIN, i32::MAX));
-        let mut height = Vec::new_with_capacity(tree.len())?;
-        height.resize(tree.len(), 0);
-        for i in 0..tree.len() {
-            const HEIGHT_LIMIT: usize = 2048;
-            if height[i] > HEIGHT_LIMIT {
-                return Err(Error::TreeTooLarge(height[i], HEIGHT_LIMIT));
+        // Validate tree height and property ranges.
```
veluca93: What do you think of moving this in a separate function?

jxl/src/frame/modular/tree.rs
```diff
@@ -357,50 +359,109 @@ impl Tree {
...
-            } = tree[i]
-            {
-                height[left as usize] = height[i] + 1;
-                height[right as usize] = height[i] + 1;
-                for p in 0..num_properties {
-                    if p == property as usize {
-                        let (l, u) = property_ranges[i * num_properties + p];
+
+            let mut stack = Vec::new();
+            stack.push(Frame {
```
veluca93: Can you add a comment explaining what this code is doing? Not sure how much a future reader is likely to understand the (nice) approach here to use O(num_properties + height) memory. While in some sense it is relatively standard, it is not _incredibly_ so.

On a related note, I don't believe we need `has_split`, and I think it would be nice to use a local enum for the three stages.

jxl/src/frame/modular/tree.rs
```diff
@@ -357,54 +357,129 @@ impl Tree {
         }
         tree_reader.check_final_state(&tree_histograms, br)?;
 
-        let num_properties = max_property as usize + 1;
-        let mut property_ranges = Vec::new_with_capacity(num_properties * tree.len())?;
-        property_ranges.resize(num_properties * tree.len(), (i32::MIN, i32::MAX));
-        let mut height = Vec::new_with_capacity(tree.len())?;
-        height.resize(tree.len(), 0);
-        for i in 0..tree.len() {
+        // Validate tree height and property ranges.
```
veluca93: Can you move the function the top level in the file, instead of nesting it?

veluca93 (approved):

PR #669, opened 2026-02-01
Implement progressive decoding for VarDCT images.

veluca93:
Smooth upsampling of DC and supporting modular images/extra channels
is still TBD.

Also remove the "enable_output" decoding option.

Includes PR #662.

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -509,7 +509,7 @@ impl Frame {
         let pass_to_render = self.last_rendered_pass[group];
         let complete = pass_to_render.is_some_and(|p| p >= last_pass_in_file);
 
-        // Render if we are decoding the last pass, or if we are requesting a eager render and
+        // Render if we are decoding the last pass, or if we are requesting an eager render and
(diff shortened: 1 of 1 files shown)


jxl/src/frame/decode.rs
```diff
@@ -365,16 +487,43 @@ impl Frame {
...
+            assert_ne!(self.last_rendered_pass[group], Some(last_pass_in_file));
+        }
+
+        if let Some((p, _)) = passes.last() {
+            self.last_rendered_pass[group] = Some(*p);
+        };
+        let pass_to_render = self.last_rendered_pass[group];
+        let complete = pass_to_render.is_some_and(|p| p >= last_pass_in_file);
+
+        // Render if we are decoding the last pass, or if we are requesting a eager render and
```
mo271:
```suggestion
        // Render if we are decoding the last pass, or if we are requesting an eager render and
```

veluca93:
> Nice!
> 
> Does this work with `./tools/progressive_decoding_demo.sh`?

yup

PR #670, opened 2026-02-02
Fix ClusterFuzz issue 480111853

hjanuschka:
Fixes ClusterFuzz report for Chromium issue 480111853.

ClusterFuzz: https://clusterfuzz.com/testcase?key=6502319399043072
Chromium issue: 480111853

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -1325,4 +1325,45 @@ pub(crate) mod tests {
             let _ = profile.try_as_icc();
         }
     }
+
+    /// Regression test for Chromium ClusterFuzz issue 480111853.
+    #[test]
+    fn test_clusterfuzz_480111853_respects_output_channel_pixel_limit() {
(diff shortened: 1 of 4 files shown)


hjanuschka:
The issue was that pixel_limit was enforced only using the header channel count (3 + extra channels).

If the caller requested RGBA output, we never re-checked, so decoded samples could exceed the limit and lead to oversized allocations. 

The ClusterFuzz input (991439x355) is under 1 GiB for RGB but over for RGBA. 

This change enforces pixel_limit based on the requested output pixel format right before pipeline setup (for both section/non-section paths) and adds a regression test for the minimized fuzz input.

veluca93:
I see.

I am not convinced that this is the correct approach. Instead, why don't we just lower the pixel limit in the fuzzer to ~half of what it is?

It is after all the caller that allocates the RGBA buffer here, and not the library.

hjanuschka:
fixed on chromium side:  https://chromium-review.googlesource.com/c/chromium/src/+/7600793

PR #671, opened 2026-02-02
Fix index out of bounds in `decode_vardct_group`

FooIbar:
Fixes https://github.com/libjxl/jxl-rs/actions/runs/21305626535

libjxl will pad the context map to `kZeroDensityContextLimit` https://github.com/libjxl/libjxl/blob/8ce9537c989cfc7adff034556c8a4b9469e874d6/lib/jxl/dec_frame.cc#L410
I'm not sure which way is better


veluca93:
I think it might be better to pad the context map.

PR #672, opened 2026-02-02
Some speedups

jonsneyers:
Some minor speedups, thanks to these two tweaks:

- drop bounds check in bucket lookup in ANS (unsafe but proof is given in comment)
- further specialize the case of gradient-only/WP-only trees with no-lz77 hybriduint420, as is used e.g. in libjxl d0 e2/e3

Results:
- ~14% faster for lossless e2
- ~4% faster for lossless e3
- ~1-2% faster for higher-effort lossless
- ~3% faster for high-quality vardct (where entropy decode is a big cost)
- ~1-2% faster for lower quality vardct (where entropy decode is less important)

Before:
| Image | Resolution | Iterations | Total (ms) | Avg (ms) | Speed (MP/s) |
|-------|-----------|------------|------------|----------|-------------|
| bridge.ppm.e2.jxl | 2749x4049 | 20 | 13529.00 | 676.45 | 16.45 |
| bridge.ppm.e3.jxl | 2749x4049 | 20 | 35746.00 | 1787.30 | 6.23 |
| bridge.ppm.jxl | 2749x4049 | 20 | 65539.00 | 3276.95 | 3.40 |
| bridge.ppm.p.jxl | 2749x4049 | 20 | 40824.00 | 2041.20 | 5.45 |
| bridge.ppm.lossy_d1.jxl | 2749x4049 | 20 | 8197.00 | 409.85 | 27.16 |
| bridge.ppm.lossy_d2.jxl | 2749x4049 | 20 | 6594.00 | 329.70 | 33.76 |

After:
| Image | Resolution | Iterations | Total (ms) | Avg (ms) | Speed (MP/s) |
|-------|-----------|------------|------------|----------|-------------|
| bridge.ppm.e2.jxl | 2749x4049 | 20 | 11663.00 | 583.15 | 19.09 |
| bridge.ppm.e3.jxl | 2749x4049 | 20 | 34135.00 | 1706.75 | 6.52 |
| bridge.ppm.jxl | 2749x4049 | 20 | 64419.00 | 3220.95 | 3.46 |
| bridge.ppm.p.jxl | 2749x4049 | 20 | 39780.00 | 1989.00 | 5.60 |
| bridge.ppm.lossy_d1.jxl | 2749x4049 | 20 | 7957.00 | 397.85 | 27.98 |
| bridge.ppm.lossy_d2.jxl | 2749x4049 | 20 | 6499.00 | 324.95 | 34.25 |

### DIFF
--- a/jxl/src/entropy_coding/ans.rs
+++ b/jxl/src/entropy_coding/ans.rs
@@ -15,7 +15,9 @@ const RLE_MARKER_SYM: u16 = LOG_SUM_PROBS as u16 + 1;
 
 #[derive(Debug)]
 struct AnsHistogram {
-    // Safety invariant: buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
+    // Safety invariant:
+    // - log_bucket_size <= LOG_SUM_PROBS
+    // - buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
     // This relationship ensures that for any ANS state (12 bits), the bucket index
     // computed as (state & 0xfff) >> log_bucket_size is always < buckets.len()
     buckets: Vec<Bucket>,
@@ -268,8 +270,7 @@ impl AnsHistogram {
         debug_assert!((5..=8).contains(&log_alpha_size));
         let table_size = (1u16 << log_alpha_size) as usize;
         // 4 <= log_bucket_size <= 7
-        let log_bucket_size = LOG_SUM_PROBS - log_alpha_size;
-        assert!(log_bucket_size <= LOG_SUM_PROBS);
+        let log_bucket_size = LOG_SUM_PROBS.checked_sub(log_alpha_size).unwrap();
         let bucket_size = 1u16 << log_bucket_size;
         let bucket_mask = bucket_size as u32 - 1;
 
@@ -285,10 +286,9 @@ impl AnsHistogram {
         } else {
             Self::decode_dist_complex(br, &mut dist)?
         };
-
-        if let Some(single_sym_idx) = dist.iter().position(|&d| d == SUM_PROBS) {
-            let buckets: Vec<Bucket> = dist
-                .into_iter()
+        let single_symbol = dist.iter().position(|&d| d == SUM_PROBS).map(|x| x as u32);
+        let buckets = if let Some(single_sym_idx) = single_symbol {
+            dist.into_iter()
                 .enumerate()
                 .map(|(i, dist)| Bucket {
                     dist,
@@ -297,25 +297,19 @@ impl AnsHistogram {
                     alias_cutoff: 0,
                     alias_dist_xor: dist ^ SUM_PROBS,
                 })
-                .collect();
-            assert_eq!(buckets.len(), 1 << (LOG_SUM_PROBS - log_bucket_size));
-            // Safety note: we just checked that buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
-            return Ok(Self {
-                buckets,
-                log_bucket_size,
-                bucket_mask,
-                single_symbol: Some(single_sym_idx as u32),
-            });
-        }
+                .collect()
+        } else {
+            Self::build_alias_map(alphabet_size, log_bucket_size, &dist)
+        };
 
-        let buckets = Self::build_alias_map(alphabet_size, log_bucket_size, &dist);
         assert_eq!(buckets.len(), 1 << (LOG_SUM_PROBS - log_bucket_size));
-        // Safety note: we just checked that buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
+        // Safety note: log_bucket_size <= LOG_SUM_PROBS by construction, and we
+        // just checked that buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
         Ok(Self {
             buckets,
             log_bucket_size,
             bucket_mask,
-            single_symbol: None,
+            single_symbol,
         })
     }
 


jxl/src/entropy_coding/ans.rs
```diff
@@ -356,7 +356,24 @@ impl AnsHistogram {
...
+            i < self.buckets.len(),
+            "bucket index {} out of bounds (len = {})",
+            i,
+            self.buckets.len()
+        );
+        // SAFETY: i < buckets.len() is guaranteed by the following invariants:
+        //   - idx = state & 0xfff, so 0 <= idx <= 4095 = 2^12 - 1
+        //   - i = idx >> log_bucket_size
+        //   - log_bucket_size = LOG_SUM_PROBS - log_alpha_size = 12 - log_alpha_size
+        //   - buckets.len() = dist.len() = table_size = 2^log_alpha_size
```
veluca93: This invariant (and the one in the line above) are non-local invariants. Usually, those are written as safety invariants on the AnsHistogram struct, and checked upon struct construction (i.e. https://github.com/libjxl/jxl-rs/blob/b790bad78a62d21d78c457930108f94483973e16/jxl/src/entropy_coding/ans.rs#L305 in this case).

See for example https://github.com/libjxl/jxl-rs/blob/b790bad78a62d21d78c457930108f94483973e16/jxl/src/image/raw.rs#L12 for some other struct-level safety invariants.

jxl/src/entropy_coding/decode.rs
```diff
@@ -393,6 +393,38 @@ impl SymbolReader {
         unpack_signed(unsigned)
     }
 
+    /// Specialized fast path for when all HybridUint configs are 420.
+    /// SAFETY: This method assumes no LZ77 is active (only handles SymbolReaderState::None).
```
veluca93: This is not a safety requirement.

In general:
- Safety requirements are conditions that the caller needs to satisfy _to avoid causing UB_ that cannot be guaranteed just from the function signature.
- if a function has a safety _requirement_, the function needs to be unsafe, and the requirement is typically documented as `# Safety\n bla bla`
- safety _comments_ (`SAFETY: bla bla`) are comments on unsafe blocks that clarify why the unsafe operations in the block are OK. They either refer to the safety documentation of other functions, or to the documentation of language-level unsafe operations

jxl/src/frame/modular/decode/specialized_trees.rs
```diff
@@ -220,6 +220,60 @@ impl WpOnlyLookup {
     }
 }
 
+/// Specialized WpOnlyLookup for when all HybridUint configs are 420
+/// This allows using the fast-path entropy decoder
+pub struct WpOnlyLookupConfig420 {
```
veluca93, line 207: If we add this, do we still need the non-420 wp-only case? are such files common?

jxl/src/frame/modular/decode/specialized_trees.rs
```diff
@@ -319,6 +373,76 @@ impl ModularChannelDecoder for GradientLookup {
     }
 }
 
+/// Config 420 specialized version of GradientLookup.
+/// This uses the specialized entropy decoder for config 420 + no LZ77.
+pub struct GradientLookupConfig420 {
```
veluca93, line 264: Same question here

jxl/src/frame/modular/decode/specialized_trees.rs
jonsneyers: Good point. Probably the non-420 gradient-only/wp-only cases are pretty rare, at least for files produced by libjxl.

I _think_ that for e2/e3 lossless, we always have only-420 no-lz77, so we don't need the partial specialization for those, while for higher effort, it's unlikely that we have gradient-only/wp-only trees (and the cases where we do accidentally have them, will still mostly be only-420 no-lz77).

So I suppose we could get rid of the existing partial specialization and only have this full specialization. Would you prefer that?

veluca93: Yep :-)

jxl/src/entropy_coding/ans.rs
jonsneyers: Modified it like this.

jxl/src/entropy_coding/decode.rs
jonsneyers: Right. In this case these are just preconditions that should be satisfied when using this function; no UB will happen if the preconditions are not met, the result will just not be correct.

Updated the comments.

jxl/src/frame/modular/decode/specialized_trees.rs
jonsneyers: OK, done.

jxl/src/frame/modular/decode/specialized_trees.rs
jonsneyers: See above.

jxl/src/entropy_coding/ans.rs
```diff
@@ -294,6 +298,8 @@ impl AnsHistogram {
                     alias_dist_xor: dist ^ SUM_PROBS,
                 })
                 .collect();
+            // Verify safety invariant: buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
+            debug_assert_eq!(buckets.len(), 1 << (LOG_SUM_PROBS - log_bucket_size));
```
veluca93: this should be an `assert_eq!`.

jxl/src/entropy_coding/ans.rs
```diff
@@ -294,6 +298,8 @@ impl AnsHistogram {
...
             return Ok(Self {
```
veluca93: I'd comment here: `// Safety note: we just checked that `buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)`.

(we probably should also assert that `log_bucket_size <= LOG_SUM_PROBS`, and write that as part of the safety invariant)

jxl/src/entropy_coding/ans.rs
```diff
@@ -15,6 +15,10 @@ const RLE_MARKER_SYM: u16 = LOG_SUM_PROBS as u16 + 1;
 
 #[derive(Debug)]
 struct AnsHistogram {
+    // Safety invariant: buckets.len() = 2^(LOG_SUM_PROBS - log_bucket_size)
+    // This relationship ensures that for any ANS state (12 bits), the bucket index
+    // computed as (state & 0xfff) >> log_bucket_size is always < buckets.len()
+    // This invariant is verified in the decode() constructor.
```
veluca93: This line seems to imply that it is not checked elsewhere - I would just omit it.

(the details of how the safety invariant is used are not necessary either, but don't hurt)

jonsneyers: Dropped that line.

jxl/src/entropy_coding/ans.rs
jonsneyers: Updated both of them.

jxl/src/entropy_coding/ans.rs
jonsneyers: Updated the comment.

Do we need an explicit assert for `log_bucket_size <= LOG_SUM_PROBS`? It's defined in the same function, as `let log_bucket_size = LOG_SUM_PROBS - log_alpha_size;`, and `log_alpha_size` is unsigned.

veluca93: either an assert, or replacing the subtraction with `checked_sub` or `saturating_sub`.

jonsneyers: added an assert

PR #678, opened 2026-02-08
Add FrameInfoDecoder and start_new_frame() for animation support

hjanuschka:
Adds animation frame scanning + seeking support using the regular decoder API.

- Add `JxlDecoderOptions::scan_frames_only` to parse frame headers/TOC without pixel output.
- Add `JxlDecoder::scanned_frames()` and `JxlDecoder<WithImageInfo>::start_new_frame(remaining_in_box)`.
- Expose `VisibleFrameInfo` with timing + seek metadata (`is_keyframe`, `decode_start_file_offset`, `remaining_in_box`).
- Use slot-based dependency tracking for keyframe/decode-start computation (reference + LF slots, conservative patch handling).
- Supports both bare codestream and boxed/container input paths.

<details>
<summary>Chromium before/after</summary>

### DecodeFrameCount()

Before -- decodes all pixels for all frames to count them:

```cpp
while (decoder_state_ != DecoderState::kDone) {
    Decode(num_decoded_frames_, /*only_size=*/false);
}
return frame_buffer_cache_.size();
```

After -- header scan only, no pixel work:

```cpp
(*frame_scanner_)->feed({buf.data(), buf.size()});
return (*frame_scanner_)->frame_count();
```

### ClearCacheExceptFrame()

Before -- disabled for animations, all frames pinned in memory:

```cpp
if (basic_info_.have_animation) {
    return 0;
}
```

After -- frames can be evicted and re-decoded via start_new_frame().

### Decode(index) rewind

Before -- rewinds to byte 0, re-decodes everything:

```cpp
(*decoder_)->rewind();
input_offset_ = 0;
```

After -- seeks to nearest keyframe:

```cpp
(*decoder_)->start_new_frame();
input_offset_ = target.decode_start_offset;
```

</details>

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -71,18 +71,6 @@ pub struct VisibleFrameSeekTarget {
     pub visible_frames_to_skip: usize,
 }
 
-/// Compute seek inputs for `visible_frame_index` from scanned frame info.
-///
-/// Returns `None` if `visible_frame_index` is out of range.
-pub fn seek_target_for_visible_frame(
-    scanned_frames: &[VisibleFrameInfo],
-    visible_frame_index: usize,
-) -> Option<VisibleFrameSeekTarget> {
-    scanned_frames
-        .get(visible_frame_index)
-        .map(|f| f.seek_target)
-}
-
 impl<S: JxlState> JxlDecoder<S> {
     fn wrap_inner(inner: Box<JxlDecoderInner>) -> Self {
         Self {
@@ -107,6 +95,9 @@ impl<S: JxlState> JxlDecoder<S> {
     /// The frame index box (`jxli`) is an optional part of the JXL container
     /// format that provides a seek table for animated files, listing keyframe
     /// byte offsets, timestamps, and frame counts.
+    ///
+    /// TODO(veluca): Provide a higher-level frame-index API aligned with
+    /// `scanned_frames()` / `VisibleFrameInfo` seek metadata.
     pub fn frame_index(&self) -> Option<&FrameIndexBox> {
         self.inner.frame_index()
     }
@@ -209,23 +200,6 @@ impl JxlDecoder<WithImageInfo> {
 
     pub fn has_more_frames(&self) -> bool {
         self.inner.has_more_frames()
-    }
-
-    /// Prepare and apply a seek to `visible_frame_index` from scanned metadata.
-    ///
-    /// This calls [`start_new_frame`](Self::start_new_frame) internally and
-    /// returns the input offset and number of visible frames to skip before
-    /// decoding the requested target frame.
-    ///
-    /// Returns `None` if `visible_frame_index` is out of range.
-    pub fn seek_to_visible_frame(
-        &mut self,
-        scanned_frames: &[VisibleFrameInfo],
-        visible_frame_index: usize,
-    ) -> Option<VisibleFrameSeekTarget> {
-        let seek_target = seek_target_for_visible_frame(scanned_frames, visible_frame_index)?;
-        self.start_new_frame(seek_target);
-        Some(seek_target)
     }
 
     /// Resets frame-level decoder state to prepare for decoding a new frame.
@@ -1476,6 +1450,26 @@ pub(crate) mod tests {
         }
     }
 
+    fn make_box(ty: &[u8; 4], content: &[u8]) -> Vec<u8> {
+        let len = (8 + content.len()) as u32;
+        let mut buf = Vec::new();
+        buf.extend(len.to_be_bytes());
+        buf.extend(ty);
+        buf.extend(content);
+        buf
+    }
+
+    fn add_container_header(container: &mut Vec<u8>) {
+        // JXL signature box
+        let sig = [
+            0x00, 0x00, 0x00, 0x0c, 0x4a, 0x58, 0x4c, 0x20, 0x0d, 0x0a, 0x87, 0x0a,
+        ];
+        // ftyp box
+        let ftyp = make_box(b"ftyp", b"jxl \x00\x00\x00\x00jxl ");
+        container.extend(&sig);
+        container.extend(&ftyp);
+    }
+
     /// Helper to wrap a bare codestream in a JXL container with a jxli frame index box.
     fn wrap_with_frame_index(
         codestream: &[u8],
@@ -1485,31 +1479,52 @@ pub(crate) mod tests {
     ) -> Vec<u8> {
         use crate::util::test::build_frame_index_content;
 
-        fn make_box(ty: &[u8; 4], content: &[u8]) -> Vec<u8> {
-            let len = (8 + content.len()) as u32;
-            let mut buf = Vec::new();
-            buf.extend(len.to_be_bytes());
-            buf.extend(ty);
-            buf.extend(content);
-            buf
-        }
-
         let jxli_content = build_frame_index_content(tnum, tden, entries);
 
-        // JXL signature box
-        let sig = [
-            0x00, 0x00, 0x00, 0x0c, 0x4a, 0x58, 0x4c, 0x20, 0x0d, 0x0a, 0x87, 0x0a,
-        ];
-        // ftyp box
-        let ftyp = make_box(b"ftyp", b"jxl \x00\x00\x00\x00jxl ");
         let jxli = make_box(b"jxli", &jxli_content);
         let jxlc = make_box(b"jxlc", codestream);
 
         let mut container = Vec::new();
-        container.extend(&sig);
-        container.extend(&ftyp);
+        add_container_header(&mut container);
         container.extend(&jxli);
         container.extend(&jxlc);
+        container
+    }
+
+    /// Helper to wrap a bare codestream in a container split across jxlp boxes.
+    ///
+    /// `chunk_starts` are codestream offsets where each new jxlp chunk begins.
+    fn wrap_with_jxlp_chunks(codestream: &[u8], chunk_starts: &[usize]) -> Vec<u8> {
+        let mut starts = chunk_starts.to_vec();
+        starts.sort_unstable();
+        starts.dedup();
+        if starts.first().copied() != Some(0) {
+            starts.insert(0, 0);
+        }
+        if starts.last().copied() != Some(codestream.len()) {
+            starts.push(codestream.len());
+        }
+        assert!(starts.len() >= 2);
+
+        let mut container = Vec::new();
+        add_container_header(&mut container);
+
+        let num_chunks = starts.len() - 1;
+        for i in 0..num_chunks {
+            let begin = starts[i];
+            let end = starts[i + 1];
+            assert!(begin <= end && end <= codestream.len());
+
+            let mut payload = Vec::with_capacity(4 + (end - begin));
+            let mut index = i as u32;
+            if i + 1 == num_chunks {
+                index |= 0x8000_0000;
+            }
+            payload.extend(index.to_be_bytes());
+            payload.extend(&codestream[begin..end]);
+            container.extend(make_box(b"jxlp", &payload));
+        }
+
         container
     }
 
@@ -1580,10 +1595,7 @@ pub(crate) mod tests {
         assert!(dec.frame_index().is_none());
     }
 
-    fn scan_frames_with_decoder(
-        mut input: &[u8],
-        chunk_size: usize,
-    ) -> (bool, Vec<VisibleFrameInfo>) {
+    fn scan_frames_with_decoder(mut input: &[u8], chunk_size: usize) -> Vec<VisibleFrameInfo> {
         let mut chunk_input = &input[0..0];
         let options = JxlDecoderOptions {
             scan_frames_only: true,
@@ -1635,13 +1647,9 @@ pub(crate) mod tests {
         }
 
         let mut decoder_with_image_info = advance_process!(initialized_decoder);
-        let is_animated = decoder_with_image_info.basic_info().animation.is_some();
 
         if !decoder_with_image_info.has_more_frames() {
-            return (
-                is_animated,
-                decoder_with_image_info.scanned_frames().to_vec(),
-            );
+            return decoder_with_image_info.scanned_frames().to_vec();
         }
 
         loop {
@@ -1652,10 +1660,7 @@ pub(crate) mod tests {
             }
         }
 
-        (
-            is_animated,
-            decoder_with_image_info.scanned_frames().to_vec(),
-        )
+        decoder_with_image_info.scanned_frames().to_vec()
     }
 
     fn assert_start_new_frame_matches_sequential(data: &[u8], expect_bare_codestream: bool) {
@@ -1663,13 +1668,12 @@ pub(crate) mod tests {
         use crate::image::{Image, Rect};
 
         // 1. Scan frame info to get seek offsets.
-        let (_is_animated, scanned_frames) = scan_frames_with_decoder(data, usize::MAX);
+        let scanned_frames = scan_frames_with_decoder(data, usize::MAX);
         assert!(scanned_frames.len() > 1, "need multiple frames");
 
         // Compare against second visible frame from regular sequential decode.
         let target_visible_index = 1;
-        let seek_target = seek_target_for_visible_frame(&scanned_frames, target_visible_index)
-            .expect("seek target for visible frame");
+        let seek_target = scanned_frames[target_visible_index].seek_target;
 
         if expect_bare_codestream {
             assert_eq!(seek_target.remaining_in_box, u64::MAX);
@@ -1713,10 +1717,7 @@ pub(crate) mod tests {
         let num_ec = requested_format.extra_channel_format.len();
 
         // 4. Seek to decode-start and advance to the target visible frame.
-        let seek_target_from_method = decoder
-            .seek_to_visible_frame(&scanned_frames, target_visible_index)
-            .expect("seek target for visible frame");
-        assert_eq!(seek_target_from_method, seek_target);
+        decoder.start_new_frame(seek_target);
         let mut input = &data[seek_target.decode_start_file_offset..];
 
         for _ in 0..seek_target.visible_frames_to_skip {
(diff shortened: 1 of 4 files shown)


jxl/src/api/frame_info.rs
```diff
@@ -0,0 +1,767 @@
...
+///     JxlDecoderOptions::default(),
+/// ).unwrap();
+/// println!("{}x{}, {} bytes", frame.width, frame.height, frame.data.len());
+/// ```
+pub fn decode_frame(
+    data: &[u8],
+    frame_index: usize,
+    pixel_format: JxlPixelFormat,
+    options: JxlDecoderOptions,
+) -> Result<DecodedFrame> {
```
zone117x: Would it be reasonable to have this use a caller provider input buffer? Similar to how the API for the `JxlDecoder<WithFrameInfo>::process(...)` function works:
https://github.com/libjxl/jxl-rs/blob/90a2bb2a31e113b465a1bb40cf3689278e5889f0/jxl/src/api/decoder.rs#L176-L190

This would make the function more memory friendly for various use cases (in particular, avoiding big copies in FFI with caller-managed memory).

jxl/src/api/frame_info.rs
```diff
@@ -0,0 +1,767 @@
...
+/// ).unwrap();
+/// for f in &frames {
+///     println!("frame {}: {}x{}, {:.0}ms", f.info.index, f.width, f.height, f.info.duration_ms);
+/// }
+/// ```
+pub fn decode_all_frames(
+    data: &[u8],
+    pixel_format: JxlPixelFormat,
+    options: JxlDecoderOptions,
+) -> Result<Vec<DecodedFrame>> {
```
zone117x: Similar question here - can the caller provide the data buffer? Assuming the caller has previously called `scan_frame_info()`, the buffer size(s) should be known, right?

jxl/src/api/frame_info.rs
hjanuschka: Done. Added `decode_frame_into(...)` plus `frame_output_layout(...)` so callers can provide their own buffers. `decode_frame` still allocates internally and calls the new API.

jxl/src/api/frame_info.rs
hjanuschka: Done. Added `decode_frame_into(...)` plus `frame_output_layout(...)` so callers can supply their own buffers. The `decode_all_frames` docs now point to using `scan_frame_info` + `decode_frame_into` in a loop for caller-managed memory.

hjanuschka:
@zone117x thanks, good input!

jxl/src/api/inner/mod.rs
```diff
@@ -141,6 +141,20 @@ impl JxlDecoderInner {
...
+    ///
+    /// Preserves image-level state (file header, decoder state including
+    /// reference frames, color profiles, pixel format). Clears frame header,
+    /// TOC, section buffers, and the box parser position so that the next
+    /// `process()` call starts parsing a new frame header.
+    ///
+    /// The caller must provide input starting from the target frame's
+    /// codestream position.
+    pub fn start_new_frame(&mut self) {
+        self.box_parser = BoxParser::new();
```
veluca93: I might be wrong, but I think this logic is broken if the frame starting point is inside a not-until-end-of-file box (i.e. a `jxli` box). The frame could span across two boxes, and I don't believe the box parser will do the right thing in that case.

Perhaps the correct approach here would be to save some box information from the frame scanning API, and pass that information in some way to the main decoder...

jxl/src/api/frame_scan.rs
```diff
@@ -0,0 +1,753 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
```
veluca93: We should not duplicate the parsing logic. Frame scanning should be implemented within the existing decoder logic.

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -82,6 +83,20 @@ pub(super) struct CodestreamParser {
 
     header_needed_bytes: Option<u64>,
 
+    // --- Frame info tracking (for frame scanning) ---
+    /// Collected visible frame info entries.
+    pub(super) scanned_frames: Vec<VisibleFrameInfo>,
+    /// Zero-based visible frame index counter.
+    visible_frame_index: usize,
+    /// Codestream offset of the most recent keyframe (for decode_start_offset).
```
veluca93: These should be _file_ offsets, not _codestream_ offsets, no?

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +156,64 @@ impl CodestreamParser {
...
+            None => return,
+        };
+        let header = frame.header();
+
+        if !header.is_visible() {
+            return;
+        }
+
+        // A visible frame is a keyframe when:
+        // 1. It's the first visible frame, OR
```
veluca93: I don't believe that's true - if this is the first visible frame, it might still depend on a previous patch frame.
Thus, the resume point should be at the start of the first (visible or not) frame, not at the start of the current frame.

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +156,64 @@ impl CodestreamParser {
...
+        // 2. It uses Replace blending for all channels, covers the full
+        //    frame (!needs_blending), and does not use patches.
+        let is_keyframe = if self.visible_frame_index == 0 {
+            true
+        } else {
+            !header.needs_blending() && !header.has_patches()
```
veluca93: Also frames with a LF frame are not keyframes.

jxl/src/api/frame_scan.rs
veluca93, line 1: Shouldn't we remove this file now?

jxl/src/api/decoder.rs
```diff
@@ -1462,6 +1505,106 @@ pub(crate) mod tests {
...
+                ProcessingResult::NeedsMoreInput { fallback, .. } => {
+                    decoder_frame = fallback;
+                }
+            }
+        };
+
+        // Verify pixels were written (not all zeros).
+        let row = color_buffer.row(0);
+        let has_nonzero = row.iter().any(|&v| v != 0.0);
+        assert!(has_nonzero, "decoded frame should have non-zero pixels");
```
veluca93: I think we should instead compare against the corresponding decoded frame in a "regular" sequential decode

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +161,71 @@ impl CodestreamParser {
...
+            self.first_frame_file_offset = Some(self.current_frame_file_offset);
+        }
+
+        if !header.is_visible() {
+            return;
+        }
+
+        // A visible frame is a keyframe when it can be decoded independently:
+        // full-frame Replace blending, no patches, and no LF-frame dependency.
+        let is_keyframe =
```
veluca93: This logic is IMO quite restrictive.

For example, consider an animation composed of a sequence of patch frames alternating with regular frames.

The current code will claim that there are no keyframes, and we always need to decode the file from the start.

Instead, what I think we should do is the following:

1) keep around information about where _each_ frame starts (visible or not)
2) keep track of, for each slot (the LF slots and the reference slots), which frame is the earliest frame that we need to decode to ensure the data in that slot is correct
3) when encountering a new visible frame, we can claim that it is a keyframe if there are no visible frames between the earliest frame it uses and the visible frame itself (to avoid weird ordering, the check of whether a frame is used could look at _all_ the saved slots, except the ones that will be overwritten by the current frame and are not otherwise used by blending and/or patches)

This algorithm should handle the situation I described above well, and should still be correct.

jxl/src/api/decoder.rs
```diff
@@ -35,6 +39,173 @@ pub struct JxlDecoder<State: JxlState> {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct VisibleFrameInfo {
+    /// Zero-based index among visible frames.
+    pub index: usize,
+    /// Duration in milliseconds (0 for still images or the last frame).
+    pub duration_ms: f64,
+    /// Duration in raw ticks from the animation header.
+    pub duration_ticks: u32,
+    /// Byte offset of this frame's header in the input file.
+    pub file_offset: usize,
```
veluca93: Why do we need to expose this?

jxl/src/api/decoder.rs
```diff
@@ -35,6 +39,173 @@ pub struct JxlDecoder<State: JxlState> {
...
+    Initial,
+    Scanning,
+    Done,
+}
+
+/// Incremental frame information scanner for animation support.
+///
+/// Internally, this drives the existing decoder implementation without pixel
+/// output, so container/header/TOC parsing is shared with normal decoding.
+pub struct FrameInfoDecoder {
```
veluca93: Do we need a separate struct for this?

IMO it could be sufficient to add a decode option "scan_frames_only" (or similar), and then using the regular decode API. Of course, based on that option we would change the decoder implementation to effectively skip everything but frame header and TOC parsing.

jxl/src/api/decoder.rs
```diff
@@ -1462,6 +1675,310 @@ pub(crate) mod tests {
         assert!(dec.frame_index().is_none());
     }
 
+    /// Test that `start_new_frame()` + scanner seek info decodes the same
```
veluca93: We should have a test with boxes too.

jxl/src/api/decoder.rs
```diff
@@ -1462,6 +1675,310 @@ pub(crate) mod tests {
...
+    /// frame as regular sequential decode.
+    #[test]
+    fn test_start_new_frame_bare_codestream() {
+        use crate::api::frame_scan::FrameInfoDecoder;
+        use crate::api::{JxlDataFormat, JxlPixelFormat};
+        use crate::image::{Image, Rect};
+
+        fn assert_images_close(expected: &Image<f32>, actual: &Image<f32>) {
```
veluca93: I think this should test for equality, not just closeness.
Moreover, I believe we already have a function that does similar checks in other test files.

jxl/src/api/mod.rs
```diff
@@ -18,6 +18,12 @@ pub use crate::image::JxlOutputBuffer;
 pub use color::*;
 pub use data_types::*;
 pub use decoder::*;
+
+/// Backward-compatible namespace re-export for frame scanning APIs.
```
veluca93: I don't believe we need this, since the code that introduced this module was never submitted.

jxl/src/api/decoder.rs
hjanuschka: good catch changed to pub(crate)

jxl/src/api/decoder.rs
```diff
@@ -35,6 +35,72 @@ pub struct JxlDecoder<State: JxlState> {
...
+    ///
+    /// A visible frame is considered a keyframe when there are no other
+    /// visible frames between it and the earliest frame needed to decode it.
+    /// That earliest frame may still be non-visible.
+    pub is_keyframe: bool,
+    /// Earliest file byte offset needed to decode this frame.
+    ///
+    /// This may point to a non-visible frame if the visible frame depends on
+    /// earlier saved LF/reference state.
+    pub decode_start_file_offset: usize,
```
veluca93: Why don't we replace this and `remaining_in_box` with a `seek_target: VisibleFrameSeekTarget`?
We can pre-compute `visible_frames_to_skip` (and then `is_keyframe` just needs to check that `visible_frames_to_skip` is 0).

jxl/src/api/decoder.rs
```diff
@@ -157,14 +229,79 @@ impl JxlDecoder<WithImageInfo> {
...
+    /// Prepare and apply a seek to `visible_frame_index` from scanned metadata.
+    ///
+    /// This calls [`start_new_frame`](Self::start_new_frame) internally and
+    /// returns the input offset and number of visible frames to skip before
+    /// decoding the requested target frame.
+    ///
+    /// Returns `None` if `visible_frame_index` is out of range.
+    pub fn seek_to_visible_frame(
+        &mut self,
+        scanned_frames: &[VisibleFrameInfo],
```
veluca93: Shouldn't we just pass a VisibleFrameSeekTarget here?

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +173,116 @@ impl CodestreamParser {
...
+        });
+
+        let mut decode_start_frame_index = current_frame_index;
+
+        // Track frame dependencies through reference slots. For blending we know
+        // exactly which slots are used. For patches we conservatively assume any
+        // reference slot may be used.
+        let mut used_reference_slots = [false; NUM_REFERENCE_SLOTS];
+        if header.needs_blending() {
+            let color_source = header.blending_info.source as usize;
```
veluca93:
```
for blending_info in header.ec_blending_info.iter().chain(std::iter::once(&header.blending_info)) { ... }
```
?

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +173,116 @@ impl CodestreamParser {
...
+            if color_source < NUM_REFERENCE_SLOTS {
```
veluca93: I would assert that `source < NUM_REFERENCE_SLOTS`, as otherwise there is an obvious bug in the library.

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +173,116 @@ impl CodestreamParser {
...
+
+        for (slot, used) in used_reference_slots.iter().enumerate() {
+            if *used && let Some(dep_start) = self.reference_slot_decode_start[slot] {
+                decode_start_frame_index = decode_start_frame_index.min(dep_start);
+            }
+        }
+
+        if header.has_lf_frame() {
+            let lf_slot = header.lf_level as usize;
+            if lf_slot < NUM_LF_SLOTS
```
veluca93: Similarly I would assert `lf_slot < NUM_LF_SLOTS`.

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -136,6 +173,116 @@ impl CodestreamParser {
...
+        if header.can_be_referenced {
+            let slot = header.save_as_reference as usize;
+            if slot < NUM_REFERENCE_SLOTS {
+                self.reference_slot_decode_start[slot] = Some(decode_start_frame_index);
+            }
+        }
+
+        if header.lf_level != 0 {
+            let slot = (header.lf_level - 1) as usize;
+            if slot < NUM_LF_SLOTS {
```
veluca93: We should assert here too.

jxl/src/api/inner/box_parser.rs
```diff
@@ -205,9 +216,43 @@ impl BoxParser {
...
+    /// buffers.  The caller must provide raw input starting from the file
+    /// position that corresponds to the target codestream offset.
+    ///
+    /// `remaining` is the number of codestream bytes left in the current
+    /// box from the target file position.  For bare-codestream files this
+    /// is `u64::MAX`.
+    pub(super) fn reset_for_codestream_seek(&mut self, remaining: u64) {
+        self.box_buffer = SmallBuffer::new(128);
+        self.state = ParseState::CodestreamBox(remaining);
+        // Keep frame_index and total_codestream_consumed unchanged.
```
veluca93: Why?

jxl/src/api/inner/box_parser.rs
```diff
@@ -205,9 +216,43 @@ impl BoxParser {
...
+    /// but are still buffered in the codestream parser are NOT included.
+    pub(super) fn remaining_in_codestream_box(&self) -> Option<u64> {
+        match &self.state {
+            ParseState::CodestreamBox(r) => Some(*r),
+            _ => None,
+        }
+    }
+
+    /// Accounts file bytes consumed directly by codestream parser reads/skips.
+    pub(super) fn note_file_consumed(&mut self, amount: usize) {
```
veluca93: I'd call this `mark_`, not `note_`.

jxl/src/api/inner/box_parser.rs
hjanuschka: removed that unnecessary state/comment

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -33,6 +34,16 @@ struct SectionBuffer {
     section: Section,
 }
 
+const NUM_REFERENCE_SLOTS: usize = DecoderState::MAX_STORED_FRAMES;
```
veluca93: Why the alias? Can't we just use `DecoderState::MAX_STORED_FRAMES`?

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -33,6 +34,16 @@ struct SectionBuffer {
...
+const NUM_LF_SLOTS: usize = 4;
```
veluca93: Can we move this to a NUM_LF_FRAMES constant on DecoderState too? (and use it in the `lf_frames` field definition)

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -304,6 +498,39 @@ impl CodestreamParser {
...
+                        .remaining_in_codestream_box()
+                        .map(|r| {
+                            if r > u64::MAX / 2 {
+                                u64::MAX
+                            } else {
+                                r.saturating_add(self.non_section_buf.len() as u64)
+                                    .saturating_add(box_parser.box_buffer.len() as u64)
+                            }
+                        })
+                        .unwrap_or(u64::MAX);
```
veluca93: What happens if the previous frame finished exactly at a box boundary? does remaining_in_codestream_box return Some(0) or None?

jxl/src/api/decoder.rs
```diff
@@ -35,6 +35,54 @@ pub struct JxlDecoder<State: JxlState> {
...
+    pub remaining_in_box: u64,
+    /// Number of visible frames to skip after seek-start before decoding the
+    /// requested target frame.
+    pub visible_frames_to_skip: usize,
+}
+
+/// Compute seek inputs for `visible_frame_index` from scanned frame info.
+///
+/// Returns `None` if `visible_frame_index` is out of range.
+pub fn seek_target_for_visible_frame(
```
veluca93: Should we still have this function? it effectively boils down to doing indexing followed by a member access, so I'd rather omit it.

jxl/src/api/decoder.rs
veluca93, line 93: I think the API here should be adapted to give similar information to what scanned_frames gives (as a follow-up, so let's leave a TODO for now).

jxl/src/api/decoder.rs
```diff
@@ -157,14 +211,78 @@ impl JxlDecoder<WithImageInfo> {
...
     }
 
+    /// Prepare and apply a seek to `visible_frame_index` from scanned metadata.
+    ///
+    /// This calls [`start_new_frame`](Self::start_new_frame) internally and
+    /// returns the input offset and number of visible frames to skip before
+    /// decoding the requested target frame.
+    ///
+    /// Returns `None` if `visible_frame_index` is out of range.
+    pub fn seek_to_visible_frame(
```
veluca93: Why do we need this API call now? The caller can just call `start_new_frame` with the opportune `seek_target` taken from `scanned_frames`.

jxl/src/api/decoder.rs
```diff
@@ -1462,6 +1580,362 @@ pub(crate) mod tests {
         assert!(dec.frame_index().is_none());
     }
 
+    fn scan_frames_with_decoder(
```
veluca93: Seems like this function would be simpler if it didn't return both is_animated and the list of visible frames -- all the usages of is_animated also check whether the frame vec is of length 1 or more.

jxl/src/api/decoder.rs
```diff
@@ -1462,6 +1580,362 @@ pub(crate) mod tests {
...
+    #[test]
+    fn test_start_new_frame_bare_codestream() {
+        let data =
+            std::fs::read("resources/test/conformance_test_images/animation_icos4d_5.jxl").unwrap();
+        assert_start_new_frame_matches_sequential(&data, true);
+    }
+
+    /// Test that `start_new_frame()` + scanner seek info also works for boxed input.
+    #[test]
+    fn test_start_new_frame_boxed_codestream() {
```
veluca93, line 1807: Ideally we should also test this for an image in which, say, each frame is in its own jxlp box.

PR #676, opened 2026-02-06
Progressive jxl_cli

jonsneyers:
Decoder changes for progressive modular: https://github.com/libjxl/jxl-rs/pull/680

This PR is now just the changes to jxl_cli to add an option `--progressive-step-size` that does streaming decode with regular flush_pixels().

PR #679, opened 2026-02-09
Better progressive decoding CLI options.

veluca93:
IMO better approach than the corresponding code in #676 .

### DIFF
--- a/jxl_cli/src/dec/mod.rs
+++ b/jxl_cli/src/dec/mod.rs
@@ -4,6 +4,7 @@
 // license that can be found in the LICENSE file.
 
 use std::{
+    io::BufReader,
     str::FromStr,
     time::{Duration, Instant},
 };
@@ -20,6 +21,7 @@ use jxl::{
 };
 
 pub struct ImageFrame {
+    pub partial_renders: Vec<Vec<OwnedRawImage>>,
     pub channels: Vec<OwnedRawImage>,
     pub duration: f64,
     pub color_type: JxlColorType,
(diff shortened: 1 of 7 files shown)


jxl_cli/src/dec/mod.rs
```diff
@@ -217,34 +248,56 @@ pub fn decode_frames<In: JxlBitstreamInput>(
...
+                    break 'partial result;
                 }
-                return Err(eyre!("Source file truncated"));
-            }
+                ProcessingResult::NeedsMoreInput { mut fallback, .. } => {
+                    // If we have more data but we're feeding it slowly, save the partial
+                    // render and retry.
+                    if render_interval.is_some() && input.available_bytes()? > 0 {
+                        fallback.flush_pixels(&mut output_bufs)?;
+                        partial_renders.push(
```
jonsneyers, line 276: It may not be a good idea to keep all the partial renders in memory and only when decoding is done write them all out to files. If you try to do a large image in tiny steps...

tools/progressive_decoding_demo.sh
```diff
@@ -44,21 +49,10 @@ echo "Input file: $INPUT"
...
-  SIZE=$((FILE_SIZE * PERCENT / 100))
-  if [ "$SIZE" -lt 1 ]; then SIZE=1; fi
-
-  echo "Processing $PERCENT% ($SIZE bytes)..."
-
-  TRUNCATED_FILE="$OUTPUT_DIR/truncated_$PERCENT.jxl"
-  OUTPUT_IMG="$OUTPUT_DIR/frame_$(printf "%03d" $i).png"
-
-  head -c "$SIZE" "$INPUT" >"$TRUNCATED_FILE"
+CHUNK_SIZE=$((FILE_SIZE * PERCENT / 100))
```
jonsneyers: Maybe do this with `bc` so it also works for fractional numbers? 0.5% could be an interesting step size once we get progressive LF frames to render :)

jxl_cli/src/dec/mod.rs
veluca93: I consider this mostly a debugging utility, so I don't mind if some combinations require potentially a lot of resources.

tools/progressive_decoding_demo.sh
veluca93: Done.

PR #682, opened 2026-02-22
Working implementation of progressive modular decoding

veluca93:
Includes a good amount of cleanup, some tests and refactors, and some fixes to other cases of incremental decoding.

Closes #680 
Closes #681

### DIFF
--- a/jxl/src/frame/mod.rs
+++ b/jxl/src/frame/mod.rs
@@ -165,9 +165,11 @@ pub struct HfMetadata {
     used_hf_types: u32,
 }
 
-#[derive(PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum RenderUnit {
+    /// VarDCT data
     VarDCT,
+    /// Modular channel with the given index
     Modular(usize),
 }
 
--- a/jxl/src/frame/modular/mod.rs
+++ b/jxl/src/frame/modular/mod.rs
@@ -17,7 +17,6 @@ use crate::{
     frame::{
         ColorCorrelationParams, HfMetadata,
         block_context_map::BlockContextMap,
-        modular::transforms::apply::RunKind,
         quantizer::{self, LfQuantFactors, QuantizerParams},
     },
     headers::{
@@ -267,7 +266,7 @@ impl ModularBuffer {
             return;
         }
         let _ = self.remaining_uses.fetch_update(
-            Ordering::Relaxed,
+            Ordering::Release,
             Ordering::Acquire,
             |remaining_pre: usize| {
                 let remaining = remaining_pre.checked_sub(1).unwrap();
@@ -374,13 +373,13 @@ pub struct FullModularImage {
     // Buffers to _start rendering from_ on the next call to process_output.
     // This is initially set to LF global and LF buffers, and populated with HF buffers
     // just before we start decoding them.
+    ready_buffers_dry_run: BTreeSet<(usize, usize)>,
     ready_buffers: BTreeSet<(usize, usize)>,
 }
 
 impl FullModularImage {
     pub fn can_do_partial_render(&self) -> bool {
-        // XXX: FIX
-        self.can_do_partial_render && false
+        self.can_do_partial_render
     }
 
     #[instrument(level = "debug", skip_all)]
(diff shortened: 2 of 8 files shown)


jxl/src/frame/modular/transforms/apply.rs
```diff
@@ -56,40 +56,44 @@ pub enum TransformStep {
...
     // Grid position this transform should produce.
     // Note that this is a lie for Palette with AverageAll or Weighted, as the transform with
     // position (0, y) will produce the entire row of blocks (*, y) (and there will be no
     // transforms with position (x, y) with x > 0).
     pub(super) grid_pos: (usize, usize),
-    // Number of inputs that are not yet available.
-    pub(super) incomplete_deps: usize,
+
+    // List of (buffer, grid) that this transform depends on.
+    pub(in super::super) deps: Vec<(usize, usize)>,
```
tirr-c, line 67: I wonder if this field could benefit from using `SmallVec`?

jxl/src/frame/modular/transforms/apply.rs
```diff
@@ -390,11 +385,34 @@ impl TransformStepChunk {
...
+            TransformStep::Palette {
+                buf_in,
+                buf_out,
+                predictor,
+                ..
+            } if buffers[*buf_in].info.size.0 != 0 && predictor.requires_full_row() => {
+                0..buffers[buf_out[0]].grid_shape.0
+            }
+            _ => 0..1,
+        };
```
tirr-c: This could be just an integer, so that `.clone()` below could be removed.

jxl/src/frame/mod.rs
```diff
@@ -165,6 +165,12 @@ pub struct HfMetadata {
     used_hf_types: u32,
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum RenderUnit {
+    VarDCT,
+    Modular(usize),
```
tirr-c, line 173: A doc comment to clarify the meaning of `usize` would be nice.

jxl/src/frame/modular/mod.rs
```diff
@@ -162,49 +170,112 @@ impl ModularChannel {
...
-                .map(ModularChannel::try_clone)
-                .transpose()?
-                .unwrap())
+    fn get_buffer(&self, can_consume: bool) -> Result<ModularChannel> {
+        if !can_consume {
+            return ModularChannel::try_clone(self.data.borrow().as_ref().unwrap());
         }
+        let mut ret = None;
+        let mut remaining_pre = self.remaining_uses.load(Ordering::Acquire);
+        loop {
```
tirr-c, line 233: Maybe we can also use `.fetch_update()` here?

jxl/src/frame/modular/transforms/apply.rs
veluca93: I thought about it, but decided that it would probably not be beneficial because the vec is created just once per transform chunk, unlike other uses of smallvec which are once per image row per channel per group per stage.

jxl/src/frame/modular/mod.rs
veluca93: I thought about it, but given that it contains a fallible operation I thought it would end up being more complicated that way. However, after trying it out it seems to be just fine also with fetch_update, so ... :-)

jxl/src/frame/modular/mod.rs
```diff
@@ -162,49 +170,112 @@ impl ModularChannel {
...
+                    );
+                }
+            } else if remaining == 0 {
+                *self.data.borrow_mut() = None;
+            }
+            match self.remaining_uses.compare_exchange_weak(
+                remaining_pre,
+                remaining,
+                Ordering::Release,
+                Ordering::Acquire,
```
tirr-c, line 255: Not sure about this, but I think synchronization is done by `AtomicRefCell`?

veluca93: The synchronization here is to avoid the following scenario:

- Thread 1 decrements remaining, it is now 1
- Thread 2 decrements remaining, it is now 0
- Thread 2 takes away the data
- Thread 1 tries to copy the data -> boom

The way the code is currently written, if there are no bugs in the rest of the code, it is guaranteed that if we observe `remaining` to be `0` then nobody else will be trying to access the data

PR #684, opened 2026-02-22
SIMD-store_u8 and clean up safety comments.

veluca93:
Also fixes some potential writes-through-unaligned-pointers, and consistently uses `.cast()` instead of `as *const` / `as *mut`.

### DIFF
--- a/jxl_simd/src/aarch64/neon.rs
+++ b/jxl_simd/src/aarch64/neon.rs
@@ -126,7 +126,7 @@ unsafe impl F32SimdVec for F32VecNeon {
     fn load(d: Self::Descriptor, mem: &[f32]) -> Self {
         assert!(mem.len() >= Self::LEN);
         // SAFETY: we just checked that `mem` has enough space. Moreover, we know neon is available
-        // from the safety invariant on `d`.
+        // from the safety invariant on `d`. vld1q_f32 supports unaligned loads.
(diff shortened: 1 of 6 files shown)


jxl_simd/src/aarch64/neon.rs
```diff
@@ -699,13 +701,28 @@ impl I32SimdVec for I32VecNeon {
...
+    #[inline(always)]
+    fn store_u8(self, dest: &mut [u8]) {
+        assert!(dest.len() >= Self::LEN);
+        // SAFETY: We know neon is available from the safety invariant on `self.1`,
+        // and we just checked that `dest` has enough space. vst1_lane_u32 supports unaligned stores.
+        unsafe {
+            // vmovn narrows i32 -> i16 -> i8
+            let narrowed_i16 = vmovn_s32(self.0);
+            let combined_i16 = vcombine_s16(narrowed_i16, narrowed_i16);
+            let narrowed_i8 = vmovn_s16(combined_i16);
```
tirr-c, line 722: I think we can do `narrowed_i8_i16 = vreinterpret_s8_s16(narrowed_i16); narrowed_i8 = vuzp1_s8(narrowed_i8_i16, narrowed_i8_i16)`, which should spare some instructions. Not so sure if this will be faster though...

veluca93: I would say that we can leave this kind of micro-optimizations to later :-)

PR #685, opened 2026-02-23
Various speedups for modular, especially -d0 -I0 -P0

### LINKED ISSUE #622: slow to decode d0P0I0
[d0P0I0_testcase.zip](https://github.com/user-attachments/files/24612227/d0P0I0_testcase.zip)
(libjxl 53042e)  cjxl -d 0 -P 0 -I 0

**69.516 MP/s**  (jxl-rs 249779) jxl_cli -s --num-reps 41
**495.818 MP/s** (libjxl 53042e) djxl --disable_output --num_threads 0 --num_reps 41

as reported by each software after 5x 1-rep warmups on an i5-13600K

veluca93:
Fixes #622.
Depends on https://github.com/libjxl/jxl-rs/pull/684.

On my machine, `inc_pat 102 2020-09-10 11-24-46.png.cjxl_0.12.0-53042e_d0P0I0.jxl` went from ~55MP/s to ~305MP/s.
`djxl` averages to ~230MP/s, so I'll consider it a win :-)

### DIFF
--- a/jxl/src/render/builder.rs
+++ b/jxl/src/render/builder.rs
@@ -6,8 +6,10 @@
 use crate::api::{JxlColorType, JxlDataFormat};
 use crate::error::{Error, Result};
 use crate::headers::Orientation;
+use crate::render::StageSpecialCase;
 use crate::render::internal::ChannelInfo;
 use crate::render::save::SaveStage;
+use crate::render::stages::ConvertI32ToU8Stage;
 use crate::util::{ShiftRightCeil, tracing_wrappers::*};
 
 use super::internal::{RenderPipelineShared, Stage};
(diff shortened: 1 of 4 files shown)


jxl/src/entropy_coding/decode.rs
```diff
@@ -259,7 +260,7 @@ impl SymbolReader {
                     min_symbol,
                     min_length,
                     dist_multiplier,
-                    window: Vec::new(),
+                    window: Vec::new_with_capacity(1 << Lz77State::LOG_WINDOW_SIZE)?,
```
jonsneyers, line 263: Once we add MT decode, we should check if this 4 MB allocation isn't causing memory usage to go up too much (on platforms where allocation does actually reserve physical memory; I suppose on Linux uninitialized allocs are essentially free).

In the typical case the window doesn't have to get larger than 256 KB so that might be a safer initial capacity.

veluca93: If this becomes a problem, there are other approaches that we can follow here (although IIRC libjxl currently does exactly this, and as far as I can tell it does not cause issues).

For now, I would just leave it as is.

jonnyawsom3:
Should also fix #546

PR #686, opened 2026-02-23
Fix OOM in FrameIndexBox::parse from crafted NF value

hjanuschka:
The ClusterFuzzLite batch fuzzer found that a crafted `jxli` box can claim billions of frame index entries (NF up to `u32::MAX`), causing `Vec::with_capacity(nf)` to attempt a multi-GB allocation and abort.

**Fix:** Cap the pre-allocation to the maximum number of entries that could actually fit in the remaining data. Each entry requires at least 3 bytes (three LEB128 varints, minimum 1 byte each), so the capacity is bounded by `reader.len() / 3`.

The loop still iterates `nf` times, but `read_varint_from_reader` returns an error when the data runs out, so we fail gracefully instead of OOMing.

**Found by:** ClusterFuzzLite batch fuzzing (`decode_header` target) - https://github.com/libjxl/jxl-rs/actions/runs/22288685274/job/64471888189

Includes a regression test (`test_huge_nf_no_oom`).

### DIFF
--- a/jxl/src/container/frame_index.rs
+++ b/jxl/src/container/frame_index.rs
@@ -16,6 +16,7 @@ use byteorder::{BigEndian, ReadBytesExt};
 
 use crate::error::{Error, Result};
 use crate::icc::read_varint_from_reader;
+use crate::util::NewWithCapacity;
 
 /// A single entry in the frame index.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


jxl/src/container/frame_index.rs
```diff
@@ -87,7 +87,9 @@ impl FrameIndexBox {
         )
         .ok_or(Error::InvalidBox)?;
 
-        let mut entries = Vec::with_capacity(nf);
+        // Each entry requires at least 3 bytes (three varints, min 1 byte each).
+        // Cap the pre-allocation to avoid OOM from a crafted NF value.
+        let mut entries = Vec::with_capacity(nf.min(reader.len() / 3));
```
veluca93: Could also use `Vec::new_with_capacity` from utils which gives a `Err()` in case of OOM :-)

veluca93 (approved):

PR #689, opened 2026-02-24
Progressive rendering of LF frames

veluca93:
Also allow flushing before having received the first visible frame.

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -160,6 +160,9 @@ impl Frame {
         } else {
             decoder_state.nonvisible_frame_index += 1;
         }
+        if frame_header.frame_type == FrameType::LFFrame && frame_header.lf_level == 1 {
+            decoder_state.lf_frame_was_rendered = false;
+        }
         let image_metadata = &decoder_state.file_header.image_metadata;
         let is_gray = !frame_header.do_ycbcr
             && !image_metadata.xyb_encoded
@@ -266,10 +269,11 @@ impl Frame {
             return false;
         }
 
-        if self.header.frame_type != FrameType::RegularFrame {
-            return false;
-        }
-        true
+        self.header.frame_type == FrameType::RegularFrame
+            || (self.header.frame_type == FrameType::LFFrame
+                && self.header.lf_level == 1
+                // TODO(veluca): this should probably be "there is no alpha".
+                && self.header.num_extra_channels == 0)
     }
 
     /// Given a bit reader pointing at the end of the TOC, returns a vector of `BitReader`s, each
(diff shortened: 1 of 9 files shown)


jxl/src/frame/decode.rs
```diff
@@ -266,10 +269,11 @@ impl Frame {
...
 
-        if self.header.frame_type != FrameType::RegularFrame {
-            return false;
-        }
-        true
+        self.header.frame_type == FrameType::RegularFrame
+            || (self.header.frame_type == FrameType::LFFrame
+                && self.header.lf_level == 1
+                // TODO(veluca): this should probably be "there is no alpha".
+                && self.header.num_extra_channels == 0)
```
jonsneyers, line 276: What happens when dropping this condition? A useless render with alpha all-zero?

In libjxl, we currently put all-zeroes extrachannel data in LfFrames (see https://github.com/libjxl/libjxl/blob/032d39a2125f7d1d28b6cee7914888a489bc875f/lib/jxl/enc_cache.cc#L173) which would make a preview based on just the LfFrame fully invisible. But in principle we could put a downscaled/lossy version of the alpha data there too, at the cost of extra filesize since this would be redundant data (only the color part of the LfFrame actually gets used in the regular frame). Or an all-ones alpha channel, which is as cheap as all-zeroes, but of course it would cause the first previews to have all the invisible parts visible so I wouldn't do that.

I would remove this check and render the extra channels of an LfFrame the way they are — which is not useful in current cjxl output (the previews are invisible so it's a waste of time), but it would leave the door open for doing something better encode-side. If you want to avoid the waste of time, probably it's possible to detect the "alpha will be all-zeroes" case and early-return when flushing, but that can be left as an optimization for later.

veluca93: At the moment the decoder does not keep around the extra channels of LF frames at all, and changing that would complicate the code (and complicated the upsampling logic here somewhat), so I prefer to keep it as is for now.

jonnyawsom3: I previously brought it up for Oxide, as with how often images have redundant alpha, nearly every image I tried had black/blank LF sections. It got changed to assume alpha is full instead of empty in https://github.com/tirr-c/jxl-oxide/pull/466, which has the risk of showing background colors/lossy optimization, ect but is better than nothing at all. Using `-p` with cjxl already squeezes the Alpha too, so when LfGlobal is handled there should be data available, even if it's slightly lower resolution.

Another option is instead of zero filling or putting all-ones into a redundant LF extra channel, we could make Alpha progressive/squeezed by default too. Then it would be available in LfGlobal before LfGroups. A quick test shows a density improvement, but encode speed and memory usage take a hit from requiring buffering... Might be worth looking into later.

jonsneyers: For non-LF-frame cases there is no issue with progressive alpha, just in case of LF frames the first alpha is only available in the main (regular) frame, at least with the jxl files libjxl currently produces. It produces LF frames with all-zero alpha since it's not used for the final image: the main frame can only use the color LF from the LF frame, not extrachannels.
The LF frame can contain actual alpha though. But there's currently no encoder that does that.

So this is not very urgent, to make LF frames + alpha work progressively before the full LF frame is available, first we'd need to change the encoder to actually put some useful alphah in the LF frame.

PR #694, opened 2026-03-02
chore: move cms impl

mertalev:
The JxlCms implementation in the CLI is useful for downstream library usage as well. This PR moves it to a new `jxl_cms` crate.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,3 +32,4 @@ Sami Boukortt
 Tomáš Král <tomas@kral.hk>
 Wonwoo Choi <chwo9843@gmail.com>
 Zoltan Szabadka
+Mert Alev <101130780+mertalev@users.noreply.github.com>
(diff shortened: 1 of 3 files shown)


veluca93:
Sorry for the delay!

I am not sure I like using a feature flag in the `jxl` crate -- what do you think of having a separate `jxl-cms` (or similar) library? That follows the approach that libjxl uses.

mertalev:
> Sorry for the delay!
> 
> 
> 
> I am not sure I like using a feature flag in the `jxl` crate -- what do you think of having a separate `jxl-cms` (or similar) library? That follows the approach that libjxl uses.

Thanks for the follow up! The main downside is that of the crates here, only `jxl` is published in crates.io. This makes the other crates inconvenient to use since it needs to be a git dependency. I think a `jxl_cms` crate would be fine if it's also published.

veluca93:
> > Sorry for the delay!
> > I am not sure I like using a feature flag in the `jxl` crate -- what do you think of having a separate `jxl-cms` (or similar) library? That follows the approach that libjxl uses.
> 
> Thanks for the follow up! The main downside is that of the crates here, only `jxl` is published in crates.io. This makes the other crates inconvenient to use since it needs to be a git dependency. I think a `jxl_cms` crate would be fine if it's also published.

Yes, the idea would be to also publish the `jxl_cms` crate.

PR #697, opened 2026-03-07
Make progressive decoding more progressive

veluca93:
In particular, this allows partial renders of the LfGlobal section, which enables the first progressive stages.

### DIFF
--- a/jxl/src/frame/modular/mod.rs
+++ b/jxl/src/frame/modular/mod.rs
@@ -358,6 +358,9 @@ pub struct FullModularImage {
     modular_color_channels: usize,
     can_do_partial_render: bool,
     can_do_early_partial_render: bool,
+    decoded_section0_channels: usize,
+    needed_section0_channels_for_early_render: usize,
+    global_header: Option<GroupHeader>,
     buffers_for_channels: Vec<usize>,
     // Buffers to _start rendering from_ on the next call to process_output.
     // This is initially set to LF global and LF buffers, and populated with HF buffers
@@ -375,12 +378,11 @@ impl FullModularImage {
 
     pub fn can_do_early_partial_render(&self) -> bool {
         self.can_do_early_partial_render
+            // Avoid green martians
+            && self.decoded_section0_channels >= self.needed_section0_channels_for_early_render
     }
 
     pub fn set_pipeline_used_channels(&mut self, used: &[bool]) {
-        if !self.pipeline_used_channels.is_empty() {
-            return;
-        }
         self.pipeline_used_channels = used.to_vec();
     }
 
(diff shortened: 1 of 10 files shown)


veluca93:
> I noticed that `JxlDecoder::<WithFrameInfo>::flush_pixels` panics if it is called when full frame data is already available.

Now it shouldn't panic anymore, please check :-)

jxl/src/frame/modular/mod.rs
```diff
@@ -372,10 +376,13 @@ impl FullModularImage {
         self.can_do_partial_render
     }
 
+    pub fn can_do_early_partial_render(&self) -> bool {
+        self.can_do_early_partial_render
+            // Avoid green martians
```
jonsneyers, line 381: It would be a pretty blurry green martian, at 8x8 pixels, and I'm not sure if it would be green in XYB :)

jxl/src/frame/modular/mod.rs
```diff
@@ -579,35 +594,87 @@ impl FullModularImage {
...
+
+        for b in self.section_buffer_indices[0]
+            .iter()
+            .take(self.decoded_section0_channels)
+        {
+            if self.buffer_info[*b].buffer_grid[0].get_status() == BUFFER_STATUS_FINAL_RENDER {
+                continue;
+            }
+            // If we did a partial decode, we cannot be 100% sure of whether we correctly
+            // decoded all the sections. Thus, mark the sections as partially decoded.
```
jonsneyers, line 666: Can't we distinguish here between sections that were partially decoded and ones that were complete?

jxl/src/render/stages/splines.rs
```diff
@@ -47,17 +51,36 @@ impl RenderPipelineInPlaceStage for SplinesStage {
         position: (usize, usize),
         xsize: usize,
         row: &mut [&mut [f32]],
-        _state: Option<&mut dyn std::any::Any>,
+        _state: Option<&mut dyn Any>,
     ) {
-        self.splines.draw_segments(row, position, xsize);
+        // TODO(veluca): this is wrong!! Race condition in MT.
```
jonsneyers, line 56: Is this a new problem caused by this PR, or was it already there?

jxl/src/frame/modular/mod.rs
veluca93: I *thought* so, but all the attemtps I made ended up with the final pixels being slightly off...

Given that this only applies to LFGlobal anyway, I don't really see the harm in being a bit conservative -- the worst that can happen is not freeing some smallish buffers early.

jxl/src/render/stages/splines.rs
veluca93: Actually a new problem :-)

PR #706, opened 2026-03-14
feat: simd128 support for wasm32

mertalev:
### Description

This PR adds simd128 support for jxl-rs when targeting wasm32. It does not include lcms2 integration for Wasm since this requires build toolchain changes and increases the scope of the PR.

In a local benchmark on an M4 Max, performance increased by 2.49x (8.9 to 22.1 MP/s), 2.02x (11.1 to 22.5 MP/s) and 1.35x (34.5 to 46.6 MP/s), respectively, for the `progressive.jxl`, `bike.jxl` and `cafe.jxl` test images.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.wasm32-wasip1]
+runner = "wasmtime"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,3 @@
 [target.wasm32-wasip1]
-runner = "wasmtime --dir=. --dir=.."
+runner = "wasmtime --dir=. --dir=.. --env ARBTEST_BUDGET_MS"
 rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,3 @@
 [target.wasm32-wasip1]
-runner = "wasmtime"
+runner = "wasmtime --dir=."
 rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,3 @@
 [target.wasm32-wasip1]
-runner = "wasmtime --dir=."
+runner = "wasmtime --dir=. --dir=.."
 rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,3 @@
 [target.wasm32-wasip1]
 runner = "wasmtime"
+rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -100,6 +100,9 @@ jobs:
             simd: avx512
           - os: ubuntu-24.04-arm
             simd: neon
+          - os: ubuntu-latest
+            simd: simd128
+            target: wasm32-wasip1
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -108,16 +111,25 @@ jobs:
 
       - name: Install latest rust toolchain
         uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Install wasmtime
+        if: ${{ matrix.target == 'wasm32-wasip1' }}
+        uses: bytecodealliance/actions/wasmtime/setup@v1
 
       - name: Rust cache
         uses: Swatinem/rust-cache@v2.8.0
         with:
           prefix-key: ${{ matrix.os }}-test-${{ matrix.simd }}
 
-
       - name: Tests with SIMD feature ${{ matrix.simd }}
-        if: ${{ matrix.simd != 'none' }}
+        if: ${{ matrix.simd != 'none' && !matrix.target }}
         run: cargo test --release --all --no-fail-fast --no-default-features --features ${{ matrix.simd }}
+
+      - name: Tests with SIMD feature ${{ matrix.simd }} (wasm)
+        if: ${{ matrix.target == 'wasm32-wasip1' }}
+        run: cargo test --release -p jxl_simd --no-fail-fast --no-default-features --features ${{ matrix.simd }} --target ${{ matrix.target }}
 
       - name: Tests with no features
         if: ${{ matrix.simd == 'none' }}
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -129,7 +129,11 @@ jobs:
 
       - name: Tests with SIMD feature ${{ matrix.simd }} (wasm)
         if: ${{ matrix.target == 'wasm32-wasip1' }}
-        run: cargo test --release -p jxl_simd --no-fail-fast --no-default-features --features ${{ matrix.simd }} --target ${{ matrix.target }}
+        run: |
+          cargo test --release --workspace --exclude jxl_cli --exclude jxl_cms --no-fail-fast --no-default-features --features ${{ matrix.simd }} --target ${{ matrix.target }} --
+          --skip tirr_photo # 4GiB wasm32 memory limit
+          --skip huge_image # 4GiB wasm32 memory limit
+          --skip fuzzer_smallbuffer # only panic=abort is supported in stable toolchain
 
       - name: Tests with no features
         if: ${{ matrix.simd == 'none' }}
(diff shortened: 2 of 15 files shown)


jxl_simd/src/wasm32/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+        $dname:ident,
+        $descr:ident: $descr_ty:ident,
+        $(#[$($attr:meta)*])*
+        $pub:vis fn $name:ident($($arg:ident: $ty:ty),* $(,)?) $(-> $ret:ty )? $body: block
+    ) => {
+        #[inline(always)]
+        $(#[$($attr)*])*
+        $pub fn $name<$descr_ty: $crate::SimdDescriptor>($descr: $descr_ty, $($arg: $ty),*) $(-> $ret)? $body
+        $(#[$($attr)*])*
+        $pub fn $dname($($arg: $ty),*) $(-> $ret)? {
```
veluca93, line 21: Given that simd128 must be enabled globally in wasm, I am not sure we need the dispatch logic here - we might as well call the simd128 version with cfg(simd128) and the scalar version otherwise.

jxl_simd/src/wasm32/simd128.rs
```diff
@@ -0,0 +1,1105 @@
...
+    fn maybe_downgrade_128bit(self) -> Self {
+        self
+    }
+
+    fn call<R>(self, f: impl FnOnce(Self) -> R) -> R {
+        f(self)
+    }
+}
+
+macro_rules! fn_simd128 {
```
veluca93: again, since simd128 can only be enabled globally, I don't think we need this macro.

.github/workflows/pull_request.yml
```diff
@@ -108,17 +111,26 @@ jobs:
...
 
-
       - name: Tests with SIMD feature ${{ matrix.simd }}
-        if: ${{ matrix.simd != 'none' }}
+        if: ${{ matrix.simd != 'none' && !matrix.target }}
         run: cargo test --release --all --no-fail-fast --no-default-features --features ${{ matrix.simd }}
 
+      - name: Tests with SIMD feature ${{ matrix.simd }} (wasm)
+        if: ${{ matrix.target == 'wasm32-wasip1' }}
+        run: cargo test --release -p jxl_simd --no-fail-fast --no-default-features --features ${{ matrix.simd }} --target ${{ matrix.target }}
```
veluca93: Why does this case only run the jxl_simd tests?

jxl_simd/src/wasm32/simd128.rs
```diff
@@ -0,0 +1,1105 @@
...
+        let a = i64x2_shuffle::<0, 2>(ac02, ac13); // [a0, a1, a2, a3]
+        let b = i64x2_shuffle::<0, 2>(bd02, bd13); // [b0, b1, b2, b3]
+        let c = i64x2_shuffle::<1, 3>(ac02, ac13); // [c0, c1, c2, c3]
+        let dd = i64x2_shuffle::<1, 3>(bd02, bd13); // [d0, d1, d2, d3]
+        (Self(a, d), Self(b, d), Self(c, d), Self(dd, d))
+    }
+
+    #[inline(always)]
+    fn transpose_square(d: Simd128Descriptor, data: &mut [[f32; 4]], stride: usize) {
+        if stride == 1 {
```
veluca93: Do we really benefit from the special case here?

.github/workflows/pull_request.yml
mertalev: jxl_cli and jxl_cms don't compile because of the lcms2 dependency. The other crates do work though, so I've made it only exclude those two.

jxl_simd/src/wasm32/simd128.rs
mertalev: Not really - it ends up being the same load instruction either way

.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[target.wasm32-wasip1]
+runner = "wasmtime --dir=. --dir=.."
+rustflags = ["-C", "target-feature=+simd128,+relaxed-simd"]
```
veluca93, line 3: Is there a way to test both the relaxed-simd and the non-relaxed-simd paths?

jxl_simd/src/wasm32/simd128.rs
```diff
@@ -0,0 +1,1081 @@
...
+                dest.as_mut_ptr().cast::<u8>(),
+                8,
+            );
+        }
+    }
+
+    #[inline(always)]
+    fn store_f16_bits(self, dest: &mut [u16]) {
+        assert!(dest.len() >= F32VecSimd128::LEN);
+        // WASM SIMD128 has no hardware f16 conversion; use scalar fallback.
```
veluca93: We could in theory build our own SIMD conversion via a sequence of bit-shifts (this is also true of sse4). Maybe we should leave a todo here.

mertalev: I did my best to get this right, but I'm not especially confident about it. It does have a very healthy improvement over scalar of 6.6x for store and 3.7x for load.

veluca93: Can you use `arbtest` to add a test for arbitrary f16 conversions and compare SIMD with scalar?

.cargo/config.toml
veluca93: Hey @mertalev - any updates here?

mertalev: As it happens, I'm working on this branch right now! arbtest found a rounding bug in the scalar f16 code so I'm working through that. I can update the testing to test with and without relaxed-simd after that

jxl_simd/src/wasm32/simd128.rs
mertalev: Done - it found an issue with both the scalar and Wasm versions compared to NEON that I fixed. I ran arbtest for 2 minutes each for load and store comparing scalar vs Wasm and scalar vs NEON and it passes, so I'm more confident that it behaves correctly.

.cargo/config.toml
mertalev: CI now tests both `+simd128` and `+simd128,+relaxed-simd`

veluca93:
Hey @mertalev, for whatever reason auto-merge didn't merge this PR and now it's a bit stale - do you have plans to rebase it on top of latest main?

mertalev:
No worries! I rebased it but it seems like it needs a bit more work due to trait changes - I'll get those sorted out later today.

jxl_simd/src/x86_64/avx.rs
```diff
@@ -680,7 +680,7 @@ impl F32SimdVec for F32VecAvx {
         #[inline]
         fn store_f16_bits_impl(v: __m256, dest: &mut [u16]) {
             assert!(dest.len() >= F32VecAvx::LEN);
-            let bits = _mm256_cvtps_ph::<{ _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC }>(v);
+            let bits = _mm256_cvtps_ph::<{ _MM_FROUND_TO_NEAREST_INT }>(v);
```
veluca93: Why this change? (also below)

veluca93 (approved):

veluca93:
### Performance Summary (Commit `b41e0c2`)

| desktop | Single | 79.56 | 80.14 | +1.31% ± 0.37% |
| desktop | Multi | 79.83 | 80.25 | +0.42% ± 0.39% |
| framework-desktop | Single | 96.62 | 97.10 | +1.68% ± 0.33% |
| framework-desktop | Multi | 94.95 | 97.75 | +2.07% ± 0.36% |
| pixel7a | Single (Fast) | 30.62 | 30.94 | +3.59% ± 3.65% |
| pixel7a | Single (Mid) | 23.07 | 22.92 | -0.17% ± 0.32% |
| pixel7a | Multi | 30.52 | 31.17 | +2.50% ± 0.95% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_706.html)

jxl_simd/src/x86_64/avx.rs
mertalev: I remember getting an error here unless I removed that flag.

veluca93: That is surprising to me, is it still true now if you revert the change?

mertalev: Yup, it still gets the error if you look at the CI. I don't have a good sense of *why*, but I guess the compiler sees an issue with it with the added tests / code paths where it didn't before?

veluca93: That appears to be a rustc doc bug, from what I can tell - the Intel docs claim that the function should be used without `_MM_FROUND_NO_EXC` (and I can reproduce the compilation failure on the playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c9caebc90afe9ef13c9ce3144510c6d2). Can you remove it again then?

PR #708, opened 2026-03-16
perf: optimize noise hot path and inline RNG helpers

hjanuschka:
This speeds up noise rendering by removing per-element u64 splitting branches and reusing a direct u32 view of the RNG batch. It also inlines Noise::strength and Xorshift128Plus::fill in hot call paths. Unsafe is used only for a temporary [u64; N] to [u32; 2N] reinterpret cast, which is safe because the layouts are identical and the data is read-only.

### DIFF
--- a/jxl/src/features/noise.rs
+++ b/jxl/src/features/noise.rs
@@ -17,6 +17,7 @@ impl Noise {
         }
         Ok(noise)
     }
+    #[inline(always)]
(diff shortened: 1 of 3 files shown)


veluca93 (commented):
Noise is not used by most images, and a much more helpful speedup would be to SIMDfy the noise generation. I think we should close this for now, and get back to it with a SIMD implementation (or ignore it until noise-enabled images are more prevalent)

PR #710, opened 2026-03-16
perf: use StackVec in render channel plumbing

hjanuschka:
This introduces a fixed-capacity StackVec and uses it in render channel/row hot paths instead of SmallVec. It removes stack-vs-heap branch overhead in frequently executed indexing and row collection paths while keeping behavior unchanged. Unsafe in StackVec is only for MaybeUninit element access and is safe because reads are restricted to initialized elements in 0..len and writes respect capacity invariants.


veluca93:
I think this needs some extra benchmarking - started a run.

If this is beneficial, then I think I'd rather modify smallvec than create a new smallvec-like type.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `ca00721`)

| desktop | Single | 80.94 | 82.23 | +1.48% ± 0.39% |
| desktop | Multi | 80.76 | 81.78 | +1.56% ± 0.32% |
| framework-desktop | Single | 97.09 | 97.90 | +1.52% ± 0.26% |
| framework-desktop | Multi | 96.16 | 94.45 | +1.02% ± 0.35% |
| pixel7a | Single (Fast) | 30.96 | 31.79 | +11.61% ± 9.34% |
| pixel7a | Single (Mid) | 22.93 | 23.33 | +0.74% ± 1.36% |
| pixel7a | Multi | 31.58 | 31.51 | +0.79% ± 0.39% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_710.html)

PR #711, opened 2026-03-16
perf: speed up distinct index collection in low-memory pipeline

hjanuschka:
This removes Option staging in get_distinct_indices and builds the result directly in position order. It reduces overhead in a hot helper used during low-memory pipeline row setup without changing behavior. Unsafe is used only to reborrow previously collected mutable pointers, and is safe because indices are distinct and sorted, so aliasing cannot occur.


veluca93 (commented):
Please (1) add an arbtest that checks that the code is fine (2) run this test under miri

veluca93:
This also needs a benchmark, enqueued it :-)

PR #712, opened 2026-03-16
perf: avoid zero-initializing modular channel payloads

hjanuschka:
This adds an uninitialized image constructor that zeros only offset/padding regions, then uses it for modular channel allocation. It removes large buffer zeroing work in hot decode paths while preserving prediction boundary behavior. Unsafe is confined to allocation/row access internals and is safe because modular decode fully writes the data region before any read, and padding is explicitly zeroed.


veluca93:
I don't think we want to keep around uninit image buffers. I will close this, but reusing image buffers within the modular decoder could be a reasonable replacement for this.

PR #713, opened 2026-03-16
perf: add palette transform fast paths

hjanuschka:
This optimizes palette transforms by prefetching palette rows and adding a direct no-delta lookup fast path, while keeping delta handling behavior unchanged. It also keeps delta prediction on the cross-grid-safe path to preserve conformance correctness. Unsafe is used only for unchecked palette index reads in the proven in-bounds fast path where idx < palette_size and palette_size <= pal_row.len().

### DIFF
--- a/jxl/src/frame/modular/transforms/palette.rs
+++ b/jxl/src/frame/modular/transforms/palette.rs
@@ -24,6 +24,7 @@ const SMALL_CUBE_BITS: usize = 2;
 // SMALL_CUBE ** 3
 const LARGE_CUBE_OFFSET: usize = SMALL_CUBE * SMALL_CUBE * SMALL_CUBE;
 
+#[inline(always)]
 fn scale<const DENOM: usize>(value: usize, bit_depth: usize) -> i32 {
     // return (value * ((1 << bit_depth) - 1)) / DENOM;
     // We only call this function with SMALL_CUBE or LARGE_CUBE - 1 as DENOM,
@@ -37,9 +38,10 @@ fn scale<const DENOM: usize>(value: usize, bit_depth: usize) -> i32 {
 // The purpose of this function is solely to extend the interpretation of
 // palette indices to implicit values. If index < nb_deltas, indicating that the
 // result is a delta palette entry, it is the responsibility of the caller to
-// treat it as such.
-fn get_palette_value(
-    palette: &Image<i32>,
+/// Look up palette value. `pal_row` is the pre-fetched palette row for channel `c`.
+#[inline(always)]
+fn get_palette_value_with_row(
+    pal_row: &[i32],
     index: isize,
     c: usize,
     palette_size: usize,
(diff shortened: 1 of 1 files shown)


jxl/src/frame/modular/transforms/palette.rs
```diff
@@ -319,35 +333,74 @@ pub fn do_palette_step_one_group(
...
+                let out_row = buf_out[out_idx].data.row_mut(y);
+                #[allow(unsafe_code)]
+                for (x, &index) in index_img.iter().enumerate() {
+                    // Fast path: direct palette lookup for valid indices (common case).
+                    // Skip the multi-branch get_palette_value_with_row for the hot path.
+                    let idx = index as usize;
+                    if idx < palette_size {
+                        // SAFETY: idx < palette_size <= pal_row.len() (palette is at least
+                        // palette_size wide, validated during palette transform setup).
+                        out_row[x] = unsafe { *pal_row.get_unchecked(idx) };
```
veluca93: How sure are you that we actually need this `get_unchecked`? If we `assert` that `palette_size <= pal_row.len()`, I strongly imagine the compiler will omit the bounds check (also elsewhere).

hjanuschka: dropped the `get_unchecked` (and the `#[allow(unsafe_code)]`) at both fast-path sites

jxl/src/frame/modular/transforms/palette.rs
```diff
@@ -37,9 +38,10 @@ fn scale<const DENOM: usize>(value: usize, bit_depth: usize) -> i32 {
 // The purpose of this function is solely to extend the interpretation of
 // palette indices to implicit values. If index < nb_deltas, indicating that the
 // result is a delta palette entry, it is the responsibility of the caller to
-// treat it as such.
-fn get_palette_value(
-    palette: &Image<i32>,
+/// Look up palette value. `pal_row` is the pre-fetched palette row for channel `c`.
+#[inline(always)]
+fn get_palette_value_with_row(
```
veluca93: Please rename this back to `get_palette_value`.

jxl/src/frame/modular/transforms/palette.rs
```diff
@@ -319,30 +332,66 @@ pub fn do_palette_step_one_group(
     let num_c = buf_out.len() / (grid_xsize * grid_ysize);
     let (xsize, ysize) = buf_out[0].data.size();
 
-    for c in 0..num_c {
-        for y in 0..h {
-            let index_img = buf_in.data.row(y);
+    let palette_size = num_colors + num_deltas;
+
+    if num_deltas == 0 {
```
veluca93, line 356: In `do_palette_step_general`, this is gated by the predictor also being zero - why is this not the same here?

If the fast path is basically equivalent, shouldn't we factor it out to a function?

What images trigger this fast path?

PR #714, opened 2026-03-16
perf: tighten entropy and LZ77 hot paths

hjanuschka:
This makes ANS/Huffman/LZ77 read paths more direct in hot loops by inlining and removing repeated bounds-check overhead where invariants are already validated. It also adds a no-LZ77 clustered read helper for specialization call sites. Unsafe is only used for unchecked indexing under validated context/window invariants, with debug assertions documenting each bound.

### DIFF
--- a/jxl/src/entropy_coding/ans.rs
+++ b/jxl/src/entropy_coding/ans.rs
@@ -352,7 +352,7 @@ impl AnsHistogram {
 }
 
 impl AnsHistogram {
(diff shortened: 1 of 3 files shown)


veluca93:
This PR clearly has some speed benefits, but I am not convinced by the approach to `unsafe` - I will look at this more in depth and I will try to see how it can be improved.

PR #715, opened 2026-03-16
perf: optimize weighted predictor core and row updates

hjanuschka:
This speeds up weighted prediction by removing derived-enum conversion overhead, unrolling hot error/weight paths, and reusing per-row predictor state. It also updates row initialization call sites so the optimized predictor state layout remains behavior-equivalent. Unsafe is used only for bounds-check elimination on indices that are guarded by row-size invariants and debug assertions.

### DIFF
--- a/jxl/src/frame/modular/decode/channel.rs
+++ b/jxl/src/frame/modular/decode/channel.rs
@@ -47,11 +47,15 @@ fn decode_modular_channel_small(
 
     const { assert!(IMAGE_OFFSET.1 == 2) };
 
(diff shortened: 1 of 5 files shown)


veluca93:
This PR clearly has some speed benefits, but I am not convinced by the approach to `unsafe` - I will look at this more in depth and I will try to see how it can be improved.

TechPizzaDev:
Gotta agree with veluca93; this has a lot of unsafe... and I have a feeling some of the asserts are not enough as debug-only asserts (as in, the invariants are easy to accidentally break).

PR #718, opened 2026-03-16
perf: add no-LZ77 and config420 tree decoder variants

hjanuschka:
This adds dedicated modular tree decoder variants for no-LZ77 and config_420 histogram cases, so hot decode loops avoid unnecessary state handling. It also wires a no-LZ77 clustered read helper and specialization dispatch so the fast variants are selected automatically. The decoding behavior stays equivalent; this only narrows runtime branching in known histogram modes.

### DIFF
--- a/jxl/src/entropy_coding/decode.rs
+++ b/jxl/src/entropy_coding/decode.rs
@@ -436,6 +436,24 @@ impl SymbolReader {
         self.read_signed_clustered_inline(histograms, br, cluster)
     }
 
+    /// Fast path for reads without LZ77.
+    /// Preconditions: histograms has LZ77 disabled and reader state is None.
+    #[inline(always)]
+    pub fn read_signed_clustered_no_lz77(
+        &mut self,
+        histograms: &Histograms,
+        br: &mut BitReader,
+        cluster: usize,
+    ) -> i32 {
+        debug_assert!(matches!(self.state, SymbolReaderState::None));
+        let token = match &histograms.codes {
+            Codes::Huffman(hc) => hc.read(br, cluster),
+            Codes::Ans(ans) => self.ans_reader.read(ans, br, cluster),
+        };
+        let unsigned = histograms.uint_configs[cluster].read(token, br);
+        unpack_signed(unsigned)
+    }
+
     /// Specialized fast path for when all HybridUint configs are 420.
     ///
     /// # Preconditions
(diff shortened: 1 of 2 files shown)


veluca93:
I'm not fully convinced by the usefulness of these fast paths... @jonsneyers do you have test images for a more precise benchmark?

jonsneyers:
no-lz77, hybriduintconfig420 should be a common case in e2-e7 lossless images if the images are photographic so lz77 is unlikely to be useful. Also the LF is likely encoded like this, right? So I would expect most of the test images at https://jxl-rs-perf.lucaversari.it/ to be affected.

Whether a specialized path like this will help much or not, that's something else. Near-perfect branch prediction vs no branch might not make much of a difference, though not having to maintain the lz77 window should make at least a little bit of a difference.

veluca93:
@hjanuschka can you rebase? We can then do a more precise benchmark and see what happens :-)

jonsneyers:
ping @hjanuschka — I'm curious to see benchmark results to compare to current baseline...

hjanuschka:
rebased and pushed

veluca93 (approved):
Approved to enqueue the benchmark, will evaluate after the benchmark runs :-)

veluca93:
### Performance Summary (Commit `15362de`)

| desktop | Single | 77.61 | 78.05 | -0.25% ± 0.34% |
| desktop | Multi | 77.53 | 78.04 | -0.22% ± 0.27% |
| framework-desktop | Single | 91.33 | 92.10 | +0.65% ± 0.24% |
| framework-desktop | Multi | 91.63 | 92.76 | +0.84% ± 0.50% |
| pixel7a | Single (Fast) | 28.95 | 28.95 | +0.69% ± 0.44% |
| pixel7a | Single (Mid) | 20.66 | 20.65 | +1.00% ± 0.34% |
| pixel7a | Multi | 28.78 | 28.15 | +0.21% ± 0.40% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_718.html)

veluca93:
Ok, that's definitely a nice speedup! I have a slightly different approach in mind though, I will give it a try :-)

veluca93:
> Ok, that's definitely a nice speedup! I have a slightly different approach in mind though, I will give it a try :-)

after a re-run, this doesn't look like as much of a win as before - I am of the idea of closing this and re-evaluating later.

veluca93:
Superceded by #797

PR #719, opened 2026-03-16
perf: use padded BitReader fast path in codestream sections

hjanuschka:
This adds a padded BitReader constructor and uses it for codestream sections that already have complete buffered data. It keeps incremental/incomplete paths unchanged while avoiding slow refill handling on complete section reads.

### DIFF
--- a/jxl/src/api/inner/codestream_parser/mod.rs
+++ b/jxl/src/api/inner/codestream_parser/mod.rs
@@ -28,10 +28,37 @@ use crate::{
 mod non_section;
 mod sections;
 
+/// Buffer for a single codestream section.
+///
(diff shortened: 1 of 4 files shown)


veluca93 (commented):
I think I would prefer adding a SectionBuffer type, which wraps a vector while keeping 8 bytes of additional space (and pretending the vector is shorter).

This should significantly simplify the logic.

Also, note that the code as it currently is breaks out-of-bounds detection.

veluca93:
Can you rebase this on top of latest main? Then I will benchmark it, and we can look at a more accurate benchmark :-)

veluca93:
With the new fast paths, this doesn't seem to be a win anymore - closing it for now.

PR #720, opened 2026-03-16
perf: tighten VarDCT group coefficient hot loop

hjanuschka:
This removes unnecessary VarDCT scratch/transform zeroing and tightens coefficient writeback indexing in the hot decode loop. It keeps decode behavior unchanged while reducing redundant memory work. Unsafe is used only for permutation/current-coefficient indexing with bounds guaranteed by validated coefficient order and loop limits.

### DIFF
--- a/jxl/src/frame/group.rs
+++ b/jxl/src/frame/group.rs
@@ -61,7 +61,7 @@ impl Default for VarDctBuffers {
     }
 }
 
-#[inline(always)]
+#[inline]
 fn predict_num_nonzeros(nzeros_map: &Image<u32>, bx: usize, by: usize) -> usize {
     if bx == 0 {
         if by == 0 {
@@ -375,7 +375,6 @@ impl<'a, 'b> PassInfo<'a, 'b> {
 
 #[allow(clippy::too_many_arguments)]
 #[allow(clippy::type_complexity)]
-#[allow(unsafe_code)]
 pub fn decode_vardct_group(
     group: usize,
     passes: &mut [(usize, BitReader)],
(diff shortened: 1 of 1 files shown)


jxl/src/frame/group.rs
```diff
@@ -572,8 +573,13 @@ pub fn decode_vardct_group(
                             reader.read_signed_inline(&pass_info.histograms, br, ctx) << *shift;
                         prev = if coeff != 0 { 1 } else { 0 };
                         nonzeros -= prev;
-                        let coeff_index = permutation[k] as usize;
-                        current_coeffs[coeff_index] += coeff;
+                        // SAFETY: permutation[k] is validated to be < num_coeffs during
+                        // coeff_order decoding, and current_coeffs.len() == num_coeffs.
+                        // k < num_coeffs by loop bounds.
+                        let coeff_index = unsafe { *permutation.get_unchecked(k) } as usize;
```
veluca93: If `k < num_coeffs`, this `get_unchecked` can probably be replaced by a normal access together with an assertion that `permutation.len() >= num_coeffs`.

jxl/src/frame/group.rs
```diff
@@ -572,8 +573,13 @@ pub fn decode_vardct_group(
...
+                        // SAFETY: coeff_index comes from permutation[k], which is validated
+                        // to be < num_coeffs, and current_coeffs.len() == num_coeffs.
```
veluca93: This should either be a safety invariant, or an earlier check in the function.
How sure are we that this leads to speedups?

hjanuschka: it was part of a endless run of micro optimizations, dropped unsafe, we cannot meassure any performance increase, we could also abandon this.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `66d21dc`)

| desktop | Single | 80.94 | 81.44 | +0.54% ± 0.35% |
| desktop | Multi | 80.76 | 81.23 | +0.97% ± 0.41% |
| framework-desktop | Single | 97.09 | 94.19 | +0.58% ± 0.37% |
| framework-desktop | Multi | 96.16 | 97.19 | +1.43% ± 0.30% |
| pixel7a | Single (Fast) | 30.96 | 31.54 | +10.69% ± 9.01% |
| pixel7a | Single (Mid) | 22.93 | 23.14 | +0.64% ± 1.54% |
| pixel7a | Multi | 31.58 | 32.03 | +0.99% ± 0.44% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_720.html)

PR #705, opened 2026-03-14
perf: comprehensive single-threaded decode performance optimizations (+46%)

hjanuschka:
## Summary

Bring jxl-rs single-threaded decoding performance on par with libjxl 

### Bounds check elimination
- Unchecked array access in `predict_flat`, entropy decoding, Huffman table lookup, and weighted predictor hot paths
- Split `predict_flat` into `no_wp`/`with_wp` variants to eliminate `Option` overhead in `compute_properties`

### Inlining strategy
- Inline entropy reads (`read_signed_clustered_inline`) matching libjxl's inlining for `NoWpTree`/`GeneralTree` decoders
- Comprehensive `#[inline(always)]` audit across the entire hot path: decode, predict, entropy, blending, render pipeline, `SmallVec`, `Channels`, image buffers, utility functions

### BitReader optimization
- Extend section buffers with 8 zero-padding bytes so `refill()` always takes the fast path, eliminating `refill_slow` overhead (was ~4.5% of modular decode)

### Allocation elimination
- Pre-allocated tmp buffer + stack-allocated slice arrays in blending (eliminate per-row heap allocation)
- Hoist `property_buffer` allocation outside the row loop in modular decode
- Optimize `get_distinct_indices` with raw pointers (eliminate `Option` overhead in render pipeline) `overflow-checks=false`

PR #736, opened 2026-03-20
perf: avoid extra pull_symbol call in unsigned RLE repeat path

hjanuschka:
Carved out from the ongoing perf/autoresearch stream after PR #705 split work.

Change:
- in `SymbolReader::read_unsigned_clustered_inline`, after `push_token` consume `repeat_count/last_sym` directly instead of calling `pull_symbol` again.

Why:
- removes one extra call/branch in the hot repeat path
- behavior is unchanged


veluca93:
A more precise benchmark appears to be inconclusive. Converting this to a draft and leaving it for later evaluation.

veluca93:
Closing this for now.

PR #737, opened 2026-03-20
perf: add HybridUint::read fast path for msb_in_token == 0

hjanuschka:
Carved out from the ongoing perf/autoresearch stream after PR #705 split work.

Change:
- add a specialized fast path in `HybridUint::read` for the common `msb_in_token == 0` case.

Why:
- avoids general high-bit reconstruction on a hot path
- keeps existing behavior

### DIFF
--- a/jxl/src/entropy_coding/hybrid_uint.rs
+++ b/jxl/src/entropy_coding/hybrid_uint.rs
@@ -85,11 +85,21 @@ impl HybridUint {
         if symbol < self.split_token {
             return symbol;
         }
+
+        // Fast path: no LSB bits in token. Covers common configs (0,0), (1,0), (2,0).
+        // Eliminates low-bits extraction, symbol shift, and final shift/OR.
+        if self.lsb_in_token == 0 {
+            let nbits = (self.split_exponent - self.msb_in_token
+                + ((symbol - self.split_token) >> self.msb_in_token))
+                & 31;
+            let bits = br.read_optimistic(nbits as usize) as u32;
+            let hi = (symbol & ((1 << self.msb_in_token) - 1)) | (1 << self.msb_in_token);
+            return (hi << nbits) | bits;
+        }
+
         let bits_in_token = self.lsb_in_token + self.msb_in_token;
         let nbits =
             self.split_exponent - bits_in_token + ((symbol - self.split_token) >> bits_in_token);
-        // The bitstream is invalid if nbits >= 32. We do not report errors, and just pretend we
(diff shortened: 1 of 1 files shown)


jxl/src/entropy_coding/hybrid_uint.rs
```diff
@@ -85,6 +85,19 @@ impl HybridUint {
         if symbol < self.split_token {
             return symbol;
         }
+
+        // Common path in some LZ77/RLE configs: no MSB bits embedded in token.
```
veluca93: I think this should become a fast path for no LSB in token either (thus, just a split exponent).

In fact, just no LSB in token might be even better (as IIRC the common vardct case uses 1 msb in token but 0 lsb)

Might even make sense to do (0, 0) and (1, 0) specialized paths..

Can you try the various alternatives?

hjanuschka: Reworked to `lsb_in_token == 0` fast path as suggested. Tried all variants:

**Instrumentation** -- configs hitting `read()` (excluding `read_config_420` callers):
| Config (msb,lsb) | bike | lossless_pfm | tirr_photo |
|---|---|---|---|
| (2,0) lsb=0 | 47K | 265K | 368K |
| (1,0) lsb=0 | 444 | 104K | 34K |
| (0,0) lsb=0 | 2K | 67K | 4K |
| lsb != 0 | 67K | 344K | 227K |

You're right that (1,0) is the common vardct case -- lsb=0 covers ~50% of tirr_photo and ~56% of lossless_pfm calls.

**Benchmarks** (hyperfine, 30 runs, warmup=5):
| Variant | bike | lossless_pfm | tirr_photo | progressive_ac |
|---|---|---|---|---|
| Baseline (no fast path) | 333ms | 106ms | 3.78s | 240ms |
| `lsb_in_token == 0` (general) | 329ms | 107ms | 3.75s | 239ms |
| `(0,0)` + `(1,0)` match | slower | slower | slower | slower |

All differences are within noise (~1%). The generic code already handles lsb=0 efficiently since shifts/masks by 0 are near-free -- the branch avoidance doesn't outweigh the branch cost.

Pushed the general `lsb_in_token == 0` path since it's structurally correct (covers the right cases), doesn't regress, and is the cleaner factoring. But happy to close if you'd prefer no change given the neutral perf results.

veluca93:
I'd say that since perf is neutral the additional code complexity is not worth it, so closing it for now.

PR #744, opened 2026-03-29
Bump version to 0.4.1.

veluca93:


jxl/Cargo.toml
```diff
@@ -13,23 +13,23 @@ license = "BSD-3-Clause"
 exclude = ["resources/"]
 
 [dependencies]
-jxl_transforms = { path = "../jxl_transforms", version = "0.4.0" }
+jxl_transforms = { path = "../jxl_transforms", version = "0.4.1" }
```
tirr-c: Just noticed that this doesn't have `=` in version specifier 👀

tirr-c (approved):

jxl/Cargo.toml
veluca93: Nice catch!

tirr-c (approved):

PR #749, opened 2026-04-16
Fix scan-only panic on empty follow-up process call

hjanuschka:
In scan-only mode, a `process()` call can consume all external input and then finalize a frame from buffered internal state. When that finalization flips `has_more_frames` to false, we can currently continue the same outer loop iteration and hit the `assert!(do_flush)` path even though the call itself is valid. This change returns immediately after that transition, preserving the assert for real API misuse while avoiding this control-flow edge case.

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -1990,6 +1990,33 @@ pub(crate) mod tests {
         }
     }
 
+    /// Regression test for Chromium ClusterFuzz issue 502853162.
+    ///
+    /// Scan-only decoding may consume all external input in one `process()`
+    /// call while still having buffered frame data to finalize internally.
+    /// A subsequent empty-input `process()` call must not panic.
+    #[test]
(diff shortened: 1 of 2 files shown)


veluca93 (approved):

veluca93:
### Performance Summary (Commit `70e0b3f`)

| desktop | Single | 83.76 | 83.35 | -0.32% ± 0.36% |
| desktop | Multi | 83.91 | 83.70 | -0.01% ± 0.37% |
| framework-desktop | Single | 102.73 | 102.20 | -0.27% ± 0.58% |
| framework-desktop | Multi | 103.20 | 103.45 | -0.09% ± 0.29% |
| pixel7a | Single (Fast) | 34.05 | 33.95 | -0.17% ± 0.41% |
| pixel7a | Single (Mid) | 24.75 | 24.74 | -0.04% ± 0.37% |
| pixel7a | Multi | 33.60 | 34.19 | +0.45% ± 0.43% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_749.html)

PR #751, opened 2026-04-21
Clamp PQ and bounded HLG transfer-domain output in F16 conversion

hjanuschka:
## Summary
- clamp transfer-domain RGB values in `ConvertF32ToF16Stage`
- apply per-transfer bounds when output transfer is HDR:
  - PQ: `[0.0, 1.0]`
  - HLG: `[-0.074, 1.1]`
- keep the fix inside the jxl-rs render pipeline (no extra wrapper-side full-frame post-pass)

## Why
Tiny HDR JXL assets can produce out-of-range transfer-domain values. When those values are emitted in F16 output, downstream consumers can exhibit rendering corruption on some HDR compositor paths.

Applying the clamp in the existing F32->F16 conversion loop keeps overhead minimal and avoids a second pass over the decoded frame.

HLG is not hard-clamped to `[0,1]`; it uses a bounded headroom range to preserve plausible overshoot/undershoot workflows while still rejecting wild values.

## Testing
- `cargo test -p jxl f32_to_f16_consistency_with_unit_clamp -- --nocapture`

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -544,11 +544,7 @@ pub(crate) mod tests {
         let simple_frames = decode(&file, usize::MAX, true, false, None)?.1;
         let frames = decode(&file, usize::MAX, false, false, None)?.1;
         assert_eq!(frames.len(), simple_frames.len());
-        for (fc, (f, sf)) in frames
-            .into_iter()
-            .zip(simple_frames.into_iter())
-            .enumerate()
-        {
+        for (fc, (f, sf)) in frames.into_iter().zip(simple_frames).enumerate() {
             compare_frames(path, fc, &f, &sf)?;
         }
         Ok(())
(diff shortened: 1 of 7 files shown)


sboukortt:
Thanks for tracking down the issue!

Could we maybe clip HLG to something like [-0.074, 1.1] instead, to preserve potential overshoot/undershoot from narrow-range material? PQ can (and possibly should) still be clamped to [0, 1] as you propose.

sboukortt:
Looks good code-wise, thanks. If it’s not too much hassle to test, do you know if this still fixes the Chromium bug even with the relaxed range?

sboukortt (approved):

hjanuschka:
> Looks good code-wise, thanks. If it’s not too much hassle to test, do you know if this still fixes the Chromium bug even with the relaxed range?

sure before pushing i did test it inside chromium :)

PR #752, opened 2026-04-23
add support for out-of-order jxlp (ftyp version 1)

### LINKED ISSUE #750: Out-of-order jxlp boxes
Currently 18181-2 says the `jxlp` (partial codestream) boxes shall be in order, which is what decoders are assuming. If this is not the case, then jxl-rs will produce an error message `"Out-of-order jxlp box found"`, which is good.

In the future there might be a relaxation that allows out-of-order `jxlp` boxes. This means there will likely be a need to handle the case of out-of-order boxes. The existing syntax where in-order is obligatory will remain the default; it hasn't been decided yet how it will be signalled that the relaxed syntax is allowed but it would make sense to signal this in the `ftyp` box.

See also: https://github.com/libjxl/libjxl/issues/4735

jonsneyers:
Main change: allow out-of-order `jxlp` boxes when the `ftyp` is version 1.
Fixes https://github.com/libjxl/jxl-rs/issues/750.

Also actually check the version number in the `ftyp` box, and don't allow multiple `ftyp` boxes.

Also don't fail on "until end of file"-sized boxes that are not `jxlc` or `jxlp`, spec does allow that for any box type (say, an `Exif` box at the end of the file)


jxl/src/api/inner/box_parser.rs
```diff
@@ -39,6 +53,10 @@ pub(super) struct BoxParser {
     /// Total file bytes consumed from the underlying input.
     pub(super) total_file_consumed: u64,
     skip_jxlp_checks: bool,
+    /// From `ftyp`: `0` = `jxlp` boxes must be in order; `1` = out-of-order `jxlp` allowed.
+    jxl_file_format_version: u32,
+    jxlp_ooo_buffer: Vec<(u32, Vec<u8>, bool)>,
```
veluca93: Let's make a struct for this.

veluca93: and for that matter, shouldn't this be a BTreeMap?

jxl/src/api/inner/box_parser.rs
```diff
@@ -183,9 +339,28 @@ impl BoxParser {
...
                                 };
-                                if idx != wanted_idx {
+                                if idx < wanted_idx {
                                     return Err(Error::InvalidBox);
                                 }
+                                if idx > wanted_idx {
+                                    if self.jxl_file_format_version < 1 {
+                                        return Err(Error::InvalidBox);
+                                    }
+                                    if self.jxlp_ooo_buffer.iter().any(|(i, _, _)| *i == idx) {
```
veluca93: This is potentially quadratic (i.e. if jxlp boxes are reversed)

jxl/src/api/decoder.rs
```diff
@@ -322,6 +322,57 @@ pub(crate) mod tests {
...
+        let rgb_at = |img: &Image<f32>, x: usize, y: usize| -> (f32, f32, f32) {
+            let row = img.row(y);
+            let b = x * 3;
+            (row[b], row[b + 1], row[b + 2])
+        };
+
+        // Decoder `f32` buffer is display-referred sRGB in 0..1 (channel value ≈ sRGB8/255).
+        // Expected triples match libjxl `djxl` APNG last frame at the same coordinates.
+        let s = |c: u8| c as f32 / 255.0;
+        let checks: [((usize, usize), (f32, f32, f32)); 5] = [
```
veluca93: (fwiw there is no need to name this type)

jxl/src/api/inner/process.rs
```diff
@@ -82,6 +82,22 @@ impl SmallBuffer {
         amount
     }
 
+    /// Prepend bytes so they are returned next by [`Self::take`] (and appear first in [`Deref`]).
+    pub(super) fn inject_bytes_front(&mut self, data: Vec<u8>) {
```
veluca93, line 86: I don't love the idea of making another copy here, but ooo jxlp boxes should be rare, so this is probably not worth making the implementation more complex.

jxl/src/api/inner/box_parser.rs
jonsneyers: I'm assuming there won't be more buffered jxlp boxes than the number of sections in a frame, which is typically ~65 per 4 Mpx, so the numbers are small enough that simple linear search in a vector is faster than doing something else with better big-O complexity but worse constant factors.

In any case anyone who cares about decode perf shouldn't be using ooo jxlp boxes in the first place :)

veluca93: That doesn't help against malicious bitstreams though ;)

jxl/src/api/inner/box_parser.rs
jonsneyers: Made it a struct with a BTreeMap in it.

jxl/src/api/inner/box_parser.rs
jonsneyers: It's now a BTreeMap lookup.

jxl/src/api/decoder.rs
jonsneyers: Right, simplified.

jxl/src/api/inner/box_parser.rs
veluca93: I meant a struct for the value of the btreemap ;-)

jonsneyers: Ah right. I simplified it a bit so the value is now just the payload Vec<u8>.

jxl/src/api/inner/box_parser.rs
```diff
@@ -20,6 +20,28 @@ pub enum ParseState {
...
+    /// After the last codestream box, no more container bytes: no further codestream in file.
+    Exhausted,
+}
+
+#[derive(Default)]
+struct OooJxlpState {
+    /// From `ftyp`: `0` = `jxlp` boxes must be in order; `1` = out-of-order `jxlp` allowed.
+    file_format_version: u32,
+    /// Out-of-order `jxlp` payloads keyed by index field (MSB = is_last).
+    buffered: BTreeMap<u32, Vec<u8>>,
```
veluca93: I don't think this is an improvement over having `index -> struct { data, is_last }`.

Also, since we are keeping track of the next jxlp box, why do we need a sorted map instad of a HashMap?

jonsneyers: Right. Changed back to `index -> struct { data, is_last }`, and made it a HashMap.

veluca93 (approved):

PR #754, opened 2026-04-24
Remove CmsStage

sboukortt:


jxl/src/api/inner/codestream_parser/non_section.rs
```diff
@@ -178,11 +178,8 @@ impl CodestreamParser {
             if let Some(user_profile) = &self.output_color_profile {
                 // Validate user's output color profile choice (libjxl compatibility)
                 // For non-XYB without CMS: only same encoding as embedded is allowed
```
veluca93: Comment is stale :-)

jxl/src/api/inner/codestream_parser/non_section.rs
```diff
@@ -178,11 +178,8 @@ impl CodestreamParser {
...
-                if !xyb_encoded
-                    && decode_options.cms.is_none()
-                    && *user_profile != embedded_color_profile
-                {
-                    return Err(Error::NonXybOutputNoCMS);
+                if !xyb_encoded && *user_profile != embedded_color_profile {
+                    return Err(Error::NonXybConversionRequested);
```
veluca93: (how do we even trigger this?)

jxl/src/api/inner/mod.rs
```diff
@@ -67,13 +64,9 @@ impl JxlDecoderInner {
 
     /// Specifies the preferred color profile to be used for outputting data.
     /// Same semantics as JxlDecoderSetOutputColorProfile.
```
veluca93: I don't think this is true anymore.

jxl/src/api/inner/mod.rs
```diff
@@ -67,13 +64,9 @@ impl JxlDecoderInner {
...
-    pub fn set_output_color_profile(&mut self, profile: JxlColorProfile) -> Result<()> {
-        if let (JxlColorProfile::Icc(_), None) = (&profile, &self.options.cms) {
-            return Err(Error::ICCOutputNoCMS);
-        }
-        self.codestream_parser.output_color_profile = Some(profile);
+    pub fn set_output_color_profile(&mut self, profile: JxlColorEncoding) {
```
veluca93: Maybe we should rename this to `set_preferred_color_encoding` or somesuch?
(in particular we no longer promise to actually honour the request)

jxl/src/api/decoder.rs
```diff
@@ -741,10 +737,10 @@ pub(crate) mod tests {
 
         // Once output color profile is set by user, it will remain as is regardless of what pixel
```
veluca93: I am not sure this comment is still true either.

jxl/src/api/inner/mod.rs
veluca93: Actually, thinking about it - when would we *actually* honour the request?
If there is any blending to be done, we might need double conversion to actually output in this profile. However, I don't think we can detect reliably the _absence_ of blending, or can we? (you can always have a non-animated image with layers)
So maybe we should get rid of this entirely?

veluca93 (approved):

PR #757, opened 2026-05-01
Fix an unsound API allowing creating slices with incorrect alignment

Shnatsel:
This fixes a soundness issue in `Image<T>::from_raw`.

`Image<T>::from_raw` checked that the underlying `RawImageBuffer` base pointer, row size, and stride were aligned for `T`, but it did not check `OwnedRawImage`’s byte offset. As a result, safe code could construct an `OwnedRawImage` whose allocation was correctly aligned but whose logical image rows started at a misaligned byte offset. Calling `Image::<u32>::row()` on that image then created a misaligned `&[u32]`, which Miri reports as UB.

The fix adds an offset-alignment check in `Image<T>::from_raw`, so misaligned raw images are rejected before any typed slice can be formed. The regression test constructs such an image with a one-byte offset and verifies that `Image::<u32>::from_raw` panics instead of allowing later UB. The safety comments for typed row access were also updated to include the byte-offset alignment invariant.

## Impact analysis

This cannot be used for out-of-bounds accesses because:

- `OwnedRawImage::row()` first creates a byte slice with normal checked indexing:
  `data.row(...)[offset.0..offset.0 + byte_size.0]`.
  If `offset.0 + byte_size.0` exceeds the allocated row width, this panics before the unsafe typed cast.

- `Image<T>::row()` then does:
  `from_raw_parts(row.as_ptr().cast::<T>(), row.len() / size_of::<T>())`.
  Since the length is rounded down from the byte slice length, the typed slice covers at most the byte slice range, never past it.

I do not believe this issue to be exploitable in any way.

This issue was discovered by GPT-5.5

### DIFF
--- a/jxl/src/image/test.rs
+++ b/jxl/src/image/test.rs
@@ -67,6 +67,20 @@ fn rect_basic() -> Result<()> {
     Ok(())
 }
 
+#[cfg(miri)]
+#[test]
+fn image_from_raw_with_misaligned_offset_triggers_ub() -> Result<()> {
+    use super::OwnedRawImage;
+    // Issue: Image<T>::from_raw checks the base RawImageBuffer alignment, but it does not check
+    // that OwnedRawImage's byte offset is aligned for T. This safe construction leaves the
+    // allocation itself 4-byte aligned while making row(0) start one byte later; Image::<u32>::row
+    // then creates a misaligned &[u32], which Miri reports as UB.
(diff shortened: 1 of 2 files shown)


Shnatsel:
I have signed the CLA.

veluca93 (commented):
Looks like this slipped through when adding the byte_offset feature...

Nice catch! I agree with your assessment that this should not be exploitable.

veluca93 (approved):

PR #758, opened 2026-05-01
Remove support for uninitialized image buffers.

veluca93:


jxl/src/image/internal.rs
```diff
@@ -20,10 +19,9 @@ use super::Rect;
 #[derive(Debug, Clone, Copy)]
 pub(super) struct RawImageBuffer {
     // Safety invariant:
-    //  - uninit data is never written to `buf`.
     //  - `buf` is valid for reads for all bytes in the ranges
     //    `buf[i*bytes_between_rows..i*bytes_between_rows+bytes_per_row]` for all values of `i`
-    //    from `0` to `num_rows-1`.
+    //    from `0` to `num_rows-1` (as well as initialized).
```
anforowicz: I am not sure what this means.  

Maybe it wants to say: "okay to read padding if it was initialized"?
Maybe it wants to say: "all such readable bytes are guaranteed to have been initialized earlier"?

jxl/src/image/raw.rs
```diff
@@ -37,10 +37,13 @@ impl OwnedRawImage {
         Ok(Self {
             // Safety note: the returned memory is initialized and part of a single allocation of
             // the correct length.
-            data: RawImageBuffer::try_allocate(
-                (byte_size.0 + padding.0, byte_size.1 + padding.1),
-                false,
-            )?,
+            // SAFETY: `copy_from` is `None`.
```
anforowicz, line 40: I guess it would be nice if we could avoid `unsafe` here.  But I guess `try_allocate` takes an `Option` because sometimes we don't know at runtime if we'll use `None` vs `Some`?

anforowicz (approved):

jxl/src/image/internal.rs
veluca93: Clarified!

jxl/src/image/raw.rs
veluca93: No, this is meant to avoid duplicating the unsafe code in `try_allocate` itself. I don't think having a dedicated function would reduce the total unsafe code.

mo271 (approved):

PR #762, opened 2026-05-04
flush_pixels: gate Ok(true) on real modular decode

tnikkel:
## Summary

Without this, `flush_pixels` returns `Ok(true)` for modular files before any decoded pixels have actually reached the buffer: `decode_and_render_hf_groups` flushes zero-filled empty channels through the render pipeline, populates `requested_rects`, and `rendered = !regions.is_empty()` fires. Callers that treat `Ok(true)` as "real pixels are available" then read an empty/black surface.

This adds a `real_decode_since_last_flush` flag on `FullModularImage`:

- Set by `read_section0` when newly decoded section-0 channel count grows.
- Set by `read_stream` after a successful subbitstream decode into a section that has buffers (empty/meta sections do not produce content).
- Cleared by `flush_pixels` after consuming `pixels_dirty`.

`decode_and_render_hf_groups` checks the flag for modular frames only; VarDCT frames write pixels directly via `decode_hf_group` and are unaffected. The LF-preview path is also unaffected (it returns earlier).

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo test --workspace --release` — all tests pass

### DIFF
--- a/jxl/src/api/inner/process.rs
+++ b/jxl/src/api/inner/process.rs
@@ -147,6 +147,12 @@ impl JxlDecoderInner {
             Ok(()) | Err(crate::error::Error::OutOfBounds(_)) => {
                 let updated = self.codestream_parser.pixels_dirty;
                 self.codestream_parser.pixels_dirty = false;
+                // Clear the modular-side "real decode happened" flag so that
+                // subsequent flushes only report Ok(true) when new real decode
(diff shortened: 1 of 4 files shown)


veluca93:
I believe #764 fixes the same issues in a better way, so closing this unless we realize it doesn't :-)

PR #766, opened 2026-05-06
Fix: capture LZ77 distance cluster before context_map padding (resolves #765)

### LINKED ISSUE #765: Decoder rejects spec-valid VarDCT bitstream with Invalid AC: N nonzeros after decoding block (e9 + low distance + screen content)
## Summary

`jxl-rs` v0.4.3 (commit `e7405e0`) rejects spec-valid `.jxl` bitstreams with
`Invalid AC: N nonzeros after decoding block` at
`jxl_cli/src/dec/mod.rs:305:19` (the underlying error is raised at
`jxl/src/frame/group.rs:579`, `Error::EndOfBlockResidualNonZeros(nonzeros)`).

Both `libjxl djxl 0.12.0` (canonical reference) and `jxl-oxide 0.12.5`
decode the same files cleanly, so the bug is in `jxl-rs`'s AC token
reader path.

## Repro

Smallest reproducer: `akfcrc022_e9_d3.0.jxl` (22,416 bytes, 512×512,
attached). Encoded by an experimental Rust JXL encoder at
`--effort 9 --distance 3.0` on screen content.

```bash
cargo run --release -p jxl_cli -- akfcrc022_e9_d3.0.jxl out.png
# Error: Invalid AC: 8 nonzeros after decoding block
# Location: jxl_cli/src/dec/mod.rs:305:19

# Reference comparison:
djxl akfcrc022_e9_d3.0.jxl out_libjxl.png   # ✓ accepts
jxl-oxide decode akfcrc022_e9_d3.0.jxl --output out_oxide.png   # ✓ accepts
```

## Pattern

Found via a 7,329-file sweep over jxl-encoder output. Of 25 genuine
decoder bugs (cases libjxl djxl 0.12.0 accepts but jxl-rs rejects), 22
share the same trigger surface:

- `effort=9` encoder settings (enables `enhanced_clustering_vardct`,
  `optimize_uint_configs_vardct`, optimal LZ77, fine-grained step 1)
- Distance ≤ 2.0 (sometimes ≤ 3.0)
- Screen content (text/UI/graphs)

Error counts seen at the failure point: `EndOfBlockResidualNonZeros(N)`
with N ∈ {1, 2, 3, 4, 8, 9, 15}. A related family of 3 files reports
`Invalid AC: nonzeros N is too large for 1 8x8 blocks` (N ∈ {112, 144}
for `num_blocks=1`), and 4 files fail with `ANS stream checksum
mismatch` from the same trigger surface — these likely share root cause
since they all decode cleanly in libjxl.

## Investigation so far

The failure on `akfcrc022_e9_d3.0.jxl` localizes to block (29, 2) in
the B (blue) channel of frame 0:

- DCT8 (1×1, num_coeffs=64), `hf_mul=7`, declared `nonzeros=14`.
- The encoder emitted an LZ77 length=50 / distance=1 ("repeat last
  symbol 50 times") at an earlier block in the Y channel, which spans
  c=1 k=4..k=37 (34 emissions), c=0 nz0+k=1..14 (15 emissions), and
  c=2 nz0+k=1 (last 1 emission of the 50-copy).
- After the LZ77 copy ends, `c=2 k=2` reads from the bitstream and
  decodes mostly zeros where the encoder wrote nonzeros, leaving 8
  nonzeros undetected at the end of the block.

I checked the code paths against libjxl C and they match for:

- `zero_density_context` formula and tables
- `nonzero_context`, `block_context_map.block_context`
- `block_shape_id` table vs `kStrategyOrder`
- `predict_num_nonzeros` vs `PredictFromTopAndLeft`
- `HybridUint::read` / `HybridUint::decode`
- LZ77 trigger / length / distance / copy / window mechanics
- Special-distances table and `dist_multiplier=0` arithmetic
- `unpack_signed`
- `Lz77Params` bitstream layout
- ANS reader (jxl-rs uses jxl-oxide's ANS implementation)
- The PR #671 context-map padding fix (`ZERO_DENSITY_CONTEXT_LIMIT`) is integrated

The divergence appears to be at the symbol-decode level after the LZ77
copy ends, so the suspect surface is one of:

- ANS state evolution during the LZ77 trigger reads (1 ANS read for the
  trigger token + 1 for the distance token, with `ReadHybridUintConfig`
  bits in between)
- Cluster / histogram lookup for the post-copy context
- Some subtle bit-position offset I missed

I haven't been able to localize it by inspection alone.

## Companion files (other genuine decoder bugs from the sweep)

If useful, the other 21 bug files share the same trigger pattern; happy
to attach a representative subset. They live at
`gb82-sc_{terminal,gui,graph,codec_wi,gmessage,imac_dar,windows,windows9}_*`.

## Versions

- jxl-rs: `main` HEAD `e7405e0` (v0.4.3)
- libjxl: `v0.12.0` (commit `d2c7032`)
- jxl-oxide: `0.12.5`

lilith:
Resolves #765.

## Mechanism

When LZ77 is enabled, `Histograms::decode` follows the spec — the decoded `context_map`'s last entry is the LZ77 distance cluster, matching libjxl's `code->lz77.nonserialized_distance_context = context_map->back()` (`dec_ans.cc:362`).

But `Frame::decode` (`jxl/src/frame/decode.rs:786-789` and `:1057-1059`) then calls `histograms.resize(num_contexts + 16)` to pad `context_map` with 16 zero entries so that `ZeroDensityContext` lookups can index up to `ZERO_DENSITY_CONTEXT_LIMIT` without bounds checks (the PR #671 fix). That padding is correct for AC context lookups — but it also moves the original LZ77 distance cluster away from `context_map.last()`.

The next time the AC reader fires an LZ77 trigger, `lz_dist_cluster = *histograms.context_map.last().unwrap()` reads a zero pad byte, the distance ANS symbol is decoded against histogram `0` (wrong), and the ANS state gets the wrong `dist`/`offset` update. State is corrupted for all subsequent reads, manifesting as `Invalid AC: N nonzeros after decoding block` (or `InvalidNumNonZeros`, or `AnsChecksumMismatch`, depending on how far the corruption propagates before the validation fails).

## Why libjxl C and jxl-oxide don't have this bug

- **libjxl** captures the distance cluster as `lz77_ctx_` at `ANSSymbolReader::Create` time (`dec_ans.cc:411`) and never re-reads `context_map.back()`. It also doesn't do resize-with-padding — it bounds-checks at AC context sites instead.
- **jxl-oxide** pre-clusters and stores the distance histogram explicitly in its `Coder` structure (`crates/jxl-coding/src/lib.rs`), so even if the visible context map were padded, the distance read uses the captured handle.

## Why libjxl reference test corpora don't catch this

libjxl's encoder *never* uses `Lz77Method::Optimal` or `Lz77Method::Greedy` for VarDCT — only `kRLE`, per `lib/jxl/enc_ans_params.h` (`HistogramParams` constructor) + `lib/jxl/enc_frame.cc:1259-1263`. `Optimal`/`kLZ77` are reserved for **Modular** mode at slow tiers (`HistogramParams::ForModular` in `enc_ans.cc`). So the reference test corpus has no Optimal/Greedy LZ77 VarDCT bitstreams to exercise this code path.

Third-party encoders that emit Optimal/Greedy LZ77 in VarDCT streams (e.g. [imazen/jxl-encoder](https://github.com/imazen/jxl-encoder) at effort 9+) trip this on every output file with non-trivial backward references.

## Fix

Capture `lz_dist_cluster` in `Histograms` at decode time, before any external `resize` can shift it. Replaces both call sites of `*histograms.context_map.last().unwrap()` with `histograms.lz_dist_cluster as usize`.

Mirrors libjxl's `ANSCode::lz77.nonserialized_distance_context` pattern.

## Verification

- A 22-file regression corpus (mix of `EndOfBlockResidualNonZeros`, `InvalidNumNonZeros`, `AnsChecksumMismatch` errors — all from the same underlying state corruption) was used to validate. Before fix: 22/22 fail. After fix: 22/22 decode, all matching `djxl 0.12.0` output within standard pixel-parity tolerance.
- Verified directly on `jxl_cli` built from this branch: all 6 previously-failing repro `.jxl` files decode without error.
- This same fix was first applied and validated in our zenjxl-decoder fork (which is a thin downstream port of jxl-rs); full lib test suite (1231 tests) passed there with zero regressions.

## Files

`jxl/src/entropy_coding/decode.rs` — adds `lz_dist_cluster: u8` field to `Histograms`, captures it at decode time, uses it in the two SymbolReader call sites instead of re-reading `context_map.last()`. Test helpers `Histograms::reverse_octet` and `Histograms::rle` updated to set the new field.

Net diff: +33 / −2 lines.

### DIFF
--- a/jxl/src/entropy_coding/decode.rs
+++ b/jxl/src/entropy_coding/decode.rs
@@ -77,7 +77,7 @@ pub struct Histograms {
     /// `context_map.last()` after a resize would silently return a pad byte
     /// (0) and route LZ77 distance ANS reads through the wrong histogram.
     ///
-    /// Set to 0 when LZ77 is disabled.
+    /// Explicitly stored: context_map can get padded so we cannot use .last()
     lz_dist_cluster: u8,
     // TODO(veluca): figure out why this is unused.
     #[allow(dead_code)]
--- a/jxl/src/entropy_coding/decode.rs
+++ b/jxl/src/entropy_coding/decode.rs
@@ -63,6 +63,22 @@ pub struct Histograms {
     lz77_params: Lz77Params,
     lz77_length_uint: Option<HybridUint>,
     context_map: Vec<u8>,
+    /// Cluster index used for LZ77 distance reads.
+    ///
+    /// Captured at `Histograms::decode` time as the original last entry of
+    /// `context_map`, mirroring libjxl's
+    /// `code->lz77.nonserialized_distance_context = context_map->back()`
+    /// (`dec_ans.cc:362`).
+    ///
+    /// Stored separately because callers may pad `context_map` with zeros
+    /// after decode (see `Frame::decode` calling `Histograms::resize` to
+    /// avoid bounds checks at AC `ZeroDensityContext` lookup sites). That
+    /// padding moves the original last entry, so reading
+    /// `context_map.last()` after a resize would silently return a pad byte
+    /// (0) and route LZ77 distance ANS reads through the wrong histogram.
+    ///
+    /// Set to 0 when LZ77 is disabled.
+    lz_dist_cluster: u8,
     // TODO(veluca): figure out why this is unused.
     #[allow(dead_code)]
     log_alpha_size: usize,
@@ -243,7 +259,7 @@ impl SymbolReader {
             let min_length = min_length.unwrap();
             let dist_multiplier = image_width.unwrap_or(0) as u32;
 
-            let lz_dist_cluster = *histograms.context_map.last().unwrap() as usize;
+            let lz_dist_cluster = histograms.lz_dist_cluster as usize;
             let lz_conf = &histograms.uint_configs[lz_dist_cluster];
             let is_rle = histograms.codes.single_symbol(lz_dist_cluster) == Some(1)
                 && lz_conf.is_split_exponent_zero();
@@ -370,7 +386,7 @@ impl SymbolReader {
                     return 0;
                 };
 
-                let lz_dist_cluster = *histograms.context_map.last().unwrap() as usize;
+                let lz_dist_cluster = histograms.lz_dist_cluster as usize;
                 let distance_sym = match &histograms.codes {
                     Codes::Huffman(hc) => hc.read(br, lz_dist_cluster),
                     Codes::Ans(ans) => self.ans_reader.read(ans, br, lz_dist_cluster),
@@ -592,6 +608,17 @@ impl Histograms {
         };
         assert_eq!(context_map.len(), num_contexts);
 
+        // Capture the LZ77 distance cluster BEFORE any later resize() can pad
+        // context_map with zeros (see Histograms::resize and Frame::decode).
+        // Mirrors libjxl's
+        // ANSCode::lz77.nonserialized_distance_context = context_map->back();
+        // (dec_ans.cc:362).
+        let lz_dist_cluster = if lz77_params.enabled {
+            *context_map.last().unwrap()
+        } else {
+            0
+        };
+
         let use_prefix_code = br.read(1)? != 0;
         let log_alpha_size = if use_prefix_code {
             HUFFMAN_MAX_BITS
@@ -616,6 +643,7 @@ impl Histograms {
             lz77_params,
             lz77_length_uint,
             context_map,
+            lz_dist_cluster,
             log_alpha_size,
             uint_configs,
             codes,
@@ -659,6 +687,7 @@ impl Histograms {
             uint_configs,
             log_alpha_size: 15,
             context_map: vec![0u8; num_contexts],
+            lz_dist_cluster: 0,
             codes,
         }
     }
@@ -669,6 +698,7 @@ impl Histograms {
         let uint_configs = vec![HybridUint::new(8, 0, 0), HybridUint::new(0, 0, 0)];
         let mut context_map = vec![0u8; num_contexts + 1];
         *context_map.last_mut().unwrap() = 1;
+        let lz_dist_cluster = *context_map.last().unwrap();
         Self {
             lz77_params: Lz77Params {
                 enabled: true,
@@ -679,6 +709,7 @@ impl Histograms {
             uint_configs,
             log_alpha_size: 15,
             context_map,
+            lz_dist_cluster,
             codes,
         }
     }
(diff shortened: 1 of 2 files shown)


lilith:
Upstream test suite confirmed clean on this branch:

\`\`\`
$ cargo test -p jxl --lib
test result: ok. 852 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 1609.02s
\`\`\`

(Built and run from the branch tip on x86_64 Linux. The single ignored test was already ignored on \`origin/main\`.)

Plus direct functional verification: all 6 previously-failing repro \`.jxl\` files at \`/mnt/v/output/zenjxl-decoder-bug/\` now decode without error through the patched \`jxl_cli\` release binary.

jonsneyers:
Good catch!
This is why it's good to have not just independent decoder implementations but also encoders...

Thanks!

jxl/src/entropy_coding/decode.rs
```diff
@@ -63,6 +63,22 @@ pub struct Histograms {
...
+    /// (`dec_ans.cc:362`).
+    ///
+    /// Stored separately because callers may pad `context_map` with zeros
+    /// after decode (see `Frame::decode` calling `Histograms::resize` to
+    /// avoid bounds checks at AC `ZeroDensityContext` lookup sites). That
+    /// padding moves the original last entry, so reading
+    /// `context_map.last()` after a resize would silently return a pad byte
+    /// (0) and route LZ77 distance ANS reads through the wrong histogram.
+    ///
+    /// Set to 0 when LZ77 is disabled.
```
jonsneyers: This comment is nice but a bit too verbose. We can always do a git blame to find more details on why this is there; in the code I'd keep the comments shorter, just something like
```suggestion
    /// Explicitly stored: context_map can get padded so we cannot use .last()
```

jxl/src/entropy_coding/decode.rs
```diff
@@ -592,6 +608,17 @@ impl Histograms {
         };
         assert_eq!(context_map.len(), num_contexts);
 
+        // Capture the LZ77 distance cluster BEFORE any later resize() can pad
+        // context_map with zeros (see Histograms::resize and Frame::decode).
+        // Mirrors libjxl's
+        // ANSCode::lz77.nonserialized_distance_context = context_map->back();
+        // (dec_ans.cc:362).
```
jonsneyers: Same here, the first two lines of the comment suffice, no need to add pointers to libjxl code.

veluca93 (approved):

PR #773, opened 2026-05-21
Fix clipped blending with missing references

### LINKED ISSUE #772: bug in frame blending
Taking the 'Layers example' image from https://sneyers.info/jxltran/ and changing the blend mode for the bottom layer (the red one) to alpha-weighted add produces a file that causes a panic:
[blendbug.zip](https://github.com/user-attachments/files/28101740/blendbug.zip)

```
thread 'main' (18491565) panicked at jxl/src/features/blending.rs:104:32:
copy_from_slice: source slice length (750) does not match destination slice length (253)
```

hjanuschka:
Fixes #772.

When frame blending uses a missing reference slot, we fall back to a zero slice. That fallback needs to match the clipped blend width, not the full image width, otherwise alpha-weighted-add can panic on edge chunks.

Added the crash sample to `resources/test/conformance_test_images`. The simple/low-memory compare gets a tiny float tolerance, since this sample differs by ~1e-7 between pipelines.

Tested:
- cargo fmt --check
- cargo test -p jxl decode_test_file_issue772_blendbug --lib
- cargo test -p jxl decode_test_file_chunks_issue772_blendbug --lib
- cargo test -p jxl compare_pipelines_issue772_blendbug --lib
- cargo test -p jxl render::stages::blending::test::blending_consistency --lib


jxl_macros/src/lib.rs
```diff
@@ -718,10 +718,19 @@ pub fn for_each_test_file(input: TokenStream) -> TokenStream {
         .join("resources")
         .join("test");
     let conformance_test_dir = root_test_dir.join("conformance_test_images");
+    let decode_only_test_dir = root_test_dir.join("decode_only");
```
veluca93: Why?

jxl/src/api/decoder.rs
```diff
@@ -584,10 +584,12 @@ pub(crate) mod tests {
             }
             for y in 0..sz.1 {
                 for x in 0..sz.0 {
-                    assert_eq!(
+                    let delta = (b.row(y)[x] - sb.row(y)[x]).abs();
```
veluca93: We should leave a TODO to switch back to equality here - the fact that equality checks don't pass is IMO indicative of some other bug.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `15903cb`)

| framework-desktop | Single | 97.31 | 97.68 | +0.44% ± 0.42% |
| framework-desktop | Multi | 98.06 | 98.41 | +0.29% ± 0.23% |
| pixel7a | Single (Fast) | 31.76 | 32.04 | +0.27% ± 0.53% |
| pixel7a | Single (Mid) | 22.94 | 23.14 | +0.28% ± 0.32% |
| pixel7a | Multi | 31.71 | 31.46 | -0.10% ± 0.37% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_773.html)

veluca93 (approved):

PR #779, opened 2026-05-25
Make progressive previews of modular images smoother.

veluca93:
This makes progressive_dc images show progressive previews that are not blocky.

### DIFF
--- a/jxl/src/frame/modular/transforms/apply.rs
+++ b/jxl/src/frame/modular/transforms/apply.rs
@@ -3,15 +3,17 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-use std::fmt::Debug;
+use std::{collections::HashMap, fmt::Debug};
 
 use num_traits::FromPrimitive;
 
 use crate::{
     error::{Error, Result},
     frame::modular::{
-        ChannelInfo, ModularBufferInfo, ModularChannel, ModularGridKind, Predictor,
+        BUFFER_STATUS_ZERO_FILLED, ChannelInfo, ModularBufferInfo, ModularChannel, ModularGridKind,
+        Predictor,
         borrowed_buffers::with_buffers,
+        transforms::squeeze::{smooth_2d_unsqueeze, smooth_h_unsqueeze, smooth_v_unsqueeze},
     },
     headers::{
         self,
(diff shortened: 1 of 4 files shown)


veluca93:
### Performance Summary (Commit `7da090d`)

| framework-desktop | Single | 97.56 | 97.54 | -0.24% ± 0.25% |
| framework-desktop | Multi | 98.27 | 98.57 | -0.10% ± 0.25% |
| pixel7a | Single (Fast) | 31.49 | 30.91 | -0.16% ± 0.39% |
| pixel7a | Single (Mid) | 23.16 | 22.57 | -1.19% ± 0.28% |
| pixel7a | Multi | 31.73 | 30.70 | -0.81% ± 0.34% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_779.html)

jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -675,3 +675,346 @@ pub fn do_vsqueeze_step(
...
+        [0, 0, 0],
+        [474, 3145, 474],
+        [6787, 27370, 6787],
+        [3145, 14093, 3145],
+        [0, 116, 0],
+    ],
+];
+
+#[allow(clippy::needless_range_loop)]
+pub fn smooth_2d_unsqueeze(input: &Image<i32>, output: &mut Image<i32>) {
```
jonsneyers, line 749: Is this better than reusing some existing upsampling method like JPEG chroma upsampling or the fancy-nonseparable 2x upsampling? It might save some lines of code / binary size if we don't have separate codepaths with different upsampling methods for all kinds of upsampling that are done. Though it may be justified if the existing methods look weird/worse than this for this purpose.

jonsneyers (approved):

jxl/src/frame/modular/transforms/apply.rs
```diff
@@ -540,15 +608,39 @@ fn meta_apply_single_transform(
...
+                        if let Some(t) = transform_step_for_buf.get(&buf_out)
+                            && !t.1
+                        {
+                            let TransformStep::VSqueeze { buf_in_avg, .. } =
+                                &mut transform_steps[t.0]
+                            else {
+                                unreachable!()
+                            };
+                            *buf_in_avg = Some([buf_0, buf_1]);
+                        }
```
tirr-c: It seems that it's possible to move this out of `if horizontal` block and merge with `else` block by adding `HSqueeze` pattern to let-else, wdyt?

tirr-c (approved):

jxl/src/frame/modular/transforms/squeeze.rs
veluca93: discussed offline, but there is no good way to do that here :-)

jxl/src/frame/modular/transforms/apply.rs
veluca93: I like that :-) (I am also wondering if it makes sense to keep the separate enum cases for H and V squeeze..)

PR #786, opened 2026-06-01
Add public TOC inspection API for progressive streaming

### LINKED ISSUE #387: Way to tell the byte range needed for full lower-resolution image
To integrate the progressive decoding better with the browser, one would need to tell how much bytes are needed to show the full lower-res image. Having an API would be nice for that.

(Not sure how exactly such integration would happen yet. Ideally `` should be able to automatically pause and resume the network stream from information in the header. One can also use `<picture>` or `srcset` to achieve similar, but that would require extra server side logic and also redownload of the initial bits)

tokyovigilante:
Expose the frame Table of Contents through the public decoder API so streaming consumers can locate frame-section byte boundaries without fully decoding the bitstream: `TocEntry` / `TocGroupKind` plus `JxlDecoder::{toc_num_entries, toc_entry, frame_data_size, frame_data_offset}`, available once the decoder reaches `WithFrameInfo`.

Entries are reported in bitstream order, each kind resolved through the TOC permutation when present. The section-data offset accumulates the TOC byte size across `process()` calls, since the TOC may be parsed incrementally.

Tests cover TOC invariants (single/multi-group, permuted), bare-vs- container consistency, and a chunked-vs-single-shot regression.

AI disclosure: written with Claude Code (Opus 4.8) but all code reviewed and unit tested, as well as against jxl-oxide and libjxl outputs. Also tested against an in-house JXL streaming server and client previously using jxl-oxide. 

Closes #387

### DIFF
--- a/jxl/src/api/data_types.rs
+++ b/jxl/src/api/data_types.rs
@@ -288,3 +288,21 @@ pub struct JxlFrameHeader {
     /// Frame size (width, height)
     pub size: (usize, usize),
 }
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum TocGroupKind {
+    All,
+    LfGlobal,
+    LfGroup(u32),
+    HfGlobal,
+    GroupPass { pass_idx: u32, group_idx: u32 },
+}
+
+#[derive(Debug, Copy, Clone)]
+pub struct TocEntry {
+    pub kind: TocGroupKind,
+    /// Offset in bytes from the start of frame data (after the frame header).
+    pub offset: u64,
+    /// Size of the entry in bytes.
+    pub size: u32,
+}
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -5,7 +5,7 @@
 
 use super::{
     JxlBasicInfo, JxlBitstreamInput, JxlColorProfile, JxlDecoderInner, JxlDecoderOptions,
-    JxlOutputBuffer, JxlPixelFormat, ProcessingResult,
+    JxlOutputBuffer, JxlPixelFormat, ProcessingResult, TocEntry,
 };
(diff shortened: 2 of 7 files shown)


tokyovigilante:
Rebased onto `v0.6.0` with a few updates.

#828 rewrote box/codestream parsing and deleted `codestream_parser/non_section.rs`, where the old byte accounting lived, so the API is re-implemented against the new structure. Notes on what changed:

- **The second patch dropped.** It accumulated TOC byte size across incremental parses because the old code recomputed it from the final `BitReader`. `SmallBuffer` now carries a monotone `consumed` and `with_br` accounts for partial reads including the `OutOfBounds` path, so there is nothing left to get wrong.

- **New: `BoxParser::file_position_at`.** `total_bytes_consumed` resolves a codestream position via `codestream_pos_to_box.last_key_value()`. This is correct at EOF but wrong for a position sampled mid-parse, since the box parser can have read ahead into a later `jxlp`. The new sibling looks up the box that actually contains the position. Same result for bare/`jxlc` files.

- **`data_offset` is on `FrameInfo` as `Option<u64>`**, not a `Frame` field. This  keeps byte bookkeeping in the parser layer, and the `Option` encodes "this offset belongs to the current frame" (skipped frames build no `Frame`).

- **`num_completed_passes` is re-added.** #828 dropped it as a non-useful API method;  SectionState::completed_passes` still exists, nothing exposes it.  A progressive viewer needs it to decide when a re-flush is worth doing. Happy to drop it again if you'd rather that stayed internal.

- Scalar getters return `0` rather than panicking when there is no current frame: `scan_frames_only` legitimately reaches `WithFrameInfo` with no frame.

Tests: `jxl/src/tests/toc.rs`, 10 cases.

tokyovigilante:
Updated for 0.7.3.

PR #787, opened 2026-06-01
Revisit modular special cases, and add a single-symbol path.

veluca93:
This gives a significant speedup on jxl-art images (i.e. 1.4x on sunset_logo.jxl)

### DIFF
--- a/jxl/src/entropy_coding/decode.rs
+++ b/jxl/src/entropy_coding/decode.rs
@@ -651,6 +651,10 @@ impl Histograms {
     pub fn can_use_config_420_fast_path(&self) -> bool {
         !self.lz77_params.enabled && self.uint_configs.iter().all(|cfg| cfg.is_config_420())
     }
+
+    pub fn single_symbol(&self, ctx: usize) -> Option<u32> {
+        self.codes.single_symbol(ctx)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


veluca93:
### Performance Summary (Commit `cd18d3b`)

| desktop | Single | 81.50 | 91.13 | +6.88% ± 0.34% |
| desktop | Multi | 81.08 | 90.98 | +7.34% ± 0.52% |
| framework-desktop | Single | 97.65 | 109.50 | +7.72% ± 0.47% |
| framework-desktop | Multi | 97.68 | 109.57 | +8.16% ± 0.30% |
| pixel7a | Single (Fast) | 31.96 | 33.16 | +4.84% ± 0.41% |
| pixel7a | Single (Mid) | 23.31 | 23.68 | +3.35% ± 0.28% |
| pixel7a | Multi | 31.62 | 32.97 | +5.40% ± 0.68% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_787.html)

jonsneyers (approved):

jonsneyers:
Nice, seems to help not only for jxl art but also a bit for VarDCT images, which is a bit surprising. I don't expect LF decoding would benefit from this (except for grayscale with all-zero XB channels), so would it be the aux fields decoding that makes a difference?

veluca93:
My guess is that this helps due to the rewrite of the generic logic, not due to the new fast path :-)

PR #789, opened 2026-06-03
perf: always shift in modular::error_weight

TechPizzaDev:
Unlocks auto-vec at opt-level=3; mostly effective for AVX512 like Zen4, and shouldn't regress older targets

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -30,6 +30,7 @@ Martin Bruse <zondolfin@gmail.com>
 Moritz Firsching
 Mrmaxmeier <3913977+Mrmaxmeier@users.noreply.github.com>
 Sami Boukortt
+TechPizza <23627133+TechPizzaDev@users.noreply.github.com>
 Tomáš Král <tomas@kral.hk>
 Wonwoo Choi <chwo9843@gmail.com>
 Zoltan Szabadka
--- a/jxl/src/frame/modular/predict.rs
+++ b/jxl/src/frame/modular/predict.rs
@@ -332,7 +332,12 @@ fn add_bits(x: i32) -> i64 {
 
 #[inline(always)]
 fn error_weight(x: u32, maxweight: u32) -> u32 {
-    let shift = 0.max(floor_log2_nonzero(x as u64 + 1) as i32 - 5);
+    // Inline log2 to merge the XOR and subtraction by 5. 
+    // This avoids branching and permits 32-bit lzcnt instead of 64-bit. 
+    // It also unlocks some auto-vectorization, especially on ARM and AVX512 
+    // which have native 4x32 lzcnt (aligning with NUM_PREDICTORS). 
+    let carry = ((x & 0b11111) == 0b11111) as u32;
+    let shift = 31u32.saturating_sub(((x >> 5) + carry).leading_zeros());
     4u32 + ((maxweight * DIVLOOKUP[(x >> shift) as usize & 63]) >> shift)
 }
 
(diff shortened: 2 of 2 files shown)


TechPizzaDev:
Seems that [libjxl does the same](https://github.com/libjxl/libjxl/blob/1dfa3c26b5ac199135bddcb9d4d096ff12a7b837/lib/jxl/modular/encoding/context_predict.h#L105).

veluca93 (approved):

TechPizzaDev:
I'm not terribly happy with those CI benchmark regressions.
On my R7 7700X, `bicycles.jxl` barely regresses (~0.4%).
I'll play around a bit further, and maybe cross-compile to other targets to validate LLVM isn't doing anything too silly.

TechPizzaDev:
I just checked `aarch64-linux-gnu-gcc` locally, and it seems to produce the SIMD I'd expect it to, 
like `clz v0.4s, v0.4s` followed by `uqsub v0.4s, v2.4s, v0.4s` versus the previous unrolled loop.
Stack usage and code size seem fairly smaller on A64 too.

veluca93 (approved):

TechPizzaDev:
I don't feel good merging this with the few but big regressions (let alone all the CI regressions).
Hate to say it but this may be a dead end. The branch predictor is just too good.
Even with the seemingly nice auto-vectorization, it does not play well with scattered `DIVLOOKUP` gathers, 
so it's probably best left alone as just an auto-unrolled loop.

I didn't test it on ARM hardware, but I can't imagine a benefit there.
The only finding I came upon is Apple Silicon allegedly being super fast at int division; 
https://ridiculousfish.com/blog/posts/benchmarking-libdivide-m1-avx512.html.
No idea if other A64 chips are like this, but it would be funny to cfg! `DIVLOOKUP` behind x86 and use normal div on A64. 
From what I can tell, my Zen4 seems to perform better with idiv too, but using accurate div causes `SectionTooShort` errors :)

veluca93:
### Performance Summary (Commit `7d6960a`)

| desktop | Single | 77.06 | 77.16 | -0.91% ± 0.28% |
| desktop | Multi | 77.23 | 76.76 | -1.15% ± 0.34% |
| framework-desktop | Single | 90.84 | 91.64 | -0.98% ± 0.21% |
| framework-desktop | Multi | 91.55 | 92.02 | -1.10% ± 0.18% |
| pixel7a | Single (Fast) | 28.68 | 28.37 | -1.23% ± 0.36% |
| pixel7a | Single (Mid) | 20.68 | 20.17 | -1.48% ± 0.30% |
| pixel7a | Multi | 28.56 | 28.25 | -1.20% ± 0.35% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_789.html)

TechPizzaDev:
Cool, not as abysmal as I thought, but the percentage tells us it's not worth it, right?
Especially since the files with modular in the name regressed heavily.

And are the numbers supposed to be weird? All percentages are negative, but only pixel7a regressed in MP/s... (guessing it's large MP/s for some images offseting it)

veluca93:
This very likely suffers from different aggregation :-)
Hard for me to say whether this is a win or not (it seems at least somewhat hardware dependent). I'll be looking at the weighted predictor as part of reviewing/improving on #715 anyway, and I might take some inspiration from here :-)

TechPizzaDev:
Oh, I just noticed that @\hjanuschka already arrived at similar code in that PR to what I did today, that's funny. I left a comment/review there as a reminder.

PR #792, opened 2026-06-05
Add NEON support for ARMv7

FooIbar:
`compare_incremental_tirr_photo`, `compare_pipelines_progressive` and `compare_pipelines_tirr_photo` are skipped in CI due to OOM.

Local benchmark shows 2.7x speedup on `bike.jxl` (2.66 to 7.21 MP/s) and 2.2x speedup on `green_queen_vardct_e3.jxl` (3.79 to 8.47 MP/s).

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -67,6 +67,14 @@ jobs:
       matrix:
         features: [all, none]
         os: [ubuntu-24.04-arm, ubuntu-latest]
+        include:
+          - os: ubuntu-24.04-arm
+            nightly: true
+            features: neon
+            target: thumbv7neon-unknown-linux-gnueabihf
+          - os: ubuntu-latest
+            features: simd128
+            target: wasm32-wasip1
     steps:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -78,6 +86,7 @@ jobs:
         uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
         with:
           components: clippy
+          targets: ${{ matrix.target }}
 
       - name: Rust cache
         uses: Swatinem/rust-cache@98c8021b550208e191a6a3145459bfc9fb29c4c0 # v2.8.0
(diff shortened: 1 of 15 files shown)


.github/workflows/pull_request.yml
```diff
@@ -87,11 +96,17 @@ jobs:
 
       - name: Clippy with all features
         if: ${{ matrix.features == 'all' }}
-        run: cargo clippy --release --all-targets --all-features --tests --all -- -D warnings
+        run: cargo clippy --release --all-targets --all-features --all -- -D warnings
```
FooIbar, line 99: This is removed because [`--all-targets`](https://doc.rust-lang.org/cargo/commands/cargo-check.html#option-cargo-check---all-targets) include `tests`.

veluca93 (commented):
Some high-level thoughts: AFAIU NEON on ARMv7 is currently nightly because there are still some soundness issues with the backend. Thus, I'd prefer not to add the support unless we have some clear and important use cases.

If we do, then I imagine it should be fine if we rename the "nightly" feature to something a bit more scary ("enable-potentially-broken-targets"?) and leave it off by default. As a side note, why do we need the feature flag on jxl_transforms?

FooIbar:
> AFAIU NEON on ARMv7 is currently nightly because there are still some soundness issues with the backend.

I'm aware of that, but not sure how it would affect us. Does jxl-rs rely on subnormals?

> As a side note, why do we need the feature flag on jxl_transforms?

It's necessary because `jxl_transforms` calls `simd_function!`, which requires `#![feature(arm_target_feature)]`.

veluca93:
> > AFAIU NEON on ARMv7 is currently nightly because there are still some soundness issues with the backend.
> 
> I'm aware of that, but not sure how it would affect us. Does jxl-rs rely on subnormals?
> 
> > As a side note, why do we need the feature flag on jxl_transforms?
> 
> It's necessary because `jxl_transforms` calls `simd_function!`, which requires `#![feature(arm_target_feature)]`.

I dislike soundness issues as a general rule :-) 

Also, doesn't the `jxl` crate itself also call `simd_function!`? What is different?

FooIbar:
> Also, doesn't the `jxl` crate itself also call `simd_function!`? What is different?

That's why the feature flag is added to the `jxl` crate as well

veluca93:
> > Also, doesn't the `jxl` crate itself also call `simd_function!`? What is different?
> 
> That's why the feature flag is added to the `jxl` crate as well

My bad, I missed that...

PR #795, opened 2026-06-08
Reduce memory use for progressive lossless images

hjanuschka:
Fixes #782.

The inverse squeeze transforms read two neighbor grids per step (the next average and the previously decoded output), which the transform graph counts as buffer uses, but `do_run` only released the primary inputs. So those intermediate modular buffers stayed allocated for the whole frame. This releases the neighbor grids on the final render, mirroring how the graph registers them (with a dedup guard for when a coarser average channel maps the neighbor onto the same grid index).

`jxl_cli` also no longer stores progressive render snapshots when there's no output path to write them to.

Peak RSS on the repro:

| scenario | before | after |
|---|---:|---:|
| decode (`-s`) | 2740 MiB | 1191 MiB |
| `--render-interval 1000` | 4208 MiB | 3626 MiB |
| `--render-interval 250` | 7411 MiB | 5815 MiB |

For full decode that's now below djxl (~1516 MiB) and ~1.5x jxl-oxide (~792 MiB), down from 3.4x.

Issue #782, opened 2026-05-29
High memory use for progressive lossless images


jonnyawsom3:
While testing progressive loading in Chrome, I noticed 6 GB of memory used while partially loading a 50MP image.
jxl-rs against jxl-oxide and libjxl shows 2-4x higher memory usage depending on the image. [Test.zip](https://github.com/user-attachments/files/28411514/Test.zip)
```
wintime -- jxl-oxide --num-threads 1 Test.jxl
2026-05-29T22:30:19.902780Z  INFO jxl_oxide_cli::decode: Image dimension: 8192x8192
2026-05-29T22:30:20.287561Z  INFO jxl_oxide_cli::decode: Took 384.68 ms (174.46 MP/s)
2026-05-29T22:30:20.287751Z  INFO jxl_oxide_cli::decode: No output path specified, skipping output encoding
PageFaultCount: 203330
PeakWorkingSetSize: 792.2 MiB
QuotaPeakPagedPoolUsage: 46.78 KiB
QuotaPeakNonPagedPoolUsage: 15.99 KiB
PeakPagefileUsage: 859 MiB
Creation time 2026/05/29 23:30:19.896
    Exit time 2026/05/29 23:30:20.309
    Wall time:  0 days, 00:00:00.412 (0.41 seconds)
    User time:  0 days, 00:00:00.343 (0.34 seconds)
  Kernel time:  0 days, 00:00:00.062 (0.06 seconds)

wintime -- djxl --num_threads 0 --disable_output Test.jxl
JPEG XL decoder v0.12.0 ceeb489 [_AVX2_,SSE4,SSE2] {Clang 21.1.8}
Decoded to pixels.
8192 x 8192, 29.744 MP/s, 0 threads.
PageFaultCount: 859029
PeakWorkingSetSize: 1.516 GiB
QuotaPeakPagedPoolUsage: 44.64 KiB
QuotaPeakNonPagedPoolUsage: 9.953 KiB
PeakPagefileUsage: 1.908 GiB
Creation time 2026/05/29 23:31:08.997
    Exit time 2026/05/29 23:31:11.324
    Wall time:  0 days, 00:00:02.327 (2.33 seconds)
    User time:  0 days, 00:00:01.500 (1.50 seconds)
  Kernel time:  0 days, 00:00:00.828 (0.83 seconds)

wintime -- jxl_cli -s Test.jxl
Decoded 67108864 pixels in 3.074 seconds: 21.833 MP/s
[...]

Issue #771, opened 2026-05-09
`flush_pixels` does not work with partial non-section data


FooIbar:
If `NeedsMoreInput` is returned while parsing non-section data (causing `codestream_parser.sections` to be empty), `flush_pixels` will fail immediately without rendering anything.

```
jxl_cli --allow-partial-files fox-truncated.jxl output.png
```
[fox-truncated.txt](https://github.com/user-attachments/files/27552136/fox-truncated.txt) has a complete LF frame but it won't be rendered.

hjanuschka:
most likely fix with https://github.com/libjxl/jxl-rs/pull/794 ?

FooIbar:
> most likely fix with [#794](https://github.com/libjxl/jxl-rs/pull/794) ?

I don't think so, that patch does not change the logic.
The problem is `process` only renders LF frames in the section branch, but since `self.sections` is empty, it will take the non-section branch and error out with `NeedsMoreInput` again.

PR #796, opened 2026-06-08
WIP: Interleave modular transform processing

hjanuschka:
WIP follow-up to #795.

This tries the deeper path for #782: for safe gridded Modular frames, decode a group and immediately run any modular transforms that became ready, instead of decoding all groups before transform processing. This lets large squeeze images free intermediates during decode.

The optimization is gated to straight non-flush Modular decodes, before any flush happened, gridded outputs, and sufficiently tiled frames. Flush/progressive and small-frame paths keep the existing batched behavior.

Peak RSS on the repro:

| decoder | peak RSS |
|---|---:|
| jxl-rs main | 2739 MiB |
| #795 | 1198 MiB |
| this branch | 333 MiB |
| jxl-oxide 0.12.6 | 421 MiB |
| djxl -> PPM | 1657 MiB |

This avoids the global scratch-pool cap from #795 by only skipping center-buffer caching while interleaved modular output is actively feeding the low-memory pipeline. Normal scratch-buffer reuse is left unchanged.

### DIFF
--- a/jxl/src/frame/modular/mod.rs
+++ b/jxl/src/frame/modular/mod.rs
@@ -961,6 +961,12 @@ impl FullModularImage {
         Ok(())
     }
 
+    pub fn outputs_are_gridded(&self) -> bool {
+        self.buffer_info
+            .iter()
+            .all(|b| b.info.output_channel_idx.is_none() || b.grid_kind != ModularGridKind::None)
+    }
+
     pub fn channel_range(&self) -> Range<usize> {
         if self.modular_color_channels != 0 {
             0..self.buffers_for_channels.len()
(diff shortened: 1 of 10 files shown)


veluca93:
I had a different approach in mind that *might* be simpler and more effective during progressive renders. I will give that some thought and hopefully write something up in the next day or two :-)

veluca93 (commented):
approving for benchmark purposes ;-)

veluca93 (approved):

veluca93:
### Performance Summary (Commit `a42d5b2`)

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_796.html)

jonnyawsom3:
> I had a different approach in mind that _might_ be simpler and more effective during progressive renders. I will give that some thought and hopefully write something up in the next day or two :-)

No rush, but wanted to check if there's been any progress on that
Been bumping into #771 while testing and memory limits on my phone occasionally

veluca93:
> > I had a different approach in mind that _might_ be simpler and more effective during progressive renders. I will give that some thought and hopefully write something up in the next day or two :-)
> 
> No rush, but wanted to check if there's been any progress on that Been bumping into #771 while testing and memory limits on my phone occasionally

Working on it!

PR #800, opened 2026-06-10
Render the full LF preview on its first flush.

tnikkel:
`maybe_preview_lf_frame` rendered the whole image only when called with
`changed_regions = None`, otherwise just those regions. Its only caller always
passed `Some(&regions[..])`, so the full-image path was never taken and
`lf_frame_was_rendered` (only set in the `None` branch) was never set.

Streaming still looks right because the per-flush changed regions accumulate in
the retained output buffer. But a truncated / one-shot decode only flushes once,
so the preview covers a few regions and the rest stays at the buffer's init
(black).

Render the whole image on the first flush of an LF frame and only the changed
regions afterwards (keyed on `lf_frame_was_rendered`), and drop the now-unused
`Option` parameter. Truncated decodes now show the full LF preview; full decodes
are unchanged.

Repro: a `three-pass.jxl` truncated to 220 KB decoded with
`jxl_cli --allow-partial-files` shows mostly black before / the full blurry
preview after. Full (untruncated) decode is byte-identical (PSNR ∞).

### DIFF
--- a/jxl/src/frame/lf_preview.rs
+++ b/jxl/src/frame/lf_preview.rs
@@ -280,7 +280,7 @@ impl Frame {
         &mut self,
         pixel_format: &JxlPixelFormat,
         output_buffers: &mut [JxlOutputBuffer<'_>],
-        changed_regions: Option<&[Rect]>,
+        changed_regions: &[Rect],
         output_profile: &JxlColorProfile,
     ) -> Result<bool> {
         if self.header.needs_blending() {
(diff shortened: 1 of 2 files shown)


veluca93 (approved):

veluca93:
### Performance Summary (Commit `d9645d2`)

| desktop | Single | 81.08 | 80.62 | -0.07% ± 0.35% |
| desktop | Multi | 80.78 | 80.47 | +0.01% ± 0.37% |
| framework-desktop | Single | 94.17 | 94.38 | +0.08% ± 0.24% |
| framework-desktop | Multi | 95.19 | 95.37 | +0.02% ± 0.28% |
| pixel7a | Single (Fast) | 28.84 | 29.33 | -0.09% ± 0.41% |
| pixel7a | Single (Mid) | 21.42 | 20.89 | -0.72% ± 0.39% |
| pixel7a | Multi | 28.68 | 28.55 | -0.82% ± 0.40% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_800.html)

PR #802, opened 2026-06-12
do two TODOs

jonsneyers:
Does two TODOs, one small one in tf.rs, and a bigger one to use 16-bit HF coeff buffers (only relevant in case of progressive passes) instead of 32-bit ones.

Gives only an almost immeasurable speedup on my laptop, but a substantial memory reduction in case of multiple HF passes:

Before:
```
Decoded 333921030 pixels in 6.153 seconds: 54.271 MP/s
        6.39s real              6.35s user              0.02s sys
           236355584  maximum resident set size
```

After:
```
Decoded 333921030 pixels in 6.147 seconds: 54.322 MP/s
        6.38s real              6.35s user              0.01s sys
           154320896  maximum resident set size
```

### DIFF
--- a/jxl/src/frame/group.rs
+++ b/jxl/src/frame/group.rs
@@ -76,8 +76,6 @@ fn predict_num_nonzeros(nzeros_map: &Image<u32>, bx: usize, by: usize) -> usize 
     }
 }
 
-/// Abstraction over coefficient slice types (i32 or i16) for zero-copy SIMD dequant.
-/// Implemented for both `&[i32]` (direct) and `&[i16]` (sign-extend via load_from_i16).
 trait CoeffSlice<D: SimdDescriptor>: Copy {
     fn load_vec(self, d: D, k: usize) -> D::I32Vec;
 }
@@ -498,9 +496,7 @@ pub fn decode_vardct_group(
     let raw_quant_map = hf_meta.raw_quant_map.get_rect(block_group_rect);
     let quant_lf_rect = quant_lf.get_rect(block_group_rect);
     let block_context_map = lf_global.block_context_map.as_mut().unwrap();
-    // Coefficient storage: i16 when the histogram guarantees all values fit (saves ~50% RAM for
-    // persistent multi-pass coefficients). In the i16 case we expand per block before dequant
-    // using the scratch coeffs_storage buffer (which is unused in multi-pass mode).
+
     let mut coeffs_i16: Option<[&mut [i16]; 3]>;
     let mut coeffs_i32: Option<[&mut [i32]; 3]>;
     match hf_global.hf_coefficients.as_mut() {
--- a/jxl/src/frame/group.rs
+++ b/jxl/src/frame/group.rs
@@ -82,14 +82,14 @@ trait CoeffSlice<D: SimdDescriptor>: Copy {
     fn load_vec(self, d: D, k: usize) -> D::I32Vec;
 }
 
-impl<'a, D: SimdDescriptor> CoeffSlice<D> for &'a [i32] {
+impl<D: SimdDescriptor> CoeffSlice<D> for &[i32] {
     #[inline(always)]
     fn load_vec(self, d: D, k: usize) -> D::I32Vec {
         D::I32Vec::load(d, &self[k..])
     }
 }
 
-impl<'a, D: SimdDescriptor> CoeffSlice<D> for &'a [i16] {
+impl<D: SimdDescriptor> CoeffSlice<D> for &[i16] {
     #[inline(always)]
     fn load_vec(self, d: D, k: usize) -> D::I32Vec {
         D::I32Vec::load_from_i16(d, &self[k..])
@@ -515,8 +515,7 @@ pub fn decode_vardct_group(
         None => {
             // Use pooled buffer (already reset to zero in buffers.reset() above)
             coeffs_i16 = None;
-            let (coeffs_x, coeffs_y_b) =
-                buffers.coeffs_storage.split_at_mut(GROUP_DIM * GROUP_DIM);
+            let (coeffs_x, coeffs_y_b) = buffers.coeffs_storage.split_at_mut(GROUP_DIM * GROUP_DIM);
             let (coeffs_y, coeffs_b) = coeffs_y_b.split_at_mut(GROUP_DIM * GROUP_DIM);
             coeffs_i32 = Some([coeffs_x, coeffs_y, coeffs_b]);
         }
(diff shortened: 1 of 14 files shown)


jxl/src/frame/mod.rs
```diff
@@ -66,11 +66,16 @@ pub struct PassState {
...
+    I32(Image<i32>, Image<i32>, Image<i32>),
+    I16(Image<i16>, Image<i16>, Image<i16>),
+}
+
 pub struct HfGlobalState {
     num_histograms: u32,
     passes: Vec<PassState>,
     dequant_matrices: DequantMatrices,
-    hf_coefficients: Option<(Image<i32>, Image<i32>, Image<i32>)>,
+    pub(super) hf_coefficients: Option<HfCoefficients>,
```
veluca93: If we're doing a match anyway, why not make HfCoefficients have a None / I16(...) / I32(...) enum? Also, any reason not to make the tuple into an array of size 3?

jxl_simd/src/aarch64/neon.rs
```diff
@@ -626,6 +626,15 @@ impl I32SimdVec for I32VecNeon {
         Self(unsafe { vld1q_s32(mem.as_ptr()) }, d)
     }
 
+    #[inline(always)]
+    fn load_from_i16(d: Self::Descriptor, mem: &[i16]) -> Self {
+        assert!(mem.len() >= Self::LEN);
+        // SAFETY: we just checked that `mem` has enough space. Moreover, we know neon is available
+        // from the safety invariant on `d`. vld1_s16 loads 4 i16 values; vmovl_s16 sign-extends
+        // them to 4 i32 values.
```
veluca93: "vld1_s16 loads 4 i16 values; vmovl_s16 sign-extends them to 4 i32 values." should not be part of the safety comment.

veluca93: (and similarly in other implementations)

jxl/src/frame/group.rs
```diff
@@ -571,60 +657,113 @@ pub fn decode_vardct_group(
...
-                    let current_coeffs = &mut coeffs[c][coeffs_offset..coeffs_offset + num_coeffs];
                     // Asserting once lets the compiler elide the bounds check on
                     // `permutation[k]` inside the loop given `k < num_coeffs`.
-                    assert!(permutation.len() >= num_coeffs);
-                    for k in num_blocks..num_coeffs {
-                        if nonzeros == 0 {
-                            break;
+                    if let Some(ref mut ci16) = coeffs_i16 {
+                        let current = &mut ci16[c][coeffs_offset..coeffs_offset + num_coeffs];
+                        assert!(permutation.len() >= num_coeffs);
```
veluca93: This assert seems misplaced now.

jxl/src/frame/group.rs
```diff
@@ -571,60 +657,113 @@ pub fn decode_vardct_group(
...
+                                break;
+                            }
+                            let ctx = histo_offset
+                                + zero_density_context(nonzeros, k, log_num_blocks, prev);
+                            let coeff =
+                                reader.read_signed_inline(&pass_info.histograms, br, ctx) << *shift;
+                            prev = if coeff != 0 { 1 } else { 0 };
+                            nonzeros -= prev;
+                            let coeff_index = permutation[k] as usize;
+                            current[coeff_index] = current[coeff_index].wrapping_add(coeff as i16);
```
veluca93: Didn't we do checks to be sure this wouldn't overflow? if so, we should `+` and not `wrapping_add`.

jxl/src/frame/group.rs
```diff
@@ -196,10 +216,10 @@ fn dequant_and_transform_to_pixels<D: SimdDescriptor>(
     block_rect: Rect,
     num_blocks: usize,
     num_coeffs: usize,
-    qblock: &[&[i32]; 3],
+    qblock: [C; 3],
```
veluca93: I think it would be better, code-size wise, to instead pass an enum here, which holds either a `[&[i32]; 3]` or a `[&[i16]; 3]`, and then only specialize the code for `dequant_block`.

jxl/src/frame/group.rs
```diff
@@ -420,20 +498,28 @@ pub fn decode_vardct_group(
...
-    let coeffs = match hf_global.hf_coefficients.as_mut() {
-        Some(hf_coefficients) => [
-            hf_coefficients.0.row_mut(group),
-            hf_coefficients.1.row_mut(group),
-            hf_coefficients.2.row_mut(group),
-        ],
+    // Coefficient storage: i16 when the histogram guarantees all values fit (saves ~50% RAM for
+    // persistent multi-pass coefficients). In the i16 case we expand per block before dequant
+    // using the scratch coeffs_storage buffer (which is unused in multi-pass mode).
+    let mut coeffs_i16: Option<[&mut [i16]; 3]>;
```
veluca93: (see comment on the enum)

jxl/src/frame/mod.rs
veluca93: Alternatively, have you considered using three `OwnedRawImage`s and a boolean flag to remember whether it is i32 or i16? This should have a bit less code.

veluca93 (approved):
Approving to get benchmark results as usual ;)

veluca93:
### Performance Summary (Commit `d4958fd`)

| desktop | Single | 79.50 | 79.36 | -0.43% ± 0.33% |
| desktop | Multi | 79.11 | 79.11 | -0.23% ± 0.39% |
| framework-desktop | Single | 92.68 | 96.39 | -0.32% ± 0.30% |
| framework-desktop | Multi | 93.50 | 97.07 | -0.32% ± 0.20% |
| pixel7a | Single (Fast) | 28.57 | 28.29 | +0.13% ± 0.33% |
| pixel7a | Single (Mid) | 20.37 | 20.86 | +1.00% ± 0.37% |
| pixel7a | Multi | 28.47 | 28.67 | +0.51% ± 0.54% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_802.html)

jxl/src/frame/mod.rs
jonsneyers: Did the enum thing. With `OwnedRawImage` it seems hard to avoid unsafe reinterpret casts, no?

jonsneyers:
> ### Performance Summary (Commit `0aabc58`)
> [...]

Probably this is just noise since there's no image where this should make a difference (requires multiple HF passes).
Let's take another look after adding a multi-pass test image.

jxl/src/frame/mod.rs
veluca93: It shouldn't be, there is a safe way to get a typed image view out of a OwnedRawImage ;)
See https://github.com/libjxl/jxl-rs/blob/36c9c3b43ee372a09e82bfed8ca2b75f5a2ec134/jxl/src/image/typed.rs#L113

veluca93: and https://github.com/libjxl/jxl-rs/blob/36c9c3b43ee372a09e82bfed8ca2b75f5a2ec134/jxl/src/image/typed.rs#L265 for the rect version

veluca93:
> > ### Performance Summary (Commit `0aabc58`)
> > [...]
> 
> Probably this is just noise since there's no image where this should make a difference (requires multiple HF passes). Let's take another look after adding a multi-pass test image.

It seems to help a bit with progressive images.

veluca93:
Most of this got done in another PR so closing :-)

PR #808, opened 2026-06-20
Return error when flushing without pixel format

hjanuschka:
Chromium's blink_jxl_decoder_fuzzer found that `flush_pixels` can be reached before the inner decoder has reached the point where the output pixel format is set. That path panicked at `pixel_format.unwrap()` instead of returning a decoder error.

Return a normal `PixelFormatNotSet` error instead, which callers can already handle as decode failure. The regression test covers `flush_pixels` before pixel format setup.

https://issues.chromium.org/issues/526010666

### DIFF
--- a/jxl/src/api/decoder.rs
+++ b/jxl/src/api/decoder.rs
@@ -2034,6 +2034,16 @@ pub(crate) mod tests {
         }
     }
 
+    /// Regression test for Chromium ClusterFuzz issue 526010666.
+    #[test]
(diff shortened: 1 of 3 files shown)


veluca93:
I'm not sure this is the correct fix - this sounds like an API usage error on the Chrome side.

How hard is it to stop Chrome from calling flush_pixels before pixel format is set? (i.e. before image info is available)

hjanuschka:
https://chromium-review.googlesource.com/c/chromium/src/+/7965920 


feel free to close, if having it as a additional defensive guard in the lib or not

PR #810, opened 2026-06-26
Support CMYK -> RGB conversion in jxl_cli

FooIbar:


jxl_cms/src/lcms2.rs
```diff
@@ -103,6 +109,17 @@ impl JxlCmsTransformer for Lcms2Transformer {
             ));
         }
 
+        // We use [0 = max ink, 1], but lcms2 expects [0 = no ink, 100].
+        let input = if let Some(buf) = &mut self.cmyk_buffer {
+            buf.resize(input.len(), 0.0);
+            for (dst, &src) in buf.iter_mut().zip(input.iter()) {
+                *dst = 100.0 - src * 100.0;
```
sboukortt:
```suggestion
                *dst = 100.0 * (1.0 - src);
```

would be numerically more stable.

sboukortt (approved):
Thanks!

jxl_cms/src/lcms2.rs
FooIbar: Done

PR #812, opened 2026-07-07
Rewrite transform application in Modular decoding

veluca93:
This has the same benefits and thus closes #796, as well as part of the improvements of #795;  a follow-up to handle deallocation of temporary images should also handle #782, although this PR as is should already improve memory and probably CPU usage for progressive lossless by avoiding re-rendering unused buffers.

The general approach is that, for each decode step, we build a DAG of those transforms that we want to run *during this step* (either because they would produce the final buffers, or because we want to update the pixels in a certain image region). This DAG is related to, but not a subset of, the DAG of final transforms (the difference comes from smooth upsampling of squeeze'd channels).

The PR is split in a few commits with more or less independent changes to facilitate review.


veluca93:
### Performance Summary (Commit `fab1137`)

| desktop | Single | 78.97 | 84.30 | +5.45% ± 0.30% |
| desktop | Multi | 78.77 | 84.34 | +5.75% ± 0.40% |
| framework-desktop | Single | 92.10 | 94.49 | +3.03% ± 0.28% |
| framework-desktop | Multi | 92.80 | 95.42 | +3.09% ± 0.23% |
| pixel7a | Single (Fast) | 26.94 | 26.83 | +1.76% ± 0.71% |
| pixel7a | Single (Mid) | 20.77 | 19.53 | +0.33% ± 0.29% |
| pixel7a | Multi | 26.91 | 26.55 | +1.43% ± 0.28% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_812.html)

jonsneyers:
Do we have a way to see the effect on intermediate previews during progressive decode? Conformance testing and the other tests give some confidence that this doesn't introduce bugs on the final image, but I wonder if there's a way to check what's happening at different stages of progressive decode, e.g. if this doesn't introduce regressions where an intermediate image becomes black in some regions or things like that.
Would it be possible to generate filmstrips showing the previews in steps of, say, 10% of the bitstream loaded, with one row for 'before' and one row for 'after'? This would make it easy to visually check that there are no regressions.

veluca93:
> Do we have a way to see the effect on intermediate previews during progressive decode? Conformance testing and the other tests give some confidence that this doesn't introduce bugs on the final image, but I wonder if there's a way to check what's happening at different stages of progressive decode, e.g. if this doesn't introduce regressions where an intermediate image becomes black in some regions or things like that.
> Would it be possible to generate filmstrips showing the previews in steps of, say, 10% of the bitstream loaded, with one row for 'before' and one row for 'after'? This would make it easy to visually check that there are no regressions.

There's a script to generate a video at various loading stages, that I ran over a few images and saw the previews look reasonable. We should probably have a way to automate that check though...

jonnyawsom3:
I compared some intermediate progressive renders and there's errors on group boundaries, but generally not very noticeable during viewing and the final image is still identical, so may be fine.

```
Decoded 8294400 pixels in 2.406 seconds: 3.447 MP/s
PeakPagefileUsage: 2.805 GiB
    Wall time:  0 days, 00:00:08.359 (8.36 seconds)
    User time:  0 days, 00:00:06.062 (6.06 seconds)
  Kernel time:  0 days, 00:00:02.296 (2.30 seconds)
  
Decoded 8294400 pixels in 1.348 seconds: 6.151 MP/s
PeakPagefileUsage: 1.53 GiB
    Wall time:  0 days, 00:00:06.886 (6.89 seconds)
    User time:  0 days, 00:00:05.625 (5.62 seconds)
  Kernel time:  0 days, 00:00:01.250 (1.25 seconds)
  ```
  Actual CPU usage only went down slightly, but decode speed nearly doubled when decoding to 50 steps, probably thanks to the memory reduction.
  
  Trying the test image in #782
  ```
  Decoded 67108864 pixels in 2.015 seconds: 33.301 MP/s
PeakPagefileUsage: 2.688 GiB
    Wall time:  0 days, 00:00:04.001 (4.00 seconds)
    User time:  0 days, 00:00:02.812 (2.81 seconds)
  Kernel time:  0 days, 00:00:01.187 (1.19 seconds)
  
  Decoded 67108864 pixels in 1.575 seconds: 42.609 MP/s
PeakPagefileUsage: 1.872 GiB
    Wall time:  0 days, 00:00:03.209 (3.21 seconds)
    User time:  0 days, 00:00:02.218 (2.22 seconds)
  Kernel time:  0 days, 00:00:00.953 (0.95 seconds)
  ```
  Not as much of a speed improvement, probably due to already using a fast path, but still 1/3 less memory so far.

jonsneyers (approved):

PR #813, opened 2026-07-08
Less strict ICC size limit

sboukortt:
1 MiB is arguably quite tight.


jonnyawsom3:
I was going to say even 16 MB might still bump into issues, but in https://github.com/libjxl/libjxl/pull/4225 I made >16 MB the threshold for using faster compression, so that actually works out nicely.

veluca93 (approved):

PR #842, opened 2026-07-30
Add support for multithreaded decoding

veluca93:
This PR is for initial feedback and benchmarking, some cleanup is still needed.


veluca93:
### Performance Summary (Commit `e83e7a8`)

| framework-desktop | Single | 100.14 | 88.34 | -7.97% ± 0.97% |
| framework-desktop | Multi | 100.37 | 517.12 | +402.41% ± 3.43% |
| pixel7a | Single (Fast) | 27.27 | 26.49 | -1.46% ± 0.24% |
| pixel7a | Single (Mid) | 19.79 | 19.38 | -1.03% ± 0.28% |
| pixel7a | Multi | 26.94 | 39.77 | +85.35% ± 3.12% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_842.html)

jxl/src/util/per_thread_storage.rs
```diff
@@ -26,56 +26,41 @@ impl<T> Debug for PerThreadStorage<T> {
...
+            if let Some(mut a) = self.storage.try_borrow_mut() {
+                if let Some(x) = a.pop() {
+                    break x;
+                }
+                // go to the call to `init()` if storage is empty
+            } else {
+                continue;
             }
+            break (self.init)();
         };
```
sboukortt, line 50:
```suggestion
        let t = loop {
            if let Some(mut a) = self.storage.try_borrow_mut() {
                break a.pop().unwrap_or_else(self.init);
            }
        };
```

Or am I missing something?

veluca93: That's not equivalent - it holds the lock on self.storage during the call to self.init.

jxl/src/render/mod.rs
```diff
@@ -42,13 +42,15 @@ pub(crate) use low_memory_pipeline::LowMemoryRenderPipeline;
 #[cfg(test)]
 pub(crate) use simple_pipeline::SimpleRenderPipeline;
 
+pub(crate) type ErasedLocalState = dyn Any + Send + Sync;
```
tirr-c, line 45: Do we need `Send + Sync` for thread local state?

veluca93: Something somewhere wants it, and there's nothing that we want to put there that is not Sync so...

PR #845, opened 2026-08-01
Fix group border detection in render pipeline

veluca93:
The second-to-last group in a row could still include the image border if the amount of border is greater than the last group.

This PR makes the computation of "are we on a border" more precise.


veluca93:
### Performance Summary (Commit `d1fdb48`)

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_845.html)

tirr-c (approved):

PR #850, opened 2026-08-01
Improve handling of input buffers in the render pipeline

veluca93:
The two commits in this PR:
1) separate part of the fill_input_buffer logic out to a precomputation phase
2) switch input buffers to AtomicRefCell
3) add an assertion that the buffer is not ready when setting a channel in it, and fix a lack of mark_group_to_rerender that caused the assert to trigger.

### DIFF
--- a/jxl/src/render/low_memory_pipeline/input_buffers.rs
+++ b/jxl/src/render/low_memory_pipeline/input_buffers.rs
@@ -1,0 +1,211 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
+
+use crate::error::Result;
+use crate::image::OwnedRawImage;
+use crate::render::internal::RenderPipelineShared;
+use crate::render::low_memory_pipeline::row_buffers::RowBuffer;
+use crate::util::{AtomicRefCell, NewWithCapacity};
+
+pub(super) struct InputBuffer {
+    // One buffer per channel.
+    pub(super) data: Vec<AtomicRefCell<Option<OwnedRawImage>>>,
+    // Storage for left/right borders. Includes corners.
+    pub(super) leftright: Vec<AtomicRefCell<Option<OwnedRawImage>>>,
+    // Storage for top/bottom borders. Includes corners.
+    pub(super) topbottom: Vec<AtomicRefCell<Option<OwnedRawImage>>>,
+    // Number of ready channels in the current pass.
+    pub(super) ready_channels: AtomicUsize,
+    is_ready: AtomicBool,
+}
+
+impl InputBuffer {
+    pub(super) fn set_buffer(&self, chan: usize, buf: OwnedRawImage) {
+        assert!(!self.is_ready.load(Ordering::Relaxed));
+        assert!(self.data[chan].borrow_mut().is_none(), "chan: {chan}");
+        *self.data[chan].borrow_mut() = Some(buf);
+        self.ready_channels.fetch_add(1, Ordering::Relaxed);
+    }
+
+    pub(super) fn new(num_channels: usize) -> Self {
+        let b = || {
+            (0..num_channels)
+                .map(|_| AtomicRefCell::new(None))
+                .collect()
+        };
+        Self {
+            data: b(),
+            leftright: b(),
+            topbottom: b(),
+            ready_channels: AtomicUsize::new(0),
+            is_ready: AtomicBool::new(false),
+        }
+    }
+}
+
+pub(super) struct InputBuffers {
+    buffers: Vec<InputBuffer>,
+    // We use the cell that would map to the buffer itself to keep
+    // track of the number of complete groups in its neighbourhood.
+    remaining_flags: Vec<AtomicU8>,
+    size: (usize, usize),
+}
+
+impl InputBuffers {
+    fn increase_borders(flags: &[AtomicU8], (gx, gy): (usize, usize), (xs, _): (usize, usize)) {
+        let stride = xs * 2 + 1;
+        let center = (gy * 2 + 1) * stride + gx * 2 + 1;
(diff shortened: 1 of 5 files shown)


veluca93:
### Performance Summary (Commit `28ce86e`)

| desktop | Single | 85.75 | 86.86 | +0.98% ± 0.22% |
| desktop | Multi | 85.64 | 86.82 | +1.03% ± 0.23% |
| framework-desktop | Single | 100.15 | 96.79 | -0.14% ± 0.27% |
| framework-desktop | Multi | 100.74 | 96.77 | -0.20% ± 0.25% |
| pixel7a | Single (Fast) | 27.36 | 26.99 | -0.66% ± 0.30% |
| pixel7a | Single (Mid) | 19.75 | 19.65 | -0.10% ± 0.23% |
| pixel7a | Multi | 27.06 | 26.85 | -0.55% ± 0.30% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_850.html)

tirr-c (approved):

jxl/src/render/low_memory_pipeline/input_buffers.rs
```diff
@@ -0,0 +1,211 @@
...
+        flags[center + 1].fetch_add(1, Ordering::Relaxed);
+        flags[center + stride - 1].fetch_add(1, Ordering::Relaxed);
+        flags[center + stride].fetch_add(1, Ordering::Relaxed);
+        flags[center + stride + 1].fetch_add(1, Ordering::Relaxed);
+    }
+
+    pub fn mark_not_ready(&mut self, group: usize) {
+        if !self.buffers[group]
+            .is_ready
+            .fetch_and(false, Ordering::Relaxed)
```
robryk, line 75: Nit: `swap` also works (and might be simpler on some platforms).

jxl/src/render/low_memory_pipeline/input_buffers.rs
```diff
@@ -0,0 +1,211 @@
...
+        let gx = group % self.size.0;
+        let gy = group / self.size.0;
+        Self::increase_borders(&self.remaining_flags, (gx, gy), self.size);
+    }
+
+    pub fn mark_ready(&self, group: usize) -> [bool; 9] {
+        assert!(
+            !self.buffers[group]
+                .is_ready
+                .fetch_or(true, Ordering::Relaxed)
```
robryk, line 88: Same

jxl/src/render/low_memory_pipeline/input_buffers.rs
```diff
@@ -0,0 +1,211 @@
...
+                gy * gw + gx,
+                gy * gw + gxp1,
+                gyp1 * gw + gxm1,
+                gyp1 * gw + gx,
+                gyp1 * gw + gxp1,
+            ] {
+                let gx = g % self.size.0;
+                let gy = g / self.size.0;
+                let idx = (gy * 2 + 1) * (2 * self.size.0 + 1) + gx * 2 + 1;
+                if self.remaining_flags[idx].fetch_add(1, Ordering::AcqRel) != 8 {
```
robryk, line 169: Nit: it's confusing that the direction in which things progress is opposite on centers and edges/corners.

jxl/src/render/low_memory_pipeline/input_buffers.rs
```diff
@@ -0,0 +1,211 @@
...
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
+
+use crate::error::Result;
+use crate::image::OwnedRawImage;
+use crate::render::internal::RenderPipelineShared;
+use crate::render::low_memory_pipeline::row_buffers::RowBuffer;
+use crate::util::{AtomicRefCell, NewWithCapacity};
+
+pub(super) struct InputBuffer {
+    // One buffer per channel.
```
robryk, line 15: How is access to everything here synchronized (i.e. what prevents a reader from seeing very out-of-order contents of different fields)?

I'd naively expect that is_ready plays that role, by being release-written after writing and acquire-read before reading, but (a) it's relaxed-{read,written} (b) it's sometimes unset (and so something else then takes the role of ensuring that writes are strictly after all reads that should be reading pre-that-write values?)

I suspect that the observed ordering between storing in `data` and changing `ready_channels`/`is_ready` might matter (with former being release and latter being relaxed, it's perfectly fine to reorder latter before former).

veluca93: is_ready is only set to false from a single thread, after which individual threads start setting it to true.

At some point I was thinking that only a single thread would simultaneously call set_buffer for the same group, but that's certainly not true now. I believe this means that we should make https://github.com/libjxl/jxl-rs/blob/5a689fc31812100053fa075f5446722ef5201a3d/jxl/src/render/low_memory_pipeline/input_buffers.rs#L32 into a AcqRel fetch_add, right?

robryk: You are pointing at a commit (5a689fc) that's not in this PR. I wrote everything below before I realized that. Which diff should I be looking at?

> is_ready is only set to false from a single thread, after which individual threads start setting it to true.

I assume by "after" you mean "there's a h-b path inbetween". If so, that should make it safe to set it relaxedly. But also, if you can convince the borrow checker of that, you can use `get_mut` on the atomic instead.

> I believe this means that we should make (...) into a AcqRel fetch_add, right?

What matters here is not the relationship between various set_buffer calls, but the desired relationship between them and the only place that makes decisions based on ready_channels in `render_with_new_group`. My wild guess is that:

1. You want `render_with_new_group`, if it sees `ready_channels` to be full to also see all the channels. Barring other h-b source that requires the increments to be release and the read in `render_...` to be acquire.
2. You have some way to ensure that `render_with_...` only runs after, if all channels were to be ready in this ~step, they had an opportunity to be ready (because otherwise whether it sees all channels as ready or not is up to scheduling vagaries). If so, that does provide h-b that you need.

robryk: Ok, so with that change on the second topic: IIUC you want to do something exactly once after all channels are ready. The fetch-and-add that counts ready channels should be acqrel and that's enough for it to work.

PR #851, opened 2026-08-02
Make the render pipeline interface MT-friendly

veluca93:
Includes #850.

This PR adds logic to BufferSplitter to allow simultaneously mutably-borrowing non-overlapping rects from a &BufferSplitter, and changes the render pipeline interface to not use &mut for methods that are meant to be MT-compatible.

### DIFF
--- a/jxl/src/render/buffer_splitter.rs
+++ b/jxl/src/render/buffer_splitter.rs
@@ -3,7 +3,97 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-use crate::{api::JxlOutputBuffer, headers::Orientation, image::Rect, util::ShiftRightCeil};
+use std::{
+    collections::BTreeSet,
+    ops::{Deref, DerefMut},
+};
+
+use crate::{
+    api::JxlOutputBuffer,
+    headers::Orientation,
+    image::Rect,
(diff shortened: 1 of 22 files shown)


veluca93:
### Performance Summary (Commit `71b5ef7`)

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_851.html)

jxl/src/render/buffer_splitter.rs
```diff
@@ -3,7 +3,97 @@
...
+        }
+    }
+
+    #[allow(unsafe_code)]
+    pub fn borrow_rect(&self, rect: Rect) -> OutputChannelRef<'a, '_> {
+        let mut rects = loop {
+            if let Some(b) = self.borrowed_rects.try_borrow_mut() {
+                break b;
+            }
+        };
```
tirr-c: Would it be better if we added spin loop hint here? (And maybe we need a blocking borrow method in `AtomicRefCell`...)

tirr-c (approved):

jxl/src/render/buffer_splitter.rs
veluca93: Valid point - I added `spin_borrow(_mut)` to the AtomicRefCell interface (with the optimization hint) and started using it.

tirr-c (approved):

PR #852, opened 2026-08-02
Prepare Modular and VarDCT decoding for multithreading.

veluca93:
Includes #851

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -11,7 +11,7 @@ use super::{
     block_context_map::BlockContextMap,
     coeff_order::decode_coeff_orders,
     color_correlation_map::ColorCorrelationParams,
-    group::{VarDctBuffers, decode_vardct_group},
+    group::decode_vardct_group,
     modular::{FullModularImage, ModularStreamId, Tree, decode_hf_metadata, decode_vardct_lf},
     quant_weights::DequantMatrices,
     quantizer::{LfQuantFactors, QuantizerParams},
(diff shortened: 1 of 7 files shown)


veluca93:
### Performance Summary (Commit `f8e5187`)

| desktop | Single | 85.53 | 85.53 | -0.16% ± 0.36% |
| desktop | Multi | 85.80 | 85.45 | -0.38% ± 0.30% |
| framework-desktop | Single | 92.02 | 101.53 | +1.57% ± 0.20% |
| framework-desktop | Multi | 92.24 | 101.37 | +1.45% ± 0.17% |
| pixel7a | Single (Fast) | 40.07 | 36.75 | -1.44% ± 0.51% |
| pixel7a | Single (Mid) | 28.23 | 26.98 | -0.92% ± 0.41% |
| pixel7a | Multi | 38.27 | 36.39 | -0.99% ± 0.60% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_852.html)

jxl/src/frame/modular/mod.rs
```diff
@@ -600,37 +607,54 @@ impl FullModularImage {
...
-            self.delayed_ready_sections.insert((1, grid));
+            loop {
+                if let Some(mut d) = self.delayed_ready_sections.try_borrow_mut() {
+                    d.insert((1, grid));
+                    break;
+                }
+            }
         } else {
-            self.mark_section_ready(section_id, grid);
+            self.mark_section_ready(section_id, grid, &mut ready_steps);
```
tirr-c, line 621: I'm curious why output vector is passed as an argument. Are we going to do multiple `mark_section_ready` calls in the future?

jxl/src/frame/decode.rs
```diff
@@ -524,15 +525,11 @@ impl Frame {
...
-                let xs = GROUP_DIM * GROUP_DIM;
-                let ys = self.header.num_groups();
-                Some((
-                    Image::new((xs, ys))?,
-                    Image::new((xs, ys))?,
-                    Image::new((xs, ys))?,
-                ))
+                (0..self.header.num_groups())
+                    .map(|_| AtomicRefCell::new(vec![0; GROUP_DIM * GROUP_DIM * 3]))
+                    .collect()
```
tirr-c: It seems that the buffers might be quite large, as this scales with the image dimension...

jxl/src/frame/modular/mod.rs
veluca93: This is also used to insert directly into self.ready_transform_steps - I agree that here it doesn't necessarily make sense, but it looked better to me this way.

jxl/src/frame/decode.rs
veluca93: Good catch!

tirr-c (approved):

PR #856, opened 2026-08-03
Grow section buffers from available input, not TOC-declared size

hjanuschka:
blink_jxl_decoder_fuzzer found a 9-byte truncated codestream (12x8 image) whose TOC declares a ~743MB section. `fill_sections` allocated each section buffer at its full TOC-declared size as soon as 1 byte of input was available, aborting with `memory allocation of 743244801 bytes failed` on the 32-bit ASAN job. Reproducible with plain `JxlDecoderInner::process` on the 9-byte input, no Chromium involved.

Grow section buffers only as far as the readable input instead, and cap the pre-section buffer move at the buffered byte count. Allocation failures now surface as `Error::OutOfMemory` via `try_reserve`. The regression test decodes the reproducer under a 128MB-capped allocator; peak RSS on the input drops ~712MB -> ~1.4MB.

https://issues.chromium.org/issues/526010666

### DIFF
--- a/jxl/src/api/inner/codestream_parser/frame_info.rs
+++ b/jxl/src/api/inner/codestream_parser/frame_info.rs
@@ -271,13 +271,17 @@ impl FrameInfo {
             self.sections = sections.into_iter().collect();
 
             // Move data from the pre-section buffer into the sections.
+            // Only allocate as much of each section as the pre-section buffer
+            // can fill; the TOC-declared section length is untrusted and may
+            // be much larger than the actual input.
             for buf in self.sections.iter_mut() {
                 if cbuf.is_empty() {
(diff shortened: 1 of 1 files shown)


jxl/tests/truncated_huge_toc.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) the JPEG XL Project Authors. All rights reserved.
```
veluca93: I don't think this test is worth it, let's remove it.

hjanuschka: Done, removed the test and its testdata file.

veluca93 (approved):

PR #858, opened 2026-08-04
Parse Mul clamp without extra channels

hjanuschka:
`Mul` blending serializes the `clamp` bit even when there are no extra channels. Skipping it shifted subsequent frame headers and made valid animations fail with `Section is too short`.

Always parse the bit for `Mul`. Conformance coverage: https://github.com/libjxl/conformance/pull/48

### DIFF
--- a/jxl/src/headers/frame_header.rs
+++ b/jxl/src/headers/frame_header.rs
@@ -126,8 +126,9 @@ pub struct BlendingInfo {
     pub alpha_channel: u32,
 
     #[default(false)]
-    #[condition(nonserialized.num_extra_channels > 0 &&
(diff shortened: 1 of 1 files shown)


jxl/src/headers/frame_header.rs
```diff
@@ -821,6 +822,25 @@ mod test_frame_header {
         assert!(!frame_header.restoration_filter.gab);
     }
 
+    #[test]
```
veluca93: I'm not sure I see the point of this test, it would make more sense to add an image with Mul blend mode to the reference corpus :-)
... Especially because as far as I can tell we decode that image wrong.

veluca93 (approved):

PR #859, opened 2026-08-04
Fix animated PNG output

hjanuschka:
`jxl_cli` created a new `png::StreamWriter` for every animation frame. The PNG crate then prefixed `IDAT` payloads with APNG sequence numbers, producing invalid files that viewers rejected with `IDAT: unknown compression method`.

Write complete frames through `Writer::write_image_data` and finish the writer explicitly. The roundtrip test covers both 8- and 16-bit APNG output.

### DIFF
--- a/jxl_cli/src/enc/png.rs
+++ b/jxl_cli/src/enc/png.rs
@@ -167,31 +167,52 @@ pub fn to_png<Writer: Write>(
     }
     let mut writer = encoder.write_header()?;
 
-    if eight_bits {
-        for frame in &image_data.frames {
-            if animated {
-                let (delay_num, delay_den) = calculate_apng_delay(frame.duration)?;
-                writer.set_frame_delay(delay_num, delay_den)?;
+    if !animated {
+        if eight_bits {
+            let mut stream = writer.stream_writer()?;
+            for frame in &image_data.frames {
+                let chan = if let Some(p) = partial_render {
+                    &frame.partial_renders[p][0]
+                } else {
+                    &frame.channels[0]
+                };
+                for y in 0..height {
+                    stream.write_all(chan.row(y))?;
+                }
             }
-            let mut ww = writer.stream_writer()?;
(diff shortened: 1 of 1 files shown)


tirr-c:
Ah yeah, streaming encoder with animation is broken... https://github.com/image-rs/image-png/issues/411

jxl_cli/src/enc/png.rs
```diff
@@ -167,51 +167,107 @@ pub fn to_png<Writer: Write>(
...
-            let mut ww = writer.stream_writer()?;
-            let chan = if let Some(p) = partial_render {
-                &frame.partial_renders[p][0]
+        writer.write_image_data(&buffer)?;
+    }
+    writer.finish()?;
+    Ok(())
+}
+
+#[cfg(test)]
```
veluca93: I'm not sure this is worth testing.

jxl_cli/src/enc/png.rs
```diff
@@ -167,51 +167,107 @@ pub fn to_png<Writer: Write>(
...
-    if eight_bits {
-        for frame in &image_data.frames {
-            if animated {
-                let (delay_num, delay_den) = calculate_apng_delay(frame.duration)?;
-                writer.set_frame_delay(delay_num, delay_den)?;
-            }
-            let mut ww = writer.stream_writer()?;
-            let chan = if let Some(p) = partial_render {
-                &frame.partial_renders[p][0]
+    let row_size = width * num_channels * if eight_bits { 1 } else { 2 };
```
veluca93: IIRC the streaming encoder was meaningfully faster / lower memory, can we keep the stream writer for non-animated PNGs?

hjanuschka:
@tirr-c https://github.com/image-rs/image-png/pull/708 fixes it, maybe a +1 there could help, i dont know how responsive the repo is. ideally if they land it, we can abandon this PR (if it's released and rolled into jxl-rs)!

PR #863, opened 2026-08-05
Don't force eager rendering when partial rendering is disallowed

hjanuschka:
blink_jxl_decoder_fuzzer found a 94-byte input that aborts with `assertion failed: self.allow_rendering_before_last_pass()` when a truncated prefix is flushed. `decode_and_render_hf_groups` computes `should_render_non_final = allow_rendering_before_last_pass() && do_flush`, but the `RenderStep::Decode` arm passed the raw `do_flush` as `force_render`, so flushing an incomplete group forced an eager VarDCT render even for frames that must not be rendered before the last pass -- here a Modular frame with a palette transform, so `can_do_partial_render()` is false. Reproducible with `JxlDecoderInner` in `JxlProgressiveMode::Pass` plus `flush_pixels`, no Chromium involved; a 47-byte prefix is enough, and it reproduces both on v0.5.1 (the version rolled into Chromium) and on current main.

Pass `should_render_non_final` instead, which already combines both conditions. The regression test flushes every truncation of the reproducer.

https://issues.chromium.org/issues/541318910


jxl/src/tests/api.rs
```diff
@@ -765,6 +765,59 @@ fn test_fuzzer_smallbuffer_overflow() {
...
 
+/// Regression test for ClusterFuzz issue 5920454820790272 (crbug.com/541318910).
+///
+/// Flushing a truncated Modular frame whose transforms do not support partial
+/// rendering used to force an eager VarDCT render of an incomplete group.
+#[test]
+fn clusterfuzz_541318910() {
+    let file = include_bytes!("../../tests/testdata/clusterfuzz_541318910.jxl");
+
+    // Emulate a progressive client that flushes whatever it has whenever the
```
veluca93: Can't we get the same behaviour with the regular decode helper?

jxl/src/tests/api.rs
```diff
@@ -765,6 +765,59 @@ fn test_fuzzer_smallbuffer_overflow() {
...
+    // decoder runs out of input, for every possible truncation of the file.
+    for len in 1..=file.len() {
+        let options = JxlDecoderOptions {
+            progressive_mode: JxlProgressiveMode::Pass,
```
veluca93: This reminds me: this enum is actually completely unused and we should remove it - will open a cleanup PR.

veluca93 (approved):

PR #866, opened 2026-08-05
Continue refilling after incremental parser progress

### LINKED ISSUE #865: Failure to decode
Renamed to `.jpg` for upload, bisected to #828 by Mario on Discord.
```
Error: Source file truncated

Location:
    jxl_cli\src\dec\mod.rs:331:32
```
Image

hjanuschka:
Large TOCs can fill the local parser buffer and then make progress before
requesting more input. Continue the refill loop when parsing consumed buffered
data, instead of returning `NeedsMoreInput` while more source data is available.

Fixes #865


jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -378,3 +379,44 @@ impl CodestreamParser {
...
+mod tests {
+    use crate::{
+        api::inner::box_parser::{BoxParser, CodestreamInput},
+        error::Error,
+    };
+
+    use super::CodestreamParser;
+
+    #[test]
+    fn refill_after_incremental_parse_progress() {
```
veluca93: I think we should be able to reproduce this just by adding the file to testdata, no?

veluca93: (or some similar test file)

veluca93 (approved):

PR #862, opened 2026-08-05
Use shuttle for concurrency testing.

veluca93:

PR #868, opened 2026-08-06
Fill vertically subsampled border rows with floor semantics

hjanuschka:
Shuttle (#862) found one corrupted row at the bottom image border in `test_compare_parallel_oneshot_shuttle_multiple_lf_420`; pre-existing on main, just schedule-dependent enough to never hit with real threads.

The low-memory pipeline fills border rows of vertically subsampled channels with ceil semantics (`stage_vy - y0 < -(border << dy)`), while `BufferFiller::new` and the whole x direction use floor. For a rect starting at unaligned y0 the chroma context row above is never filled and is read stale from the per-thread `RowBuffer` ring -- usually the right row still sits there from the previous rect, but a racy rect split (interleaved `mark_ready` decrements of neighbouring groups) lets the top rect clobber the slot the bottom rect needs.

Fill with floor semantics instead (Save stages keep ceil, they must only write rows owned by the rect) and extend the vy range so the extra parity row is reachable. With this + #867 the full shuttle suite passes, also with SHUTTLE_ITERATIONS=200.

### DIFF
--- a/jxl/src/render/low_memory_pipeline/render_group.rs
+++ b/jxl/src/render/low_memory_pipeline/render_group.rs
@@ -317,7 +317,18 @@ impl LowMemoryRenderPipeline {
         // to an actual row of the current processing stage; actual processing happens
         // when vy % (1<<vshift) == 0.
 
-        let vy0 = y0.saturating_sub(num_extra_rows);
+        // Rects are not guaranteed to start at a vertical position that is aligned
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

veluca93:
### Performance Summary (Commit `39f136d`)

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_868.html)

PR #869, opened 2026-08-06
Take read locks on palette prediction context grids

hjanuschka:
More shuttle fuzzing on top of #862, this time with the PCT scheduler: the palette transform write-locks all grids it touches, including neighbouring grids (previous row for full-row predictors, left/top/top-left otherwise) it only reads as prediction context. An `Output` step reading a previous-row grid concurrently makes the `try_write`/`try_read` calls panic with `WouldBlock` (`test_compare_parallel_progressive_shuttle_issue648_palette0`).

Split the palette helpers' buffer slices into written grids and read-only context grids, and take read locks for the latter; ordering with the producers of the context grids is already guaranteed by the dependency graph. With this (plus #867 and #868) the shuttle suite is clean under both the random and PCT schedulers.

### DIFF
--- a/jxl/src/frame/modular/transforms/palette.rs
+++ b/jxl/src/frame/modular/transforms/palette.rs
@@ -255,10 +255,10 @@ pub fn do_palette_step_general(
 
 #[allow(clippy::too_many_arguments)]
 fn get_prediction_data(
-    buf: &mut [&mut ModularChannel],
+    cur_row: &[&mut ModularChannel],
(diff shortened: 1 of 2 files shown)


veluca93 (approved):

veluca93:
### Performance Summary (Commit `e6a440d`)

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_869.html)

PR #872, opened 2026-08-10
Count partial groups when limiting render parallelism

tnikkel:
max_threads in decode_and_render_hf_groups is a product of two truncating
divisions, so any image with a side smaller than group_dim gets a zero factor,
a cap of 1, and a serial render. A 1476x156 image has six groups but only ever
gets one thread.

Use div_ceil so partial groups count. On 8 threads that is 1.5x to 4.4x faster
for images with a short side (1476x156 goes from 17.8ms to 4.0ms). Images with
both sides larger than group_dim are unchanged.

### DIFF
--- a/jxl/src/frame/render.rs
+++ b/jxl/src/frame/render.rs
@@ -329,8 +329,20 @@ impl Frame {
             }
         }
 
+        let largest_group_area = self.header.group_dim().min(self.header.size().0)
+            * self.header.group_dim().min(self.header.size().1);
+
+        const THREAD_COUNT_DENOMINATOR: usize = 16;
+
+        let mut groups_of_work = 0;
+
         if self.header.encoding == Encoding::VarDCT {
             for (group, _) in groups.iter() {
+                let sz = self.header.group_rect(*group).size;
+                let area = sz.0 * sz.1;
+                if area >= largest_group_area {
+                    groups_of_work += THREAD_COUNT_DENOMINATOR;
+                }
                 if self.group_status.channel_status[*group][0] == DataStatus::Final {
                     pipeline_mut!(self, p, p.mark_group_to_rerender(*group));
                 }
@@ -340,6 +352,13 @@ impl Frame {
         // STEP 3: Make sure modular_global is ready to run, and prepare the list of
         // all the decoding/rendering steps that we want to run.
         modular_global.prepare_render(&self.header, |g, c, is_final| {
+            let sz = self.header.group_rect(g).size;
+            let area = sz.0 * sz.1;
+            if area >= largest_group_area {
+                groups_of_work += THREAD_COUNT_DENOMINATOR / 2;
+            } else if area * 2 >= largest_group_area {
+                groups_of_work += THREAD_COUNT_DENOMINATOR / 4;
+            }
             self.group_status.update_status(
                 g,
                 c,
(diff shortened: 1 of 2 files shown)


veluca93:
This approach cause the limit to be a no-op in most cases.

Maybe a better idea would be to limit to ceil(total pixels / pixels per group)

tnikkel:
You're right, with ceil per dimension the limit never fires — `render_steps.len()` is exactly the group count everywhere I measured.

That said, removing it still beats the area version in my benchmarks. 8 thread rayon runner, ms, main / ceil / area:

```
1476x156   20.49   4.53   6.69
738x78      5.99   2.34   3.72
150x812    10.76   3.61   4.13
320x267     4.75   4.86   4.54
1280x720   12.73  12.44  12.43
```

The two thin cases are entirely partial groups and area caps below the group count there. The small squarish case the limit should help is a wash.

Can you share more about the workload and tests behind the existing limit?

tnikkel:
Switched to ceil(total pixels / pixels per group).

For the record here, since it was in chat: you found that on android, on a small image (green queen vardct), not having the limit makes things much worse.

Makes sense, phones have a couple of fast cores and several slower ones, so one group per thread lets a slow core set the completion time. My machine is 8 uniform cores.

Area isn't quite as fast on a laptop as ceil, but it's a good improvement, and we need to have good performance on android too.

tnikkel:
The shuttle failure looks like an existing race that the old limit was hiding, if I remove the limit on main then I can reproduce the failure. But still something that will need to be looked into.

jxl/src/frame/render.rs
```diff
@@ -436,10 +436,13 @@ impl Frame {
...
-        let max_threads = (self.header.size().0 / self.header.group_dim())
-            * (self.header.size().1 / self.header.group_dim())
-            + 1;
+        // Avoid significantly more than one thread per group's worth of work.
+        // Measured in whole groups of area rather than per dimension: truncating
+        // each dimension separately makes the product zero whenever a dimension
+        // is under one group, which pins max_threads to 1 and renders serially.
+        let group_dim = self.header.group_dim();
+        let max_threads =
+            (self.header.size().0 * self.header.size().1).div_ceil(group_dim * group_dim) + 1;
```
veluca93: We should probably handle overflows explicitly here (IIRC header.size() is a (u32, u32))

veluca93 (approved):

veluca93:
### Performance Summary (Commit `bb6fa45`)

| desktop | Single | 85.05 | 85.30 | +0.07% ± 0.26% |
| desktop | Multi | 416.77 | 403.29 | +0.97% ± 0.41% |
| framework-desktop | Single | 100.22 | 95.07 | -0.54% ± 0.34% |
| framework-desktop | Multi | 558.59 | 538.61 | +0.05% ± 0.87% |
| pixel7a | Single (Fast) | 37.95 | 38.30 | -0.02% ± 0.50% |
| pixel7a | Single (Mid) | 26.81 | 28.02 | +0.92% ± 0.55% |
| pixel7a | Multi | 87.87 | 89.36 | -1.67% ± 2.90% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_872.html)

veluca93:
As I feared, this shows significant performance regressions on Android (specifically an almost 2x slowdown on green_queen_vardct_e3.jxl).

I am not sure what the best way to approach this would be.

veluca93 (approved):

PR #878, opened 2026-08-16
Use the default 2x upsampling kernel for smooth unsqueeze

hjanuschka:
Progressive previews of squeezed modular images currently use the tendency-derived linear filter; this switches the smooth-unsqueeze path to the non-separable default 2x upsampling kernel with min/max clamping to the local 5x5 neighbourhood, same as the regular upsampling stage. The h-only/v-only paths use a 5-row window with 1D kernels obtained by averaging the 2D kernels. Work by @Galaxy4594 and @jonnyawsom3 (first commit is theirs, from Galaxy4594/upsampling).

Sharing the kernel code between upsample.rs and squeeze.rs cost +61KB of stripped release binary; the second commit gets that down to +38KB: the kernel tables are const-evaluated instead of rebuilt inside every monomorphized SIMD instantiation, the min/max scratch buffers live in the reusable per-thread scratch space instead of four fresh Vecs per group, and kernel_conv/DEFAULT_KERN_2 stay crate-private. The remaining delta is the inherent cost of the dense 25-tap convolutions + clamping across the 12 SIMD instantiations. Added a test checking the const tables against the original kernel construction.

The third commit fixes the arm/none CI failure, which turned out to be a real bug rather than a stale threshold: the input average channel has ceil(xs/2) columns but in_xs was computed with floor, so for odd output widths the last output column was never written and stayed zero, and deeper unsqueeze levels smeared it into a dark gradient at the right image edge (grayscale_public_university is 2880 wide, which is 45 -- odd -- after three halvings). Scalar lanes hit this for every odd width while wider lanes usually mask it via the partial-chunk store, hence only the none job failing. With the fix the scalar/SIMD checkpoint spread is back to FP noise and all incremental checkpoints measure at or below their stored expectations on scalar/sse42/avx/avx512; the changed files all improved (e.g. bicycles -24%, efb -33%, grayscale_public_university -14% MSE at the first checkpoint), so the stored values are left as now-loose upper bounds.

Partial decodes of a --responsive=1 lossless image are bit-identical between the original branch and the size-reduction commit at various truncation points, and differ from main, i.e. the new path is actually exercised.

### DIFF
--- a/jxl/src/features/upsampling.rs
+++ b/jxl/src/features/upsampling.rs
@@ -8,6 +8,11 @@ use jxl_simd::{F32SimdVec, SimdDescriptor};
 
 /// Precomputes flattened 5x5 kernels for 2x upsampling from CustomTransformData weights (15
 /// values). Output layout: kernel[oy * 2 + ox] -> [f32; 25] for oy, ox in 0..2
+///
+/// This is a const fn so that the kernel tables derived from the default weights can be
+/// const-evaluated: keeping the kernel values compile-time constants lets the compiler fold the
+/// splat loops in the (monomorphized-per-SIMD-target) smooth-unsqueeze functions into vector
+/// constants, which is considerably smaller than materializing the loops at runtime.
 const fn compute_5x5_kernels_2x(weights: &[f32; 15]) -> [[f32; 25]; 4] {
     // kernel[oy * 2 + ox], each flattened as [row * 5 + col].
     let mut kernel = [[0.0f32; 25]; 4];
@@ -134,7 +139,7 @@ pub fn compute_minmax<D: jxl_simd::SimdDescriptor>(
 #[allow(clippy::too_many_arguments)]
 pub fn kernel_conv<D: SimdDescriptor>(
     d: D,
-    kv: &[D::F32Vec; 25],
+    kv: &[f32; 25],
     r0: &[f32],
     r1: &[f32],
     r2: &[f32],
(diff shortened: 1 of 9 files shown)


jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -679,187 +681,29 @@ use super::super::{ModularBufferInfo, ModularGridKind};
...
+    col_max: Vec<f32>,
+    mins: Vec<f32>,
+    maxs: Vec<f32>,
 }
 
-fn init_buffers(buf: &mut [Vec<f32>; 5], ibuf: &mut Vec<i32>, len: usize) {
-    for b in buf {
-        b.resize(len, 0.0);
+impl SmoothUnsqueezeScratch {
+    fn prepare(&mut self, row_len: usize, minmax_len: usize) {
```
veluca93: We call equivalent functions `init()` elsewhere.

jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -1014,17 +858,35 @@ fn smooth_2d_unsqueeze_simd_impl<D: SimdDescriptor>(
     frame_header: &FrameHeader,
     rect: Rect,
     output: &mut Image<i32>,
-    (buffer, ibuf): &mut ([Vec<f32>; 5], Vec<i32>),
+    scratch: &mut SmoothUnsqueezeScratch,
 ) {
-    let (in_xs, in_ys) = (rect.size.0 / 2, rect.size.1 / 2);
+    // The input channel has ceil(xs / 2) columns; using floor here would leave
```
veluca93: Let's remove this comment

jxl/src/features/upsampling.rs
veluca93, line 1: I don't think `util` is a good place for this, as there are quite a few format-specific details here.

jxl/src/features/upsampling.rs
```diff
@@ -0,0 +1,230 @@
...
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+use crate::headers::DEFAULT_KERN_2;
+use jxl_simd::{F32SimdVec, SimdDescriptor};
+
+/// Precomputes flattened 5x5 kernels for 2x upsampling from CustomTransformData weights (15
+/// values). Output layout: kernel[oy * 2 + ox] -> [f32; 25] for oy, ox in 0..2
+const fn compute_5x5_kernels_2x(weights: &[f32; 15]) -> [[f32; 25]; 4] {
```
veluca93, line 16: What's the benefit of those being const?

ISTM that this computation could easily be done at runtime (maybe as part of scratch buffer initialization), and that would easily save ~1kb of file size

jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -1228,17 +1128,33 @@ fn smooth_v_unsqueeze_simd_impl<D: SimdDescriptor>(
...
+        rows: buffer,
+        irow: ibuf,
+        col_min,
+        col_max,
+        mins,
+        maxs,
+    } = scratch;
+
+    // 1D vertical 2x kernels, derived by horizontally averaging the 2D 2x kernels
+    let mut kernel_vecs = [[D::F32Vec::splat(d, 0.0); 25]; 2];
```
veluca93: I'm not fully convinced of the benefit of doing the splat() in advance, have you considered just passing the float array into kernel_conv and letting the optimizer deal with it?

jxl/src/features/upsampling.rs
hjanuschka: Moved it to features/upsampling.rs, next to the other format-specific modules.

jxl/src/features/upsampling.rs
hjanuschka: I implemented your suggestion (computing the kernels once in the scratch buffer's Default) and measured: it goes the other way, +18KB. With the tables const, LLVM folds the splat loops into vector constants; with runtime values it materializes the loops and spills in every monomorphized SIMD instantiation (stripped x86-64 release jxl_cli: 4479280 bytes const vs 4497264 runtime; smooth-unsqueeze symbols 90.4KB vs 108.5KB). Kept the const version and added a comment on compute_5x5_kernels_2x explaining this. Can switch to the runtime variant if you still prefer it.

jxl/src/frame/modular/transforms/squeeze.rs
hjanuschka: Done, and it's a clear improvement: kernel_conv now takes &[f32; 25] and splats internally, all the pre-splat loops are gone (squeeze and the upsample stages). Squeeze codegen is bit-identical (const kernels still fold into vector constants), the avx512 upsample loop got tighter (963 -> 757 insns, broadcasts fold into FMA memory operands), avx2 swaps stack spill-reloads for broadcast-loads with the same FMA count. -4.8KB binary, outputs bit-identical.

veluca93:
#885 upscales better :-)

PR #889, opened 2026-08-21
When doing LF smoothing, compute smoothing factor within a subsampling block, instead of a pixel

tirr-c:


jxl/src/frame/adaptive_lf_smoothing.rs
tirr-c, line 36: Oops, I need to consider group borders here...

PR #891, opened 2026-08-21
Add CMYK interleaved pixel format

hjanuschka:
Embedders currently receive K as a separate extra channel plane and have to re-interleave it into the fourth byte before handing pixels to their CMS (skcms consumes C,M,Y,K interleaved with K in the alpha slot, qcms takes interleaved CMYK too); both Firefox and Chromium grew per-pixel merge loops for this.

Add `JxlColorType::Cmyk`: four interleaved channels where K comes from the Black extra channel, keeping the color channel convention (max value = no ink). That matches the "inverted CMYK" convention CMYK JPEG decoders produce, so the buffer drops straight into existing CMYK consumer paths. Requesting it errors with `NotCmyk` unless the image is a color image with a Black extra channel. There is no CMYKA; if the image also carries an alpha extra channel it stays available as a separate extra channel plane, untouched by this format.

Tests: interleaved output matches the Rgb + Black-plane decode bit-exactly on `cmyk_layers.jxl` for both pipelines, plus the error case.

### DIFF
--- a/jxl/src/api/data_types.rs
+++ b/jxl/src/api/data_types.rs
@@ -14,6 +14,10 @@ pub enum JxlColorType {
     Rgba,
     Bgr,
     Bgra,
+    /// Four interleaved channels: C, M, Y and K, where K is taken from the
+    /// image's Black extra channel. Only valid for color images with a Black
+    /// extra channel.
+    Cmyk,
 }
 
 impl JxlColorType {
@@ -23,6 +27,7 @@ impl JxlColorType {
             Self::GrayscaleAlpha => true,
             Self::Rgb | Self::Bgr => false,
             Self::Rgba | Self::Bgra => true,
+            Self::Cmyk => false,
         }
     }
     pub fn samples_per_pixel(&self) -> usize {
@@ -30,7 +35,7 @@ impl JxlColorType {
             Self::Grayscale => 1,
             Self::GrayscaleAlpha => 2,
             Self::Rgb | Self::Bgr => 3,
-            Self::Rgba | Self::Bgra => 4,
+            Self::Rgba | Self::Bgra | Self::Cmyk => 4,
         }
     }
     pub fn is_grayscale(&self) -> bool {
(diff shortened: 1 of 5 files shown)


jxl/src/api/data_types.rs
```diff
@@ -14,6 +14,15 @@ pub enum JxlColorType {
     Rgba,
     Bgr,
     Bgra,
+    /// Four interleaved channels: C, M, Y and K, where K is taken from the
```
veluca93, line 17: This comment seems somewhat unnecessary.

jxl/src/frame/render.rs
```diff
@@ -848,13 +848,41 @@ impl Frame {
...
+            // fourth channel. This requires an actual CMYK image: a color
+            // image with a Black extra channel and a CMYK ICC profile.
+            let black_in_color = if pixel_format.color_type == JxlColorType::Cmyk {
+                let black_channel = decoder_state
+                    .file_header
+                    .image_metadata
+                    .extra_channel_info
+                    .iter()
+                    .position(|info| info.ec_type == ExtraChannel::Black);
+                // ICC.1:2022 section 7.2: bytes 16..20 of the profile header
```
veluca93: I don't think we should check this way - I would rather not check at all and not return an error then go look into the bytes of the ICC.

jxl/src/tests/api.rs
```diff
@@ -669,6 +669,144 @@ fn test_output_format_f16_matches_f32() {
     }
 }
 
+/// Helper to decode with u8 output and buffers for the color channels plus
```
veluca93: This sounds like something that can, and should, be folded in the standard `decode` helper.

jxl/src/api/data_types.rs
```diff
@@ -39,13 +44,17 @@ impl JxlColorType {
...
         }
     }
     pub fn add_alpha(&self) -> Self {
         match self {
             Self::Grayscale | Self::GrayscaleAlpha => Self::GrayscaleAlpha,
             Self::Rgb | Self::Rgba => Self::Rgba,
             Self::Bgr | Self::Bgra => Self::Bgra,
+            // There is no CMYKA; alpha is only available as a separate
+            // extra channel plane.
+            Self::Cmyk => Self::Cmyk,
```
veluca93: I think we should not silently not add the alpha channel here, but instead either panic or return a Option<Self>.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `e819568`)

| desktop | Single | 88.20 | 88.46 | +0.57% ± 0.86% |
| desktop | Multi | 415.59 | 412.30 | -0.07% ± 0.45% |
| framework-desktop | Single | 101.73 | 96.63 | -0.92% ± 0.40% |
| framework-desktop | Multi | 572.52 | 564.80 | -0.33% ± 0.80% |
| pixel7a | Single (Fast) | 39.77 | 39.93 | +4.75% ± 0.43% |
| pixel7a | Single (Mid) | 28.63 | 29.30 | +0.54% ± 0.47% |
| pixel7a | Multi | 93.44 | 92.01 | -1.24% ± 2.64% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_891.html)

jxl/src/tests/api.rs
```diff
@@ -714,19 +782,32 @@ fn decode_with_format<T: crate::image::ImageDataType>(
...
-            })
-            .into_raw(),
-    )];
+    let mut images = Vec::new();
+    if pixel_format.color_data_format.is_some() {
+        images.push(Image::<T>::new((width * num_samples, height)).unwrap());
+    }
+    for ec_format in &pixel_format.extra_channel_format {
+        if ec_format.is_some() {
+            images.push(Image::<T>::new((width, height)).unwrap());
```
sboukortt: Now that the function returns a `Result`, might this not as well be:

```suggestion
        images.push(Image::<T>::new((width * num_samples, height))?);
    }
    for ec_format in &pixel_format.extra_channel_format {
        if ec_format.is_some() {
            images.push(Image::<T>::new((width, height))?);
```

veluca93 (approved):

PR #892, opened 2026-08-21
Reject pixel format changes after the first frame header is decoded

hjanuschka:
Reworked per review: the pixel format is now frozen as soon as the first frame header is decoded. `set_pixel_format` returns `Error::PixelFormatChangedAfterFirstFrame` after that point (setting an identical format stays allowed), so the format can be changed at most right after basic info becomes available. No render-time checks or pipeline-format snapshots are needed anymore.

One subtlety: this is reachable through the type-safe API as well. Invisible frames (e.g. reference frames) are decoded inside `process()` while the decoder stays in the `WithImageInfo` state, so the typestate `set_pixel_format` now returns `Result` and propagates the error rather than hiding it.

Chrome-side reproduction (fixed in https://chromium-review.googlesource.com/c/chromium/src/+/8278890, so this is defense in depth): the frame scanner (`scan_frames_only`) processed all buffered input in a loop and only configured its pixel format (no extra channel buffers) after the loop drained. When the first network chunk contained enough data, the decoder had already parsed ahead into `patches.jxl`'s reference frame and built its pipeline with the default pixel format (color + all extra channels), so any later render aborted the renderer (`index out of bounds: the len is 5 but the index is 5` in `check_buffer_sizes`). The Chromium fix configures the format right when `process()` first reports basic info.

### DIFF
--- a/jxl/src/tests/api.rs
+++ b/jxl/src/tests/api.rs
@@ -93,7 +93,7 @@ fn test_set_pixel_format() {
         color_data_format: Some(JxlDataFormat::U8 { bit_depth: 8 }),
         extra_channel_format: vec![],
     };
-    decoder.set_pixel_format(new_format.clone());
+    decoder.set_pixel_format(new_format.clone()).unwrap();
     assert_eq!(decoder.current_pixel_format(), &new_format);
 }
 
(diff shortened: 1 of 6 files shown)


veluca93:
This doesn't feel right - Chrome should not do this in the first place. I will try to fix the code so that it doesn't happen.

As for the panic, we should limit ourselves to returning an API usage error instead.

What's the simplest way to reproduce this Chrome side?

veluca93 (commented):
Why don't we just return an error when calling set_pixel_format if we have a frame?
Note that this is already disallowed by the typestate API.

veluca93 (commented):
I think we should just disallow changing the pixel format as soon as the first frame header gets decoded, i.e. you should change pixel format at most just after you have basic info.

jxl/src/api/inner/codestream_parser/mod.rs
```diff
@@ -102,6 +102,10 @@ pub(super) struct CodestreamParser {
     pub output_color_profile: Option<JxlColorProfile>,
     pub pixel_format: Option<JxlPixelFormat>,
 
+    /// True once the first frame header has been decoded, after which the
+    /// pixel format must not change anymore.
+    pub seen_frame_header: bool,
```
veluca93: Isn't this information already available?

jxl/src/tests/api.rs
```diff
@@ -1294,3 +1296,84 @@ fn decode_test_strategic_solid_blue_grid_boundary() {
         }
     }
 }
+
+/// The pixel format can only be changed before the first frame header is
```
veluca93: I don't think we need a test for this.

veluca93 (approved):

veluca93 (approved):

veluca93:
### Performance Summary (Commit `7a98bcd`)

| desktop | Single | 91.45 | 91.44 | +0.09% ± 0.32% |
| desktop | Multi | 413.03 | 417.22 | +0.29% ± 0.44% |
| framework-desktop | Single | 101.61 | 97.31 | -0.57% ± 0.20% |
| framework-desktop | Multi | 579.40 | 575.81 | -0.27% ± 0.92% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_892.html)

PR #902, opened 2026-08-23
Reject invalid blending reference frames

hjanuschka:
Reject blending against pre-color-transform or undersized reference frames before constructing the render pipeline. This prevents out-of-bounds row access for malformed codestreams.

ClusterFuzz: https://clusterfuzz.com/testcase?key=5937058744926208
Chromium issue: https://issues.chromium.org/issues/550831596

### DIFF
--- a/jxl/src/render/stages/blending.rs
+++ b/jxl/src/render/stages/blending.rs
@@ -3,7 +3,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-use crate::error::Result;
+use crate::error::{Error, Result};
 use crate::features::blending::perform_blending;
 use crate::features::patches::{PatchBlendMode, PatchBlending};
 use crate::frame::ReferenceFrame;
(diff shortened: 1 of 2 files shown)


jxl/src/render/stages/blending.rs
```diff
@@ -61,18 +61,48 @@ impl BlendingStage {
...
+        let image_size = (
+            file_header.size.xsize() as usize,
+            file_header.size.ysize() as usize,
+        );
+        let color_source = frame_header.blending_info.source as usize;
+        if reference_frames[color_source]
+            .as_ref()
+            .is_some_and(|frame| frame.saved_before_color_transform)
+        {
+            return Err(Error::BlendingPreColorTransform(color_source));
```
veluca93: Shouldn't we check this for *all* sources, not just color?

jxl/src/render/stages/blending.rs
```diff
@@ -61,18 +61,48 @@ impl BlendingStage {
...
+        for info in
+            std::iter::once(&frame_header.blending_info).chain(frame_header.ec_blending_info.iter())
+        {
+            let source = info.source as usize;
+            let Some(frame) = reference_frames[source].as_ref() else {
+                continue;
+            };
+            let size = frame.frame[0].size();
+            if size.0 < image_size.0 || size.1 < image_size.1 {
+                return Err(Error::BlendingReferenceTooSmall(
```
veluca93: Have you found a file that actually triggers this? I think it should be impossible, because if the source is post-color-transform, then size == image_size. If anything, we should assert that they are indeed equal, I believe (unless we returned BlendingPreColorTransform above)

jxl/src/render/stages/blending.rs
hjanuschka: done, now checking `saved_before_color_transform` for the color source and every extra-channel source.

jxl/src/render/stages/blending.rs
hjanuschka: no separate file. the ClusterFuzz reproducer had a pre-color-transform reference with a smaller size, so it now returns `BlendingPreColorTransform` first. changed the post-color-transform size check to `assert_eq!(..., image_size)`.

jxl/src/render/stages/blending.rs
```diff
@@ -179,17 +195,87 @@ mod test {
...
+            Some(ReferenceFrame::blank(image_size.0, image_size.1, 4, true)?),
+            None,
+        ]);
+
+        let result = BlendingStage::new(&frame_header, &file_header, reference_frames);
+        assert!(matches!(result, Err(Error::BlendingPreColorTransform(2))));
+        Ok(())
+    }
+
+    #[test]
```
veluca93: I don't think we need this test.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `9ec5273`)

| desktop | Single | 88.94 | 88.00 | -0.66% ± 0.94% |
| desktop | Multi | 411.54 | 413.52 | -0.35% ± 0.42% |
| framework-desktop | Single | 96.64 | 96.93 | -0.17% ± 0.37% |
| framework-desktop | Multi | 570.01 | 564.96 | +0.44% ± 0.96% |
| pixel7a | Single (Fast) | 40.68 | 41.92 | +0.33% ± 0.45% |
| pixel7a | Single (Mid) | 28.78 | 29.50 | -0.04% ± 0.43% |
| pixel7a | Multi | 94.55 | 92.98 | +0.32% ± 3.16% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_902.html)

PR #906, opened 2026-08-25
Extend the Zero-predictor fast path to scaled leaves

hjanuschka:
Single-leaf Zero trees with a non-trivial multiplier or offset (common for palette-delta index channels) previously fell through to the generic flat-tree decoder, recomputing all tree properties per pixel. This routes them through the existing `NoTreeZero` fast path, applying the multiplier/offset when producing the pixel. The tight loop for `multiplier == 1, offset == 0` is unchanged.

Measured with a single-binary A/B toggle (so code layout is constant), median of 8 alternating `taskset`-pinned runs, `--release`, `--num-threads 1`:

| file | before MP/s | after MP/s | delta |
|---|---|---|---|
| palette-heavy photo (e7, palette+delta) | 12.97 | 15.51 | +19.5% |
| effort-3 lossless | 9.19 | 9.71 | +5.6% |
| 55_Cancri_e (cjxl v0.12 `-d 0`) | 6.21 | 6.30 | +1.4% |

Decoded output is bit-identical to main on all test files.

### DIFF
--- a/jxl/src/frame/modular/decode/specialized_trees.rs
+++ b/jxl/src/frame/modular/decode/specialized_trees.rs
@@ -370,6 +370,8 @@ impl<R: Reader> ModularChannelDecoder for SingleGradientOnly<R> {
 struct NoTreeZero {
     clustered_ctx: usize,
     single_value: Option<i32>,
+    multiplier: u32,
+    offset: i64,
 }
 
 impl ModularChannelDecoder for NoTreeZero {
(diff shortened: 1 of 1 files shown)


veluca93 (approved):

veluca93:
### Performance Summary (Commit `9597400`)

| framework-desktop | Single | 101.35 | 101.79 | +2.00% ± 0.30% |
| framework-desktop | Multi | 556.67 | 570.79 | +2.59% ± 0.90% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_906.html)

PR #907, opened 2026-08-25
Add a specialized decoder for gradient-only multi-property trees

hjanuschka:
Trees whose leaves are all `Gradient` and whose splits only use the local neighborhood properties 9..=13 currently fall through to the generic flat-tree decoder, which computes all 15 tree properties (plus reference properties) per pixel. This is exactly the tree shape `cjxl --faster_decoding` produces for lossless images.

This adds a `GradientMultiProp` decoder that computes only the five needed properties and carries no weighted-predictor state.

Measured with a single-binary A/B toggle (so code layout is constant), median of 8 alternating `taskset`-pinned runs, `--release`, `--num-threads 1`, on a 96 MP photo ([55_Cancri_e_Final_1_30.png](https://commons.wikimedia.org/wiki/File:55_Cancri_e_Final_1_30.png)) encoded with cjxl v0.12.0 `-d 0`:

| encode | generic MP/s | specialized MP/s | delta |
|---|---|---|---|
| `--faster_decoding=1` | 11.66 | 12.59 | +8.0% |
| `--faster_decoding=2` | 12.13 | 13.44 | +10.8% |
| effort 3 | 9.68 | 10.03 | +3.6% |
| default effort 7 | 6.27 | 6.35 | +1.4% |

Decoded output is bit-identical to main on all test files.

### DIFF
--- a/jxl/src/frame/modular/decode/specialized_trees.rs
+++ b/jxl/src/frame/modular/decode/specialized_trees.rs
@@ -337,6 +337,100 @@ impl<R: Reader> ModularChannelDecoder for GradientOnly<R> {
     }
 }
 
+/// Decoder for trees whose leaves are all `Gradient` and whose splits only use
+/// the local neighborhood properties 9..=13, which is what
+/// `cjxl --faster_decoding` produces. Skips the full property computation (and
+/// any weighted predictor state) of the generic flat-tree decoder.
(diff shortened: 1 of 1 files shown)


veluca93:
The speedup seems minor (~5% on a local benchmark on an image that is already fast enough), so closing this for now.

PR #908, opened 2026-08-28
Fix panic on grayscale non-XYB VarDCT frames

Mrmaxmeier:
Hi,

I ran into a panic while fuzzing the `decode` target. I've added a minimal reproducer as `jxl/tests/testdata/vardct_grayscale_unused_channel.jxl`:

```
thread 'tests::api::test_fuzzer_vardct_grayscale_unused_channel' panicked at jxl/src/render/internal.rs:150:15:
called `Option::unwrap()` on a `None` value
   5: <jxl::render::internal::RenderPipelineShared<..>>::group_size_for_channel
   6: <jxl::render::low_memory_pipeline::LowMemoryRenderPipeline as jxl::render::RenderPipeline>::get_buffer::<f32>
   7: <jxl::frame::Frame>::decode_and_render_varct_and_noise
   8: <jxl::frame::Frame>::decode_hf_group
```

The reproducer frame is grayscale and has no XYB or YCbCr stage, so nothing in the pipeline consumes color channels 1 and 2: They're discarded and have `ChannelInfo::ty == None`.
VarDCT still decodes these channels and asks the pipeline for a scratch buffer for each, and `group_size_for_channel` unwraps their type:

https://github.com/libjxl/jxl-rs/blob/d7ecec1e3b39c2beb3ec9fbb75ca8b3ffd0922f2/jxl/src/render/internal.rs#L144-L155

The size returned by this function doesn't depend on `ty`. It's only used to assert that caller and pipeline agree.

I've adjusted the check to only run when the channel actually has a type. This seems fine in practice: An unused channel gets a correctly sized throwaway buffer that `set_buffer_for_group` (already guarded by `channel_is_used`) drops.

Thanks!


veluca93 (approved):
Seems fine to me, do you think you could (in a follow-up :-)) modify the fuzzer to search for those cases too?

(and if you feel like it, also to test progressive decoding scenarios :-))

Mrmaxmeier:
> Seems fine to me, do you think you could (in a follow-up :-)) modify the fuzzer to search for those cases too?
> 
> (and if you feel like it, also to test progressive decoding scenarios :-))

The fuzz target already "supports" detection for this issue, libfuzzer is just not able to efficiently come up with an input that triggers this due to the jxl file format complexity. I ran into this particular thing while experimenting with format-specific fuzzing approaches. I'll report back when/if I've got something that can be nicely upstreamed or published. :slightly_smiling_face:

veluca93:
### Performance Summary (Commit `20b2cc1`)

| desktop | Single | 91.54 | 90.82 | -0.73% ± 0.32% |
| desktop | Multi | 420.52 | 420.20 | +0.01% ± 0.42% |
| framework-desktop | Single | 96.97 | 97.37 | +0.04% ± 0.19% |
| framework-desktop | Multi | 566.95 | 568.74 | +0.07% ± 0.80% |
| pixel7a | Single (Fast) | 40.58 | 41.59 | +0.09% ± 0.30% |
| pixel7a | Single (Mid) | 29.51 | 29.51 | +0.02% ± 0.46% |
| pixel7a | Multi | 92.88 | 92.40 | -3.24% ± 2.77% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_908.html)

PR #910, opened 2026-08-28
Fix panic when a transform is applied to an empty palette channel

Mrmaxmeier:
Hi,

here's another fuzzer find (`jxl/tests/testdata/modular_palette_empty_meta_channel.jxl`):

```
thread '<unnamed>' panicked at jxl/src/frame/modular/transforms/apply_local.rs:270:21:
assertion `left == right` failed
  left: (0, 1)
 right: (0, 0)
   4: <jxl::frame::modular::transforms::step::TransformStep>::local_apply
   5: jxl::frame::modular::decode::bitstream::decode_modular_subbitstream
   6: jxl::frame::modular::decode_vardct_lf
   7: <jxl::frame::Frame>::decode_lf_group
```

The group header of the input's LF group carries two nested palette transforms, the second one is applied to the palette channel produced by the first:

```
Transform { id: Palette, begin_channel: 0, num_channels: 1, num_colors: 0,    num_deltas: 0, ... }
Transform { id: Palette, begin_channel: 0, num_channels: 1, num_colors: 1796, num_deltas: 0, ... }
```

`meta_apply_single_transform` sizes the palette channel as `(num_colors + num_deltas, num_channels)`, so the inner transform declares a 0x1 channel:

https://github.com/libjxl/jxl-rs/blob/a61dfa7647e095a2cef15d4ad58f1e06c8222b92/jxl/src/frame/modular/transforms/meta_apply.rs#L192-L197

That 0x1 channel becomes the input of the outer palette transform. But `Image` deliberately allocates nothing for a channel with a zero-sized dimension and reports it as 0x0:

https://github.com/libjxl/jxl-rs/blob/a61dfa7647e095a2cef15d4ad58f1e06c8222b92/jxl/src/image/internal.rs#L204-L207

So in `local_apply` the input has already been allocated and reports 0x0, while the output is still a `Placeholder` holding the declared 0x1 and thus fails the assert_eq.

This PR normalizes the size of `LocalTransformBuffer::Placeholder` to the size the channel will actually have once allocated, so the declared and allocated sizes won't disagree. The channel list that `meta_apply_single_transform` works on keeps the declared sizes, so the `check_equal_channels` bookkeeping of later transforms is unchanged.

Thanks!


jxl/src/tests/api.rs
```diff
@@ -1327,3 +1327,29 @@ fn test_fuzzer_context_map_num_histograms_overflow() {
...
+/// Regression test: two nested palette transforms, where the inner one has no colors and no
+/// deltas and so produces a 0x1 palette channel. `Image` allocates such a channel as 0x0, and
+/// applying the outer palette on top of it used to compare the declared 0x1 size against the
+/// allocated 0x0 one and panic. The file is malformed further on, so decoding it must fail --
+/// but with an error rather than a panic.
+#[test]
+fn test_fuzzer_modular_palette_empty_meta_channel() {
+    let data = include_bytes!("../../tests/testdata/modular_palette_empty_meta_channel.jxl");
+    for chunk_size in [usize::MAX, 1] {
+        for use_simple_pipeline in [false, true] {
```
veluca93: do we need to test with both pipelines? I don't think it should affect anything (and neither the two chunk sizes, I think)

jxl/src/frame/modular/transforms/apply_local.rs
```diff
@@ -354,3 +354,45 @@ impl TransformStep {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
veluca93: I think the general regression test is sufficient :-)

veluca93 (approved):

PR #797, opened 2026-06-09
More modular fast paths and optimizations

veluca93:
Also moves the RLE entropy coder special case to be used only for fast_lossless.

General 1-2% speedup of modular images, ~20% speedup of fast_lossless.

Includes improvements from #718 .

```
main, green queen:
Decoded 77394600 pixels in 4.464 seconds: 17.336 MP/s
Decoded 77394600 pixels in 4.545 seconds: 17.029 MP/s
Decoded 77394600 pixels in 4.487 seconds: 17.248 MP/s

main, sunset:
Decoded 64033200 pixels in 5.611 seconds: 11.412 MP/s
Decoded 64033200 pixels in 5.500 seconds: 11.643 MP/s
Decoded 64033200 pixels in 5.504 seconds: 11.634 MP/s

main, artificial e1:
Decoded 314572800 pixels in 4.818 seconds: 65.286 MP/s
Decoded 314572800 pixels in 4.795 seconds: 65.601 MP/s
Decoded 314572800 pixels in 4.835 seconds: 65.066 MP/s

new, green queen:
Decoded 77394600 pixels in 4.427 seconds: 17.484 MP/s
Decoded 77394600 pixels in 4.412 seconds: 17.541 MP/s
Decoded 77394600 pixels in 4.424 seconds: 17.495 MP/s

new, sunset:
Decoded 64033200 pixels in 5.394 seconds: 11.870 MP/s
Decoded 64033200 pixels in 5.298 seconds: 12.087 MP/s
Decoded 64033200 pixels in 5.390 seconds: 11.881 MP/s

main, artificial e1:
Decoded 314572800 pixels in 4.019 seconds: 78.266 MP/s
Decoded 314572800 pixels in 3.890 seconds: 80.864 MP/s
Decoded 314572800 pixels in 3.946 seconds: 79.711 MP/s
```

PR #912, opened 2026-08-29
Fix panic in the Modular RLE fast path when LZ77 is disabled

sl91994:
## Summary

`Histograms::is_rle()` never checks whether LZ77 is enabled, so a **perfectly valid Modular stream that does not use LZ77 can be routed into the RLE fast path**, where `decode_fast_lossless()` panics on an `unwrap()`.

The very same stream decodes without any error on the normal path: every residual is `unpack_signed(1) == -1`, so the Gradient predictor yields an ordinary ramp image. This is therefore an input the decoder itself considers valid, which only the fast path mishandles.

It is a regression from `afd41c9` More modular fast paths and optimizations (#797), which introduced the RLE fast path; `jxl` 0.5.0 and later on crates.io are affected.

## Root cause

`decode_modular_subbitstream()` dispatches to the fast path on `can_decode_fast_lossless()` alone.

```rust
if can_decode_fast_lossless(tree) {
    decode_fast_lossless(buffers, tree, br, partial_decoded_buffers)?
} else {
    // normal path
}
```

`can_decode_fast_lossless()` requires three things:

1. every node is a `Split { property: 0, .. }` or a `Leaf { Gradient, offset: 0, multiplier: 1, .. }`
2. `tree.histograms.is_rle()` is `true`
3. the histograms use prefix codes

Condition 2 is the problem (`jxl/src/entropy_coding/decode.rs`):

```rust
pub fn is_rle(&self) -> bool {
    let lz_dist_cluster = self.lz_dist_cluster as usize;
    let lz_conf = &self.uint_configs[lz_dist_cluster];
    self.codes.single_symbol(lz_dist_cluster) == Some(1) && lz_conf.is_split_exponent_zero()
}
```

This asks whether every LZ77 copy has distance 1, by inspecting the histogram of the **LZ77 distance cluster** — but it never checks that LZ77 is enabled at all. `Histograms::decode()` pins `lz_dist_cluster` to `0` when LZ77 is disabled, so `is_rle()` ends up inspecting **cluster 0, which is a real pixel-data histogram**.

Any stream whose cluster 0 codes the **constant symbol `1`** with a **split-exponent-zero uint config** is therefore reported as RLE even with LZ77 off. That is not an exotic shape: a single-leaf tree gives `num_contexts == 1`, so the pixel histograms *are* cluster 0.

Conditions 1 and 3 (a single Gradient leaf and prefix codes) are close to the minimal Modular stream, and all three conditions are decided purely by values `Tree::read()` builds from the bitstream.

With all three met, `decode_fast_lossless()` panics here:

```rust
let min_sym = tree.histograms.lz77_params().min_symbol.unwrap(); // None without LZ77
let min_len = tree.histograms.lz77_params().min_length.unwrap() as usize;
// ...
let lz_conf = tree.histograms.lz77_length_uint(); // unwraps internally
```

`min_symbol` and `min_length` are `None` without LZ77 (they are `Option` via `#[condition(enabled)]` on `Lz77Params`), and so is `lz77_length_uint`, so the first `unwrap()` panics deterministically.

## Impact

- All Modular decoding converges on `decode_modular_subbitstream()` (5 call sites in `jxl/src/frame/modular/mod.rs`), so such a file panics regardless of which entry point it is handed to.
- This is a decoder-side panic; it does not lead to memory corruption, out-of-bounds access, or information disclosure.
- Downstream applications built with `panic = "abort"` will abort the process.

## Fix

Check `lz77_params.enabled` first in `is_rle()`, and do not report such histograms as RLE.

```rust
/// Whether every LZ77 copy has distance 1, i.e. the stream is a plain run-length
/// encoding. False without LZ77: there is no distance cluster then (`lz_dist_cluster`
/// falls back to cluster 0, which holds real data), and the parameters an RLE decoder
/// needs (`min_symbol`, `min_length`, `lz77_length_uint`) are `None`.
pub fn is_rle(&self) -> bool {
    if !self.lz77_params.enabled {
        return false;
    }
    let lz_dist_cluster = self.lz_dist_cluster as usize;
    let lz_conf = &self.uint_configs[lz_dist_cluster];
    self.codes.single_symbol(lz_dist_cluster) == Some(1) && lz_conf.is_split_exponent_zero()
}
```

Notes on the approach:

- **Fixed at the source.** Without LZ77 there is no distance cluster, so the question `is_rle()` asks does not even apply. That makes the predicate itself the right place to fix, rather than guarding at the call site.
- **Consistent with the existing style.** The crate's other fast paths, `Histograms::can_use_config_420_fast_path()` and `run_on_specialized_tree()` in `specialized_trees.rs`, already gate on `lz77_params.enabled`. `is_rle()` was the only one missing that guard.
- **Narrow blast radius.** `is_rle()` has exactly one production call site, `bitstream.rs:37`.
- **No performance regression.** Genuinely RLE-coded streams have LZ77 enabled and still take the fast path (covered by the tests below).
- **Falls back to the normal path rather than erroring out.** These streams are valid input, so they are decoded correctly instead of being rejected.
- `is_rle()` had no doc comment, so this adds one covering both the reason for the guard and the non-obvious criterion: `single_symbol == Some(1)` together with `split_exponent == 0` means distance 1 by way of `Lz77State::SPECIAL_DISTANCES[1] == (1, 0)`.

## Tests

Three tests are added in `jxl/src/frame/modular/decode/bitstream.rs`.

Two hand-assembled Modular sub-bitstreams are fed to `decode_modular_subbitstream()`. The `GroupHeader`, the local `Tree`, the tree histograms and the pixel histograms are all built by the real decoder from real bits, so the tests follow exactly the path a crafted file would take. Both streams code the same 8x8 image (a constant residual of `-1`, which Gradient turns into a ramp) and differ only in whether LZ77 is enabled in the pixel histograms.

| Test | What it pins down |
| --- | --- |
| `lz77_disabled_histogram_is_not_rle` | With LZ77 disabled, `is_rle()` returns `false` and `can_decode_fast_lossless()` does not select the fast path. It also asserts `single_symbol(0) == Some(1)`, so the input is guaranteed to still have the shape the old `is_rle()` accepted. |
| `lz77_disabled_stream_decodes_on_the_normal_path` | The same bytes must not merely avoid panicking: they must decode to the expected ramp on the normal path. |
| `lz77_enabled_rle_stream_still_uses_the_fast_path` | A genuinely RLE-coded stream (LZ77 enabled, distance cluster coding the constant symbol 1) must still take the fast path and produce what the normal path produces. |

The third test guards against over-fixing: a change that makes `is_rle()` always return `false`, or that disables the fast path altogether, fails it.

## Verification

| Test | Without the fix | With the fix |
| --- | --- | --- |
| `lz77_disabled_histogram_is_not_rle` | FAILED | ok |
| `lz77_disabled_stream_decodes_on_the_normal_path` | FAILED (panic below) | ok |
| `lz77_enabled_rle_stream_still_uses_the_fast_path` | ok | ok |

Before the fix (`--release`):

```
thread '...::lz77_disabled_stream_decodes_on_the_normal_path' panicked at
jxl/src/frame/modular/decode/bitstream.rs:72:64:
called `Option::unwrap()` on a `None` value
stack backtrace:
   3: core::option::unwrap_failed
   4: <core::option::Option<u32>>::unwrap
   5: jxl::frame::modular::decode::bitstream::decode_fast_lossless
   6: jxl::frame::modular::decode::bitstream::decode_modular_subbitstream
             at ./src/frame/modular/decode/bitstream.rs:200:9
```

After the fix:

```
$ cargo test -p jxl --release --lib
test result: ok. 1049 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out

$ cargo fmt --check              # no diff
$ cargo clippy --all-targets     # no warnings
$ ./ci/pull_request_checks.sh    # PASS
```

## Note

I first reported this to the JPEG XL VMT by email. The reply was "I would not consider this a security issue at this point in time - feel free to open an issue / pull request on the repository.", so I am submitting it as a regular pull request.


jxl/src/frame/modular/decode/bitstream.rs
```diff
@@ -233,3 +233,269 @@ pub(in crate::frame::modular) fn decode_modular_subbitstream(
 
     Ok(())
 }
+
+#[cfg(test)]
```
veluca93: Please transform these tests into a bitstream that reproduces the error, like the tests in https://github.com/libjxl/jxl-rs/pull/910.

sl91994: Replaced the hand-assembled sub-bitstream tests with two .jxl files in jxl/tests/testdata/, following #910: modular_rle_fast_path_without_lz77.jxl reproduces the panic (it now decodes on the normal path), and modular_rle_fast_path.jxl is a genuinely RLE-coded stream that still takes the fast path and decodes to the same pixels.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `d3c3eee`)

| desktop | Single | 90.62 | 91.18 | +0.47% ± 0.30% |
| desktop | Multi | 422.61 | 419.72 | -0.27% ± 0.39% |
| framework-desktop | Single | 102.09 | 101.67 | +0.03% ± 0.19% |
| framework-desktop | Multi | 569.32 | 573.59 | +0.30% ± 0.80% |
| pixel7a | Single (Fast) | 40.18 | 41.86 | +1.69% ± 0.42% |
| pixel7a | Single (Mid) | 28.66 | 28.70 | +0.38% ± 0.33% |
| pixel7a | Multi | 92.17 | 92.61 | -1.14% ± 3.22% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_912.html)

PR #915, opened 2026-08-30
Reduce memory usage and increase speed of noise generation.

### LINKED ISSUE #896: Speed up noise generation
https://github.com/libjxl/jxl-rs/pull/895 shows that using f16 for VarDCT color channels is not (easily) a win.
However, we can use similar ideas for improving noise generation:
1) We can use u16 to store the input to ConvolveNoise, storing the top 16 mantissa bits. This loses the 7 lowest bits of the mantissa, which should not be a problem; then, ConvolveNoise can read those bits and shift them appropriately.
2) We should SIMDfy noise generation.

veluca93:
Fixes #896

### DIFF
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -133,6 +133,7 @@ fn upsample_lf_group(
                     &input_channels,
                     &mut output_channels,
                     Some(state.as_mut()),
+                    false,
                 );
             }
 
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -615,10 +615,8 @@ impl Frame {
         let buf_xsize = buf_x1.min(upsampled_size.0) - (gx * upsampling * group_dim) as usize;
         let buf_ysize = buf_y1.min(upsampled_size.1) - (gy * upsampling * group_dim) as usize;
 
-        let bits_to_float = |bits: u32| f32::from_bits((bits >> 9) | 0x3F800000);
-
         // Get all 3 noise channel buffers upfront
-        let mut bufs = [
+        let mut bufs: [Image<u16>; 3] = [
             pipeline!(self, p, p.get_buffer(num_channels)?),
             pipeline!(self, p, p.get_buffer(num_channels + 1)?),
             pipeline!(self, p, p.get_buffer(num_channels + 2)?),
@@ -676,7 +674,7 @@ impl Frame {
                                 } else {
                                     (batch[k] & 0xFFFFFFFF) as u32
                                 };
-                                row[x] = bits_to_float(bits);
+                                row[x] = (bits >> 16) as u16;
                             }
                         }
                     }
--- a/jxl/src/frame/decode.rs
+++ b/jxl/src/frame/decode.rs
@@ -5,6 +5,7 @@
 
 use std::collections::BTreeSet;
 
+use jxl_simd::{SimdDescriptor, simd_function};
 use jxl_transforms::transform_map::*;
 
 use super::block_context_map::BlockContextMap;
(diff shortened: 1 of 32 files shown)


kepstin:
Benchmarks on my test systems:

2026-08-30-3-i7-4710mq-adwaita-l

2026-03-30-3-ryzen-5950x-adwaita-l

All benchmarks run with num iterations set to target a run time of 10-15s, short enough to stay within short term turbo boost period on the CPU. The system was allowed to cool to steady state between runs. Although not noted on the chart, libjxl is version 0.12.0, built with the same clang/llvm version as used by rustc for jxl-rs, using -O3 and thin lto. jxl-rs is built using the release profile.

veluca93:
### Performance Summary (Commit `072fbeb`)

| desktop | Single | 90.72 | 91.29 | +0.70% ± 0.27% |
| desktop | Multi | 415.20 | 415.03 | +0.22% ± 0.47% |
| framework-desktop | Single | 96.70 | 101.16 | +0.98% ± 0.18% |
| framework-desktop | Multi | 568.00 | 572.25 | +0.19% ± 0.95% |
| pixel7a | Single (Fast) | 40.44 | 40.43 | -0.23% ± 0.62% |
| pixel7a | Single (Mid) | 28.63 | 28.72 | -0.50% ± 0.56% |
| pixel7a | Multi | 91.99 | 90.34 | -0.35% ± 2.88% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_915.html)

jxl/src/render/low_memory_pipeline/render_group.rs
```diff
@@ -332,6 +332,9 @@ impl LowMemoryRenderPipeline {
             .map(|c| BufferFiller::new(self, c, (x0, xsize), (gx, gy), y0..image_area.end().1))
             .collect();
 
+        let mut previous_call_was_previous_row: SmallVec<bool, 32> =
+            std::iter::repeat_n(false, self.shared.stages.len()).collect();
```
tirr-c, line 336: Maybe we could use `smallvec::smallvec![false; self.shared.stages.len()]`.

(Edit: I just forgot that we're using our own implementation...)

jxl/src/render/simple_pipeline/run_stage.rs
```diff
@@ -51,7 +51,7 @@ impl<T: RenderPipelineInPlaceStage> RunInPlaceStage<Image<f64>> for T {
                     }
                 }
                 let mut row: Vec<_> = buffer.iter_mut().map(|x| x as &mut [_]).collect();
-                self.process_row_chunk((x, y), xsize, &mut row, state.as_deref_mut());
+                self.process_row_chunk((x, y), xsize, &mut row, state.as_deref_mut(), false);
```
tirr-c, line 54: Is it okay to always pass `false` here?

jxl/src/frame/decode.rs
```diff
@@ -148,6 +150,50 @@ fn upsample_lf_group(
...
+        let row = &mut buf.row_mut(sub_y0 + y)[sub_x0..sub_x0 + sub_xsize];
+        for batch_index in 0..num_full {
+            let dest = &mut row[batch_index * 16..][..16];
+            rng.fill_u16_simd(d, dest);
+        }
+        if rem > 0 {
+            let mut temp = [0u16; 16];
+            rng.fill_u16_simd(d, &mut temp);
+            row[num_full * 16..].copy_from_slice(&temp[..rem]);
+        }
```
tirr-c: Can we use `.chunks_exact_mut()` and `iter.into_remainder()` here?

tirr-c (approved):

jxl/src/render/simple_pipeline/run_stage.rs
veluca93: Yup, I added a comment that clarifies this.

jxl/src/frame/decode.rs
veluca93: We can, and it's a fair bit nicer

tirr-c (approved):

PR #926, opened 2026-09-02
Extract Exif box

tirr-c:
With optional decompression support via `brotli` feature. (One can always get raw data even without the feature.)


jxl/src/api/inner/box_parser.rs
```diff
@@ -42,13 +44,66 @@ struct OOOJxlpBox {
     is_last: bool,
 }
 
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash)]
+pub enum AuxBoxType {
+    #[default]
+    None,
```
veluca93: What does this variant mean? In particular, what's the difference between from_code returning None and Some(AuxBoxType::None)?

jxl/src/api/inner/box_parser.rs
veluca93, line 169: Feels like new() should call into with_aux_boxes now :-)

jxl/src/api/inner/box_parser.rs
```diff
@@ -61,6 +116,16 @@ pub(super) struct BoxParser {
     // to box info
     codestream_pos_to_box: BTreeMap<u64, BoxParserCheckpoint>,
     allow_checkpoint: bool,
+
+    aux: AuxBoxState,
+}
+
+#[derive(Default)]
+struct AuxBoxState {
+    box_to_extract: HashSet<AuxBoxType>,
```
veluca93: s/box/boxes/?

jxl/src/api/decoder.rs
```diff
@@ -74,6 +76,92 @@ pub struct VisibleFrameSeekTarget {
...
+    pub fn raw_data(&self) -> &'a [u8] {
+        self.0.data()
+    }
+
+    pub fn is_compressed(&self) -> bool {
+        self.0.is_compressed()
+    }
+
+    #[cfg(feature = "brotli")]
+    pub fn data(&self) -> Result<std::borrow::Cow<'a, [u8]>> {
```
veluca93: Can we import Cow?

jxl/src/api/decoder.rs
```diff
@@ -74,6 +76,92 @@ pub struct VisibleFrameSeekTarget {
     pub visible_frames_to_skip: usize,
 }
 
+#[derive(Debug, Clone, Copy)]
```
veluca93: I'm not sure I fully see why we need those types - can't we use AuxBox and &'a AuxBox for everything? (moving the methods there)

jxl/src/api/decoder.rs
```diff
@@ -74,6 +76,92 @@ pub struct VisibleFrameSeekTarget {
...
+pub struct JxlTrailingBox(super::AuxBox);
+
+impl From<super::AuxBox> for JxlTrailingBox {
+    fn from(value: super::AuxBox) -> Self {
+        Self(value)
+    }
+}
+
+impl JxlTrailingBox {
+    pub fn is_exif(&self) -> bool {
```
veluca93: Why not expose box_type()?

jxl/src/api/decoder.rs
```diff
@@ -171,6 +263,14 @@ impl JxlDecoder<WithImageInfo> {
         Ok(self.map_inner_processing_result(inner_result))
     }
 
+    pub fn process_trailing_data(
```
veluca93, line 180: (this is more of a general musing, I don't think we should do anything about it now)
I am wondering if it makes sense to keep around the typestate API at all - it's somewhat more cumbersome to use, and completely useless for FFI... wdyt?

jxl/src/error.rs
```diff
@@ -277,6 +277,9 @@ pub enum Error {
     TooManyExtraChannels(usize),
     #[error("No LF frame for level {0}")]
     NoLfFrame(u32),
+    #[cfg(feature = "brotli")]
```
veluca93: I would consider removing the cfg here so that the enum definition doesn't change if one enables the brotli feature.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `7ed65da`)

| desktop | Single | 89.87 | 89.72 | +0.07% ± 0.89% |
| desktop | Multi | 414.54 | 413.17 | -0.13% ± 1.43% |
| framework-desktop | Single | 98.10 | 102.83 | +0.63% ± 0.16% |
| framework-desktop | Multi | 505.93 | 506.53 | -0.30% ± 0.91% |
| pixel7a | Single (Fast) | 41.54 | 39.37 | -0.85% ± 0.52% |
| pixel7a | Single (Mid) | 29.68 | 28.25 | -0.88% ± 0.50% |
| pixel7a | Multi | 95.08 | 94.15 | +1.86% ± 2.93% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_926.html)

jxl/src/api/decoder.rs
tirr-c: Hmm maybe removing typestates makes it bit more convenient, but then I think API misuse errors should provide a bit more helpful message to guide users.

veluca93: > Hmm maybe removing typestates makes it bit more convenient, but then I think API misuse errors should provide a bit more helpful message to guide users.

I agree! But at this point I think we should consider this for 0.8.0.

jxl/src/api/inner/box_parser.rs
tirr-c: It was there to have `AuxBox` directly as a buffer instead of `Option<AuxBox>`, but I just removed it because it looked okay with `Option` :)

jxl/src/api/decoder.rs
tirr-c: `JxlTrailingBox` is there to indicate that there's no data in itself. Merged `JxlAuxBox` to `AuxBox` and renamed it to `JxlAuxBox`.

tirr-c: While fixing trailing box bug, `JxlTrailingBox` is fully merged into `JxlAuxBox` :sweat_smile:

veluca93 (approved):

tirr-c:
Oops I forgot to mark this as ready for review :stuck_out_tongue:

PR #928, opened 2026-09-02
16-bit modular buffers

veluca93:
Implement 16-bit modular buffers, giving memory reduction and speedups.

Includes the commits from #943.

### DIFF
--- a/jxl/src/frame/modular/decode/bitstream.rs
+++ b/jxl/src/frame/modular/decode/bitstream.rs
@@ -10,7 +10,7 @@ use crate::error::{Error, Result};
 use crate::frame::modular::predict::clamped_gradient;
 use crate::frame::modular::transforms::apply_local::meta_apply_local_transforms;
 use crate::frame::modular::tree::TreeNode;
-use crate::frame::modular::{ModularChannel, Predictor, TransformScratchSpace, Tree};
+use crate::frame::modular::{ModularChannel, Predictor, ScratchSpace, Tree};
 use crate::headers::JxlHeader;
 use crate::headers::modular::GroupHeader;
 
@@ -138,7 +138,7 @@ pub(in crate::frame::modular) fn decode_modular_subbitstream(
     global_tree: &Option<Tree>,
     br: &mut BitReader,
     partial_decoded_buffers: Option<&mut usize>,
-    transform_scratch_space: &mut TransformScratchSpace,
+    scratch_space: &mut ScratchSpace,
 ) -> Result<()> {
     // Skip decoding if all grids are zero-sized.
     let is_empty = buffers
@@ -229,7 +229,7 @@ pub(in crate::frame::modular) fn decode_modular_subbitstream(
     }
 
     for step in transform_steps.iter().rev() {
-        step.local_apply(&mut buffer_storage, transform_scratch_space)?;
+        step.local_apply(&mut buffer_storage, scratch_space)?;
     }
 
     Ok(())
--- a/jxl/src/frame/modular/decode/bitstream.rs
+++ b/jxl/src/frame/modular/decode/bitstream.rs
@@ -13,6 +13,7 @@ use crate::frame::modular::tree::TreeNode;
 use crate::frame::modular::{ModularChannel, Predictor, ScratchSpace, Tree};
 use crate::headers::JxlHeader;
 use crate::headers::modular::GroupHeader;
+use crate::image::ImageRectMut;
 
 // If we have at least this many bits still available to read,
 // we can be sure that none of the reads up to this point read garbage.
@@ -53,7 +54,7 @@ fn decode_fast_lossless(
 
     let mut last_safe_buf = 0;
     for (c, buf) in buffers.into_iter().enumerate() {
-        let (w, h) = buf.data.size();
+        let (w, h) = buf.size();
         if w == 0 || h == 0 {
             continue;
         }
@@ -76,7 +77,7 @@ fn decode_fast_lossless(
             .uint(tree.histograms.map_context_to_cluster(id as usize));
         let lz_conf = tree.histograms.lz77_length_uint();
 
-        let buf = &mut buf.data;
+        let mut buf = ImageRectMut::<i32>::from_raw(buf.data.as_rect_mut());
 
         let mut decode = {
             #[inline(always)]
@@ -99,7 +100,7 @@ fn decode_fast_lossless(
         };
 
         let mut last = 0i32;
-        for p in buf.row_mut(0) {
+        for p in buf.row(0) {
             // clamped gradient == left on the first row.
             *p = last.wrapping_add(decode());
             last = *p;
@@ -143,7 +144,7 @@ pub(in crate::frame::modular) fn decode_modular_subbitstream(
     // Skip decoding if all grids are zero-sized.
     let is_empty = buffers
         .iter()
-        .all(|buffer| matches!(buffer.data.size(), (0, _) | (_, 0)));
+        .all(|buffer| matches!(buffer.size(), (0, _) | (_, 0)));
     if is_empty {
         return Ok(());
     }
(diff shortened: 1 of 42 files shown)


veluca93:
### Performance Summary (Commit `a746c43`)

| desktop | Single | 90.34 | 87.11 | +0.12% ± 0.76% |
| desktop | Multi | 413.90 | 443.52 | +2.86% ± 1.27% |
| framework-desktop | Single | 99.14 | 103.42 | +1.65% ± 0.17% |
| framework-desktop | Multi | 512.68 | 564.92 | +3.12% ± 1.05% |
| pixel7a | Single (Fast) | 41.66 | 43.36 | +1.29% ± 0.52% |
| pixel7a | Single (Mid) | 28.40 | 32.49 | +2.92% ± 0.58% |
| pixel7a | Multi | 95.25 | 104.48 | +2.14% ± 2.90% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_928.html)

jxl_simd/src/lib.rs
```diff
@@ -1782,4 +1970,266 @@ mod test {
...
+                assert_eq!(res_ltz[i], if a[i] < 0 { a[i] } else { b[i] });
+            }
+
+            // eq
+            let mask_eq = va.eq(vb);
+            let mut res_eq = vec![0i16; len];
+            mask_eq.if_then_else_i16(va, vb).store(&mut res_eq);
+            for i in 0..len {
+                assert_eq!(res_eq[i], if a[i] == b[i] { a[i] } else { b[i] });
+            }
```
tirr-c, line 2098: `if a == b { a } else { b }` equals to `if a == b { b } else { b }`; this test passes even if `if_then_else_i16(va, vb)` unconditionally returns `vb`.

jxl_simd/src/lib.rs
```diff
@@ -1782,4 +1970,266 @@ mod test {
...
+
+            // all
+            assert_eq!(mask_eqz.all(), a.iter().all(|&x| x == 0));
+
+            // andnot
+            let mask_andnot = mask_gt.andnot(mask_eq);
+            let mut res_andnot = vec![0i16; len];
+            mask_andnot.if_then_else_i16(va, vb).store(&mut res_andnot);
+            for i in 0..len {
+                let expected_mask = !(a[i] > b[i]) && (a[i] == b[i]);
```
tirr-c: `!(a[i] > b[i])` is not effective at all in this expression; and then we have the same problem with `eq` above.

tirr-c (commented):
Reviewing commit by commit. These are fairly minor though 🙂 

(These are found via LLM, but I think the point is legit)

jxl/src/render/stages/convert.rs
```diff
@@ -619,6 +630,82 @@ impl RenderPipelineInOutStage for ConvertF32ToU8Stage {
...
+            .chunks_exact(simd_width)
+            .zip(output.chunks_exact_mut(simd_width))
+            .take(xsize.div_ceil(simd_width))
+        {
+            let val = D::I16Vec::load(d, input_chunk);
+            let scaled = if scale == 1 { val } else { val * scale_vec };
+            let zeroclip = scaled.lt_zero().if_then_else_i16(zero, scaled);
+            let clip = scaled.gt(max_vec).if_then_else_i16(max_vec, zeroclip);
+            clip.store_u8(output_chunk);
+        }
```
tirr-c: It seems that clip-then-scale is better, because when `val` has a large value outside the nominal range, scaling may overflow and wrap around to produce opposite value.

It's a bit of an edge case outside of the spec I think, feel free to keep this as is.

jxl/src/frame/modular/decode/bitstream.rs
```diff
@@ -134,6 +135,7 @@ fn decode_fast_lossless(
 // The intended use of passing a header is for the DcGlobal section.
 pub(in crate::frame::modular) fn decode_modular_subbitstream(
     buffers: Vec<&mut ModularChannel>,
+    is_16bit: bool,
```
tirr-c: Other functions have `is_16bit` at the end of argument list; it would be nice to match the order in one way.

jxl_simd/src/lib.rs
veluca93: Changed the test to store 1 or 0.

jxl_simd/src/lib.rs
veluca93: Changed the condition to !(a[i] > b[i]) && (a[i] < 0)

jxl/src/frame/modular/decode/bitstream.rs
tirr-c: Other functions have `is_16bit` at the end of argument list; it would be nice to match the order in one way.

jxl/src/render/stages/convert.rs
```diff
@@ -619,6 +630,82 @@ impl RenderPipelineInOutStage for ConvertF32ToU8Stage {
...
+        for (input_chunk, output_chunk) in input
+            .chunks_exact(simd_width)
+            .zip(output.chunks_exact_mut(simd_width))
+            .take(xsize.div_ceil(simd_width))
+        {
+            let val = D::I16Vec::load(d, input_chunk);
+            let scaled = if scale == 1 { val } else { val * scale_vec };
+            let zeroclip = scaled.lt_zero().if_then_else_i16(zero, scaled);
+            let clip = scaled.gt(max_vec).if_then_else_i16(max_vec, zeroclip);
+            clip.store_u8(output_chunk);
```
tirr-c: It seems that clip-then-scale is better, because when val has a large value outside the nominal range, scaling may overflow and wrap around to produce opposite value.

It's a bit of an edge case outside of the spec I think, feel free to keep this as is.

jxl/src/frame/modular/decode/bitstream.rs
veluca93: I like having is_16bit just after the buffers here :-)

jxl/src/render/stages/convert.rs
veluca93: This version matches the i32 code :-) (but I made it match slightly more)

It might be a good idea to change both - but I'd leave that as a follow-up.

jxl/src/frame/modular/decode/channel.rs
```diff
@@ -18,6 +18,51 @@ use crate::util::tracing_wrappers::*;
...
+        };
+    };
+}
+
+pub(super) fn sync_scratch(
+    buf: &mut ModularChannel,
+    y: usize,
+    scratch: Option<&mut [Vec<i32>; 3]>,
+) {
+    if let Some(scratch) = scratch {
```
tirr-c, line 59: Adding a comment about 16-bit Modular would be nice.

tirr-c (commented):
Wdyt about having some kind of enum instead of `bool` for `is_16bit`? For me it was a bit tricky to figure out the meaning of `true` and `false` in some context (especially `.size(true)`).

veluca93:
> Wdyt about having some kind of enum instead of `bool` for `is_16bit`? For me it was a bit tricky to figure out the meaning of `true` and `false` in some context (especially `.size(true)`).

Did so!

jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -184,6 +228,23 @@ fn unsqueeze_impl<D: SimdDescriptor>(
...
+    d: D,
+    avg: D::I16Vec,
+    res: D::I16Vec,
+    next_avg: D::I16Vec,
+    prev: D::I16Vec,
+) -> (D::I16Vec, D::I16Vec) {
+    let tendency = smooth_tendency_impl_i16(d, prev, avg, next_avg);
+    let diff = res + tendency;
+    let sign = shr!(diff, 15) & D::I16Vec::splat(d, 1);
+    let diff_2 = shr!(diff + sign, 1);
```
tirr-c: Maybe this could be `sign = shr!(diff, 15); diff_2 = shr!(diff - sign, 1);`

tirr-c (approved):

PR #929, opened 2026-09-02
Add image crate decoding integration

hjanuschka:
`image` keeps JPEG XL out of its built-in format set, but exposes process-wide decoder and signature hooks for external codecs. Add an opt-in `image` feature with `JxlDecoder`, an `image::ImageDecoder` adapter, and `register_image_decoding_hook()` so applications only need:

```rust
jxl::integration::register_image_decoding_hook();
let image = image::open("input.jxl")?;
```

The adapter streams from the seekable reader instead of buffering the file, selects grayscale/RGB output with alpha at 8-bit, 16-bit, or float precision, exposes the rendered ICC profile, honors `image::Limits`, and handles `ImageDecoder`'s unaligned-output-buffer contract. Registration covers both the bare codestream and container signatures as well as `.jxl` paths.

Tests cover direct decoding, alpha, unaligned 16-bit output, signature-based in-memory loading, and the path-based `image::open` flow. I also ran the provided thumbnailer-style flow end to end and wrote the decoded JXL as PNG.


jxl-image-rs-integration/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+        } else if bytes_per_sample == 2 {
+            self.limits.reserve_usize(buf.len())?;
+            let mut aligned = vec![0u16; buf.len() / 2];
+            decode_frame(
+                frame,
+                &mut self.input,
+                bytemuck::cast_slice_mut(&mut aligned),
+                self.height,
+            )?;
+            buf.copy_from_slice(bytemuck::cast_slice(&aligned));
```
veluca93, line 107: I don't love this copy - we should at least only do it if the user-provided slice is not aligned.

veluca93 (commented):
I'm not too convinced that adding image-rs support behind a feature flag is my preferred option - I think I might like having a separate crate more (i.e. jxl_image_rs_integration or something)
@tirr-c wdyt?

tirr-c:
Hmm yeah, we can isolate potential breaking changes of `image` from `jxl` if we have another crate for the integration.

hjanuschka:
done: https://github.com/hjanuschka/image-rs-jxl

hjanuschka:
moved the image-rs adapter into a separate `jxl-image-rs-integration` workspace crate, so it stays in-repo and can be refactored alongside `jxl`.

veluca93 (approved):

veluca93:
### Performance Summary (Commit `c9677d5`)

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_929.html)

PR #930, opened 2026-09-03
Better progressive decoding script

veluca93:


tools/progressive_decoding_demo.py
```diff
@@ -0,0 +1,366 @@
...
+        srt_file = output_dir / f"subtitles_{mode}.srt"
+        with open(srt_file, "w") as f:
+            f.write("\n".join(srt_lines) + "\n")
+
+        # Output video
+        video_clean = output_dir / f"progressive_{mode}.mp4"
+        print(f"Creating clean {mode} progressive video...")
+        subprocess.check_call(
+            [
+                "ffmpeg",
```
jonsneyers, line 313: Would be nice to first check if ffmpeg is actually available and show some helpful message if it's not, instead of producing all frames and then failing.

veluca93: I don't think it's too necessary, and you might be interested in the frames without the video :-)

tools/progressive_decoding_demo.py
```diff
@@ -0,0 +1,366 @@
...
+            "1",
+            "-update",
+            "1",
+            str(black_frame),
+        ]
+    )
+
+    # Gather partial frames
+    partial_files = sorted(
+        output_dir.glob("frame.partial_*.png"),
```
jonsneyers, line 203: These intermediate files (png frames, subtitles) are not cleaned up, are they? It would make sense to either clean them up or put them somewhere in output_dir/intermediate/ or something, and provide a command line to the user to clean them up.

veluca93: They are not, and I think it'd be a net negative to do so (they are fairly useful, at least the frames).
I am not sure it's too useful to move them in a subdirectory, and I think users can figure out the command line if they want to clear them up

tools/progressive_decoding_demo.py
```diff
@@ -0,0 +1,366 @@
...
+        )
+
+    if not binary.is_file():
+        print(f"Error: Binary not found at {binary} after build.", file=sys.stderr)
+        sys.exit(1)
+
+    output_dir = args.output_dir.resolve()
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    # Clean previous runs in output_dir
```
jonsneyers, line 150: Silently deleting results from previous runs is a bit harsh imo; possibly it's more convenient to produce filenames that are based on the input filename (so you can produce results for multiple inputs conveniently) and to add "-1", "-2" to that filename if the output is already there, or at least produce some warning saying that previous output has been deleted/overwritten.

tools/progressive_decoding_demo.py
jonsneyers: Agreed that keeping them around can be useful; possibly might make sense to have an option to keep them or not, with default not.

tools/progressive_decoding_demo.py
jonsneyers: Depends on how polished/user-friendly you want to make this thing; for a demo script it's indeed ok if it's not very polished, so feel free to just resolve this and the others without action.

jonsneyers (approved):

tools/progressive_decoding_demo.py
veluca93: This matches the previous script, we can change it later if we want to :-)

veluca93:
### Performance Summary (Commit `0063f1f`)

| desktop | Single | 89.63 | 89.28 | -0.85% ± 0.93% |
| desktop | Multi | 411.84 | 412.46 | -0.17% ± 1.36% |
| framework-desktop | Single | 102.68 | 98.68 | -0.59% ± 0.17% |
| framework-desktop | Multi | 521.90 | 517.53 | -0.32% ± 0.80% |
| pixel7a | Single (Fast) | 42.78 | 41.60 | -0.24% ± 0.34% |
| pixel7a | Single (Mid) | 29.69 | 30.43 | +0.43% ± 0.37% |
| pixel7a | Multi | 95.36 | 94.68 | +0.06% ± 3.51% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_930.html)

PR #937, opened 2026-09-05
Avoid recapturing auxiliary boxes after frame seeks

gregbenz:
Auxiliary boxes remain stored when `start_new_frame` resets the parser to a checkpoint. The replay guard currently compares an index over all auxiliary candidates with the number of distinct retained box types. Those values diverge when a box type repeats or an unrequested box is encountered, so replay can append boxes that were already captured. Repeated seeks can grow the list again each time.

Track the checkpoint's auxiliary candidate position separately from the number of candidates already accounted for. Replayed candidates are skipped, while genuinely new trailing boxes can still be captured. If a reset discards an unfinished capture, roll that count back so replay captures the box again.

Regression coverage is provided separately in #939. I verified locally that its test fails without this fix and passes with it.

This is a follow-up fix for the auxiliary-box capture added in #926. It does not change box contents, decoding, or the public API.

### DIFF
--- a/jxl/src/api/inner/box_parser.rs
+++ b/jxl/src/api/inner/box_parser.rs
@@ -134,7 +134,7 @@ pub struct BoxParserCheckpoint {
     codestream_left: Option<u64>,
     is_valid_checkpoint: bool,
     pub(crate) consumed_codestream: u64,
-    num_aux_boxes: usize,
+    next_aux_box_idx: usize,
 }
 
 pub(super) struct BoxParser {
@@ -156,6 +156,7 @@ struct AuxBoxState {
     boxes: HashMap<JxlAuxBoxType, Vec<JxlAuxBox>>,
     box_buffer: Option<JxlAuxBox>,
     next_box_idx: usize,
+    seen_box_count: usize,
 }
 
 impl BoxParser {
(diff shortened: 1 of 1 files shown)


jxl/src/api/inner/box_parser.rs
```diff
@@ -669,11 +682,7 @@ impl BoxParser {
                 let inner_type = JxlAuxBoxType(inner_type.try_into().unwrap());
                 self.local_buffer.consume(4);
 
-                let current_box_idx = self.aux.next_box_idx;
-                self.aux.next_box_idx += 1;
-                if current_box_idx >= self.aux.boxes.len()
-                    && self.aux.boxes_to_extract.contains(&inner_type)
-                {
+                if self.aux.should_extract(inner_type) {
```
tirr-c: It would be better if the state update code is also moved to the function. Maybe move `should_extract` to `BoxParser`, merge and rename accordingly?

tirr-c (commented):
I believe the test code can be rewritten in more concise way; I'll try writing one in another PR.

tirr-c (commented):
I opened #939 for the regression test, it should work equally well.

Also some other parser bugs are fixed in `main`, please rebase.

gregbenz:
Thanks for the regression test in #939 and the parser fixes. I’ve rebased onto current main and consolidated the auxiliary-box setup and state transitions into a BoxParser helper. I removed the overlapping test in favor of #939 and confirmed that your test fails without this fix and passes with it.

tirr-c (approved):

PR #946, opened 2026-09-07
Add a reader for captured gain-map data

gregbenz:
Applications can capture a JPEG XL gain-map box, but still need to unpack its contents before using it. This adds a reader that exposes the metadata, alternate color information, and embedded JPEG XL image without copying the payload. It also provides helpers to decode the alternate color encoding and compressed ICC profile using existing code.

The example shows how to capture a finite, uncompressed box and decode its embedded image. Applications remain responsible for interpreting the gain-map metadata and applying it to the base image.

Tests cover the color-field combinations, embedded containers, malformed input, and exact image samples. A local downstream experiment also verified that the extracted metadata and decoded profiles work with libultrahdr and LittleCMS. All 20 Adobe samples and seven synthetic color variants passed through the example on the final source. The library suite also passed 1,164 tests with one ignored. The Adobe images and downstream experiment are not included in this PR.

This follows up on @hjanuschka's work in #504. Thanks, Helmut, for the original work and your support.


veluca93:
FYI for the next week or two I plan to only merge bugfixes (and/or improvements to the test infrastructure to more easily find bugs).

jxl/src/api/gain_map.rs
```diff
@@ -0,0 +1,128 @@
...
+use super::JxlColorEncoding;
+use crate::bit_reader::BitReader;
+use crate::error::{Error, Result};
+use crate::headers::color_encoding::{ColorEncoding, ColorSpace};
+use crate::headers::encodings::{Empty, UnconditionalCoder};
+use crate::icc::IncrementalIccReader;
+
+/// The alternate color encoding carried by a gain-map bundle.
+#[derive(Clone, Debug, PartialEq)]
+pub enum JxlGainMapColorEncoding {
```
veluca93: Why is this not a JxlColorProfile?

gregbenz: Good point. I used the separate enum to keep ICC decoding separate, but we can use JxlColorProfile here. The helper now returns Simple for structured color and decodes the ICC field when the color encoding requires it. The raw fields and separate ICC helper are still available.

jxl/examples/gain_map.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+fn run(path: &Path) -> ExampleResult<()> {
+    let input = fs::read(path)?;
+    let raw_bundle = read_gain_map_box(&input)?;
+    let bundle = JxlGainMapBundle::parse(&raw_bundle)?;
+    let (color, alternate_icc_bytes) = match bundle.decode_color_encoding()? {
+        None => (
+            "absent".to_owned(),
+            bundle.decode_alternate_icc()?.map_or(0, |icc| icc.len()),
```
hjanuschka, line 121: Should this call jump_to_byte_boundary() and verify that the entire compressed_icc field was consumed? As written, a valid ICC stream followed by arbitrary bytes, or non-zero padding, is accepted. The normal codestream ICC path calls jump_to_byte_boundary() after finalize().

jxl/examples/gain_map.rs
```diff
@@ -0,0 +1,146 @@
...
+        complete(frame.process(&mut input, std::slice::from_mut(&mut output), None)?)?;
+    }
+    let checksum = pixels.iter().map(|&sample| u64::from(sample)).sum();
+    Ok((width, height, pixels.len(), checksum))
+}
+
+fn run(path: &Path) -> ExampleResult<()> {
+    let input = fs::read(path)?;
+    let raw_bundle = read_gain_map_box(&input)?;
+    let bundle = JxlGainMapBundle::parse(&raw_bundle)?;
```
hjanuschka, line 112: Since parse() deliberately leaves unknown versions for the caller to handle, should this example reject bundle.version != 0 before interpreting the fields? 

Otherwise it demonstrates silently decoding future bundle versions as version 0.

jxl/tests/gain_map.rs
```diff
@@ -0,0 +1,298 @@
...
+        assert!(parsed.decode_color_encoding().is_ok());
+    }
+}
+
+#[test]
+fn captures_finite_aux_box_and_decodes_bare_and_container_gain_maps() {
+    let outer = include_bytes!("testdata/gain_map/synthetic-container.jxl");
+    let captured = capture_gain_map(outer);
+    assert_eq!(captured, include_bytes!("testdata/gain_map/combined.jhgm"));
+    let captured_bundle = JxlGainMapBundle::parse(&captured).unwrap();
```
hjanuschka, line 406: Is accepting an embedded container intentional? libjxl documents this field as a naked JPEG XL codestream. 

This test appears to bless container-wrapped data as valid jhgm content. If this is compatibility behavior for real-world files, could we document that distinction explicitly?

jxl/tests/testdata/gain_map/README.md
```diff
@@ -0,0 +1,17 @@
+# Gain-map reader fixtures
+
+These small fixtures were generated locally for the `JxlGainMapBundle` API
```
hjanuschka: Could we include the fixture-generation script or exact commands? The commit hash helps, but "JPEG XL bundle writer" and "generated locally" are not enough to reproduce these binary test files when they need updating.

jxl/tests/gain_map.rs
```diff
@@ -0,0 +1,298 @@
...
+            valid_icc.gain_map,
+        );
+        let parsed = JxlGainMapBundle::parse(&data).unwrap();
+        assert!(parsed.decode_color_encoding().is_err());
+        assert_eq!(parsed.decode_alternate_icc().unwrap().as_deref(), Some(ICC));
+    }
+}
+
+#[test]
+fn rejects_truncated_and_invalid_fields_without_losing_raw_slices() {
```
hjanuschka, line 271: Could we add an `arbtest` property that calls `JxlGainMapBundle::parse()` and both decode helpers on arbitrary input? This is a small untrusted-input parser with nested length fields and would benefit from a no-panic fuzz-style test beyond the selected truncation cases.

jxl/tests/gain_map.rs
```diff
@@ -0,0 +1,298 @@
...
+    ] {
+        let data = raw_bundle(&[], color, &[], &[]);
+        let parsed = JxlGainMapBundle::parse(&data).unwrap();
+        assert!(parsed.decode_color_encoding().is_ok());
+    }
+}
+
+#[test]
+fn captures_finite_aux_box_and_decodes_bare_and_container_gain_maps() {
+    let outer = include_bytes!("testdata/gain_map/synthetic-container.jxl");
```
hjanuschka, line 403: Could we add a `#[cfg(feature = "brotli")]` case for a `brob`-wrapped `jhgm` box and pass `JxlAuxBox::data()` into `JxlGainMapBundle::parse()`? That would test the supported path from compressed box capture through bundle parsing.

jxl/tests/gain_map.rs
```diff
@@ -0,0 +1,298 @@
...
+            .is_err()
+    );
+
+    let mut unknown_version = include_bytes!("testdata/gain_map/structured.jhgm").to_vec();
+    unknown_version[0] = 0xff;
+    let parsed_unknown = JxlGainMapBundle::parse(&unknown_version).unwrap();
+    assert_eq!(parsed_unknown.version, 0xff);
+    assert!(parsed_unknown.decode_color_encoding().is_ok());
+
+    for color in [
```
hjanuschka: Why should these encodings be accepted? They decode as `all_default`, but contain non-zero padding or an entire trailing byte. Since the outer field provides an exact length, accepting ignored data makes malformed encodings indistinguishable from canonical ones. Could we validate byte alignment and full field consumption instead?

jxl/examples/gain_map.rs
gregbenz: Thanks for the careful / detailed review, Helmut - very much appreciated. The ICC helper now checks zero padding and full consumption after decoding. Added tests for nonzero padding and trailing bytes.

jxl/examples/gain_map.rs
gregbenz: Agreed. The example now rejects versions other than 0 before decoding the fields.

jxl/tests/gain_map.rs
gregbenz: You're right. This is a synthetic byte-preservation case, not evidence of valid jhgm content or behavior seen in real files. I've separated the test and made that distinction explicit in the fixture documentation.

jxl/tests/testdata/gain_map/README.md
gregbenz: Added the generation tools and reproduction commands. All 9 fixtures regenerate byte-for-byte.

jxl/tests/gain_map.rs
gregbenz: Added arbtest coverage for parsing and both helpers, including generated envelopes with arbitrary field contents so the helpers get exercised too.

jxl/tests/gain_map.rs
gregbenz: Added a Brotli-gated test that captures a brob-wrapped jhgm box, passes data(&[]) to the bundle parser, and checks the decoded contents.

jxl/tests/gain_map.rs
gregbenz: Agreed. I changed those tests to require complete field validation. The helper now checks zero padding and full consumption for both structured and ICC-required encodings.

PR #956, opened 2026-09-09
Don't treat an empty out-of-order jxlp box as end of input

lilith:
Fixes a decode failure on valid files from stock `cjxl 0.12.0 --output_mode=2`.

**Full write-up with traces, bisection and repro files:** https://claude.ai/code/artifact/84edd665-369b-4baa-84d0-5a39d4409128

## The bug

`jxl-rs` fails with **"Source file truncated"** on lossless files written with `--output_mode=2` (out-of-order `jxlp` streaming output) when the frame has **four or more DC groups**. `djxl 0.12.0` decodes all of them. DC groups are 2048 px, so one pixel of width flips it:

| file | dims | DC groups | before | after |
|---|---|---|---|---|
| `pair-6144x256-3dcgroups-OK.jxl` | 6144×256 | 3 (3×1) | OK | OK |
| `pair-6145x256-4dcgroups-FAIL.jxl` | 6145×256 | 4 (4×1) | **truncated** | OK |
| `min-8192x128-lossless-oo-FAIL.jxl` | 8192×128 | 4 (4×1) | **truncated** | OK |
| `square-2049x2049-lossless-oo-FAIL.jxl` | 2049×2049 | 4 (2×2) | **truncated** | OK |
| `control-8192x128-lossy-oo-OK.jxl` | 8192×128 | 4 (4×1), lossy | OK | OK |

Repro files (synthetic gradients, ~62 KB total) and a `make_repro.sh` that regenerates them from scratch: <https://codec-corpus.r2.imazen.org/repro/jxl-rs-ooo-jxlp-2026-09-08/>

```sh
BASE=https://codec-corpus.r2.imazen.org/repro/jxl-rs-ooo-jxlp-2026-09-08
curl -sSO $BASE/pair-6144x256-3dcgroups-OK.jxl
curl -sSO $BASE/pair-6145x256-4dcgroups-FAIL.jxl
cargo build --release -p jxl_cli
./target/release/jxl_cli pair-6144x256-3dcgroups-OK.jxl  /tmp/a.png   # decodes
./target/release/jxl_cli pair-6145x256-4dcgroups-FAIL.jxl /tmp/b.png  # Error: Source file truncated
```

## Cause

The failing lossless file and the working lossy file at the same dimensions have **identical `jxlp` box index order** (`0,1,3,4,9…16,5,17…24,6,25…32,7,(33ᴸ),8,2`). They differ only in payload length: for a modular frame `cjxl` writes the DC-group `jxlp` slots with **zero payload**, one per DC group.

`injected_jxlp` returns a buffered out-of-order box without checking whether it has bytes left:

https://github.com/libjxl/jxl-rs/blob/9e6b9f9/jxl/src/api/inner/box_parser.rs#L291-L300

so `read` reads an empty slice and returns `Ok(0)`:

https://github.com/libjxl/jxl-rs/blob/9e6b9f9/jxl/src/api/inner/box_parser.rs#L750-L757

and `available_bytes` returns `0`:

https://github.com/libjxl/jxl-rs/blob/9e6b9f9/jxl/src/api/inner/box_parser.rs#L736-L739

Callers take a 0-byte read as end of input — both of these `break` on `num == 0`:

https://github.com/libjxl/jxl-rs/blob/9e6b9f9/jxl/src/api/inner/process.rs#L46-L51

https://github.com/libjxl/jxl-rs/blob/9e6b9f9/jxl/src/api/inner/codestream_parser/frame_info.rs#L418-L421

With three DC groups the run of empty boxes is short enough that the retry path survives it; with four it is not. Traced on the 6145 px file:

```
read: INJECTED Jxlp(3) data_len=7 -> n=7 ; after=Jxlp(4) ooo_left=30
read: INJECTED Jxlp(4) data_len=0 -> n=0 ; after=Jxlp(5) ooo_left=29
read: INJECTED Jxlp(5) data_len=0 -> n=0 ; after=Jxlp(6) ooo_left=28
read: INJECTED Jxlp(6) data_len=0 -> n=0 ; after=Jxlp(7) ooo_left=27
read: INJECTED Jxlp(7) data_len=0 -> n=0 ; after=Jxlp(8) ooo_left=26
available_bytes: INJECTED -> 0
read: INJECTED Jxlp(8) data_len=0 -> n=0 ; after=Jxlp(9) ooo_left=25
Error: Source file truncated            <- 25 boxes of codestream still buffered
```

## The change

Step over exhausted or empty buffered boxes inside `injected_jxlp`. Doing it there covers `read`, `available_bytes` and `skip` together, since all three go through it. `check_ooo_jxlp_done` removes the box it advances past, so the loop terminates.

Happy to restructure this if you would rather the check live at the three call sites, or reject empty out-of-order boxes at parse time instead.

## Test

The regression test builds a container in memory, so it needs no fixture — this sits below the codestream parser and the payloads can be arbitrary. `ftyp` at version 1 (out-of-order `jxlp` is illegal at version 0), then boxes in the order that reproduces the bug: an empty box buffered ahead of real data, with the low index delivered last. It reads until `read` returns 0, the way every caller does.

Without the fix it reads back `"AAABBB"` instead of `"AAABBBCCCDDD"` — verified failing before and passing after.

## Verification

- All previously-failing files decode, and are **pixel-identical to `djxl`** (`magick compare -metric AE` = 0), up to a 63 MB 12000×9000 lossless file.
- `cargo test --release --workspace`: 1165 tests pass, 0 failures. `cargo fmt --check` and `cargo clippy --all-targets` clean.
- Apple M4 Pro, macOS 25.5, base `9e6b9f9` (v0.7.1), libjxl 0.12.0.

## Aside, not addressed here

Passing `jxl_cli` an output path whose format cannot be inferred (e.g. `/dev/null`) also reports "Source file truncated", which made this easy to misdiagnose. Happy to file that separately if it is worth fixing.

---

*Investigated, patched and written by Claude (Anthropic) driving the tools on my behalf. Every command and number above was actually run and the outputs transcribed, but the analysis is model-generated and has not been reviewed line by line by a human — please read the patch critically.*


jxl/src/api/inner/box_parser.rs
```diff
@@ -827,4 +841,53 @@ mod tests {
         // All input should be consumed.
         assert!(input.is_empty());
     }
+
+    /// Regression: an out-of-order `jxlp` box with no payload must not surface
```
veluca93: Please remove this test, and replace it by a regular jxl file using the existing test infrastructure - that should be able to reproduce the problem.

lilith: Done in dd3f456 — hand-written test removed, replaced by `resources/test/ooo_jxlp_empty_dc_group_boxes.jxl` (878 bytes) plus a `declare_test_file!` entry, so it runs through the standard harness.

The file is 6145×1 lossless from `cjxl 0.12.0 -q 100 -e 7 --output_mode=2`. 6145 px is four 2048 px DC group columns, which is the smallest shape that reproduces — a modular frame has no DC data, so the container carries one empty `jxlp` box per DC group, and four consecutive empty boxes exhaust the caller's retry path. At 6144 px (three DC groups) it decodes even without the fix.

All nine generated tests fail on main and pass with the patch, including `test_decode_test_file_chunks_*`, which feeds the file a byte at a time.

veluca93 (approved):

PR #960, opened 2026-09-09
Fix undersized row buffer.

veluca93:
A single call to process_row_chunk() might have an xsize of up to chunk_size + 2*border_size_x. This mostly comes up in progressive decoding (which might have unusual group orderings).

### DIFF
--- a/jxl/src/render/low_memory_pipeline/mod.rs
+++ b/jxl/src/render/low_memory_pipeline/mod.rs
@@ -69,7 +69,7 @@ impl LowMemoryRenderPipelinePerThread {
                     *next_y_border as usize,
                     stage.shift().1 as usize,
                     stage.shift().0 as usize,
-                    p.shared.chunk_size >> *dsx,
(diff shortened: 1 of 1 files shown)


jxl/src/render/low_memory_pipeline/mod.rs
```diff
@@ -69,7 +69,7 @@ impl LowMemoryRenderPipelinePerThread {
                     *next_y_border as usize,
                     stage.shift().1 as usize,
                     stage.shift().0 as usize,
-                    p.shared.chunk_size >> *dsx,
+                    (p.shared.chunk_size + p.border_size.0 * 2) >> *dsx,
```
sboukortt: Extremely minor but I wonder if I wouldn’t find this more readable as:

```suggestion
                    (p.shared.chunk_size + 2 * p.border_size.0) >> *dsx,
```

sboukortt (approved):

PR #965, opened 2026-09-10
Fix undersized input row buffer.

tnikkel:
5dfeb9e5fddb1ffebb709745f9c4739ec0cc2e82 grew the per stage row buffers to
chunk_size + 2*border_size, but the input row buffers allocated just above them
were left at chunk_size. BufferFiller::fill writes xsize + 2*input_border
pixels into those, so a rect from ready_image_area that is wider than
chunk_size runs off the end of one.

The padding inside RowBuffer hides this until 2*border_size + input_border goes
past 32 elements. The test image has an extra channel with 8x upsampling, which
puts border_size at 16 while the colour channels are still allocated for a 256
wide group. It panics with a slice index out of range, and the incremental
decode reaches it without several threads.


veluca93 (approved):

veluca93:
Good catch :-)

PR #967, opened 2026-09-10
Fix division by zero and empty channel handling during squeeze

veluca93:
- Evaluate empty size rects upfront in get_grid_rect to avoid division by zero in origin calculation
- Only increment avg_rect in SqueezeInfo::new if the rectangle is non-empty
- Forbid squeezing empty channels in check_squeeze_params

### DIFF
--- a/jxl/src/frame/modular/transforms/squeeze.rs
+++ b/jxl/src/frame/modular/transforms/squeeze.rs
@@ -34,6 +34,11 @@ pub fn check_squeeze_params(
     if channels[params.begin_channel as usize].1.is_meta() && !params.in_place {
         return Err(Error::MetaSqueezeRequiresInPlace);
     }
+    for c in &channels[params.begin_channel as usize..end_channel] {
+        if c.1.size.0 == 0 || c.1.size.1 == 0 {
+            return Err(Error::TooManySqueezes);
+        }
+    }
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


jxl/src/frame/modular/transforms/squeeze.rs
```diff
@@ -34,6 +34,11 @@ pub fn check_squeeze_params(
     if channels[params.begin_channel as usize].1.is_meta() && !params.in_place {
         return Err(Error::MetaSqueezeRequiresInPlace);
     }
+    for c in &channels[params.begin_channel as usize..end_channel] {
+        if c.1.size.0 == 0 || c.1.size.1 == 0 {
+            return Err(Error::TooManySqueezes);
```
jonsneyers, line 39: Are you sure this isn't causing trouble in valid bitstreams that are using the default squeeze script on an image with an extreme aspect ratio?

I don't think 0xH / Wx0 channels have to return error, as long as they're ignored during decode and counted towards the `nb_channels_tr` limit.

veluca93: I believe libjxl has the same check: https://github.com/libjxl/libjxl/blob/b5def9fb509d0f2421c8a5bcd7aa6f5a627363c4/lib/jxl/modular/transform/squeeze.cc#L492

veluca93: And in particular, the default squeeze script only squeezes the avg channel, whose size is rounded up every time, so that should never get to zero size.

jonsneyers (approved):

PR #969, opened 2026-09-10
Fix Passes::downsampling_bracket computation to conform to spec

veluca93:
Ensure that downsampling shift brackets partition [0, 3) across passes without overlapping when min_shift reaches 0, and return an empty bracket when no channels belong to a pass.


tirr-c (approved):
(side note: I precomputed a `BTreeMap` for `last_pass`-to-bracket mapping in jxl-oxide)

PR #971, opened 2026-09-12
Avoid allocation for upsample kernels

hjanuschka:
A Chromium Android WebView crash stack for issue 560379320 points to the `Vec::with_capacity` in `Upsample<8, 3>::new`.

Store the flattened kernels directly in the stage and build them in place. This removes the separate kernel allocation and the temporary four-dimensional array while preserving the factor-specific storage size.

The new test guards that the 2x, 4x, and 8x stages contain their kernels inline. Existing upsampling consistency and custom-kernel tests cover the generated values.


veluca93:
I'm pretty sure the reported stack trace is bogus and the issue is a stack overflow, fixing it in https://github.com/libjxl/jxl-rs/pull/973.

Issue #985, opened 2026-09-18
Replace `get_unchecked` buffer access in the weighted predictor with safe slicing (no perf regression)


MuxammadiyevG:
`WeightedPredictorState::predict_and_property` / `update_errors` in `jxl/src/frame/modular/predict.rs` use ~15 `get_unchecked` calls: 10 on the two double-row scratch buffers (`pred_errors_buffer`, `error`) and 5 on the `DIVLOOKUP` table. The buffer ones rely on a hand-written length invariant in a comment.

I have a patch that removes the 10 buffer accesses in favour of safe indexing that the optimizer can prove in bounds, so no per-access bounds checks are emitted:

- a `double_rows` helper narrows each buffer to exactly `2 * row_len` before `split_at(row_len)`, so both the current and previous row are known to be exactly `row_len` long;
- one `assert!(pos.0 + 1 < row_len)` per call covers every column used (all are `<= pos.0 + 1`), and the north-east column is clamped against `row_len` so that bound is visible too.

`update_errors` then no longer needs `#[allow(unsafe_code)]`. I left the `DIVLOOKUP` lookups alone: I tried `DIVLOOKUP[i & 63]` and measured it at +3–4% in the predictor micro-benchmark, so the existing comment ("the compiler doesn't seem to realize this") still holds.

Measurements (I'd include them in the PR):

- Predictor micro-benchmark (new `jxl/benches/modular_predict.rs`, criterion, behind a `bench_internals` feature so the private module stays private by default): −1.1% / +1.1% / +1.4% on 256², 512², 1024×256, with a ±0.5% noise floor. I checked the generated asm: the only remaining cold paths are the asserts.
- End-to-end `jxl_cli` decode bench on the five test images with the most weighted-predictor samples (sunset_logo, issue648_palette0, spot, green_queen_modular_e3, lossless_pfm): no statistically significant change on any of them.
- All 159 modular tests (incl. conformance decodes) and the golden-number predictor test pass unchanged.

Before I open the PR: is this something you'd take? Happy to drop the benchmark commit / the `bench_internals` feature if you'd rather not have that surface, or to adjust the approach.

The branch is here if you want to look at the diff before deciding: https://github.com/libjxl/jxl-rs/compare/main...MuxammadiyevG:jxl-rs:harden/predict-safe-rows

veluca93:
Are you sure this holds across architectures?

If you open a PR with just the WP changes I am happy to run a benchmark.

[...]

PR #986, opened 2026-09-18
modular: use safe row slicing in the weighted predictor

MuxammadiyevG:
Refs #985.

Removes 10 of the 15 `unsafe` blocks in `jxl/src/frame/modular/predict.rs` (all the scratch-buffer accesses in `WeightedPredictorState`) without adding per-access bounds checks. Only the weighted-predictor change, as requested — the micro-benchmark I used is on a separate branch and can be a follow-up if wanted.

### Change

`double_rows` / `double_rows_mut` narrow each double-row buffer to exactly `2 * row_len` entries and then `split_at(row_len)`, so the optimizer knows both rows are exactly `row_len` long. With a single `assert!(pos.0 + 1 < row_len)` per call (every column used is `<= pos.0 + 1`), the north-east column clamped against `row_len`, and the north-east `error` lookup done with `get` (so the clamp condition and the bounds check are the same branch), every row access is provably in bounds.

`update_errors` no longer needs `#[allow(unsafe_code)]`. The 5 `DIVLOOKUP` lookups keep `get_unchecked`: `DIVLOOKUP[i & 63]` measured at +3–4%.

### Across architectures

Counting `panic_bounds_check` sites on the library build (`cargo rustc -p jxl --lib --release -- --emit=asm`), main → this branch: x86_64 2121 → 2123, aarch64 1513 → 1512, none in the predictor. What remains per sample is the two asserts and the slice narrowing/split checks.

### Numbers (x86_64, my machine)

Predictor micro-benchmark (criterion, back-to-back vs. main, noise about ±1%): 256×256 +0.2%, 512×512 +0.4%, 1024×256 −0.4%.

End-to-end `cargo bench -p jxl_cli --bench decode` on the five test images with the most weighted-predictor samples (sunset_logo, issue648_palette0, spot, green_queen_modular_e3, lossless_pfm): no statistically significant change on any of them.

Happy for your benchmark to be the deciding one.

### Tests

No functional change: the golden-number predictor test and all 159 `modular` tests (including conformance decodes) pass unchanged. `cargo fmt` / `clippy` clean.


jxl/src/frame/modular/predict.rs
```diff
@@ -470,20 +480,16 @@ impl WeightedPredictorState {
         ((pred + PREDICTION_ROUND) >> PRED_EXTRA_BITS, p as i32)
     }
 
-    #[allow(unsafe_code)]
     #[inline(always)]
     pub fn update_errors(&mut self, correct_val: i32, pos: (usize, usize)) {
         assert!(pos.0 < self.xsize);
```
veluca93, line 470: Do we still need this assert?

jxl/src/frame/modular/predict.rs
```diff
@@ -217,16 +217,44 @@ fn add_bits(x: i32) -> i64 {
...
+        (second, first)
+    }
+}
+
 #[derive(Debug)]
 pub struct WeightedPredictorState {
     prediction: [i64; NUM_PREDICTORS],
     pred: i64,
-    // Safety invariant:
+    // Invariant (established in `new`):
```
veluca93: Given this is no longer a safety invariant, I think it's a bit overly verbose.

veluca93 (commented):
In general, the comments could be a bit less verbose.

jxl/src/frame/modular/predict.rs
MuxammadiyevG: Trimmed to two lines, and shortened the other new comments as well.

jxl/src/frame/modular/predict.rs
MuxammadiyevG: Semantically no, it's implied by the one below — but it's what lets the optimizer rule out `pos.0 + 1` wrapping. Measured on the library build (`cargo rustc -p jxl --lib --release -- --emit=asm`, counting `panic_bounds_check` sites), main → this branch: x86_64 2121 → 2123, aarch64 1513 → 1512. Dropping this assert: 2129 / 1520 (checks back on `pe_prev[pos.0]`, `pe_cur[pos.0]` on aarch64). Folding both into one `assert!(a && b)`: 2155 / 1542. Kept both, with a two-line comment.

veluca93:
### Performance Summary (Commit `428282f`)

| Machine | Threading | Base MP/s | PR MP/s | Avg Improvement | Base Mem | PR Mem | Mem Change |
|---------|-----------|-----------|---------|-----------------|----------|--------|------------|
| desktop | Single | 50.11 | 49.66 | -0.91% ± 0.85% | 49.5 MB | 49.6 MB | +0.17% |
| desktop | Multi | 222.20 | 221.39 | -0.37% ± 0.94% | 170.6 MB | 165.7 MB | -2.91% 🚀 |
| framework-desktop | Single | 55.41 | 54.99 | -0.76% ± 0.23% | 51.1 MB | 51.2 MB | +0.12% |
| framework-desktop | Multi | 248.47 | 245.98 | -1.00% ± 0.72% | 135.1 MB | 133.4 MB | -1.27% 🚀 |
| pixel7a | Single (Fast) | 21.98 | 21.99 | +0.05% ± 0.55% | 29.3 MB | 29.3 MB | +0.05% |
| pixel7a | Single (Mid) | 16.13 | 16.17 | +0.24% ± 0.41% | 29.2 MB | 29.3 MB | +0.21% |
| pixel7a | Multi | 51.48 | 52.56 | +2.10% ± 3.42% | 37.0 MB | 37.2 MB | +0.45% |

[Detailed per-image results](https://jxl-rs-perf.lucaversari.it/pr_986.html)

MuxammadiyevG:
Thanks for running it. That was on `bac7f32`; `aaf7234` (pushed since) removes the derived-index bounds checks that were still emitted on the library build, which I'd expect to be the framework-desktop −2.3%. Could the bot rerun on the current head?

